PR #1, opened 2022-04-05
Initial implementation of packet (de)serialization.

davidv1992:


folkertdev (approved):

ntp-proto/src/timetypes.rs
```diff
@@ -0,0 +1,45 @@
+#[derive(Debug, Copy, Clone, Eq, PartialEq, PartialOrd, Ord)]
+pub struct NtpTimestamp {
+    timestamp: u64,
+}
+
+impl NtpTimestamp {
+    pub(crate) fn from_bits(bits: [u8; 8]) -> NtpTimestamp {
```
djc, line 7: Style nit: I prefer to use `Self` pervasively in return types and struct initializers. I like to `warn(clippy::use_self)` for this.

ntp-proto/src/timetypes.rs
```diff
@@ -0,0 +1,45 @@
...
+        assert!(self.duration > 0);
+        assert!(self.duration <= 0x0000FFFFFFFFFFFF);
+        (((self.duration & 0x0000FFFFFFFF0000) >> 16) as u32).to_be_bytes()
+    }
+
+    #[cfg(test)]
+    pub(crate) fn from_fixed_int(duration: i64) -> NtpDuration {
+        NtpDuration{ duration }
+    }
+}
```
djc, line 45: Nit: would be nice to configure your editor so it puts in trailing newlines by default.

ntp-proto/src/packet.rs
```diff
@@ -0,0 +1,239 @@
...
+}
+
+impl NtpLeapIndicator {
+    fn from_bits(bits: u8) -> NtpLeapIndicator {
+        match bits {
+            0 => NtpLeapIndicator::NoWarning,
+            1 => NtpLeapIndicator::Leap61,
+            2 => NtpLeapIndicator::Leap59,
+            3 => NtpLeapIndicator::Unknown,
+            _ => panic!("Internal error: Trying to interpret more than 2 bits as leap indicator"),
```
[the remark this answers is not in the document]
folkertdev: We know for sure here that the input `u8` is only 2 bits and hence that the match is exhaustive: this is guaranteed by the decoder. We'll document this invariant.

Totally on board of course with not having random panics, but I think it's justified here

PR #4, opened 2022-04-05
Proto default instances

folkertdev:


ntp-proto/src/timetypes.rs
```diff
@@ -1,45 +1,69 @@
...
+}
+
 #[derive(Debug, Copy, Clone, Eq, PartialEq, PartialOrd, Ord)]
 pub struct NtpDuration {
     duration: i64,
 }
 
+impl Default for NtpDuration {
+    fn default() -> Self {
+        Self::from_bits_short([0u8; 4])
```
davidv1992, line 48: This is somewhat convoluted, wouldn't a simple `NtpDuration { duration: 0 }` be better?

davidv1992 (commented):
Minor comment, but otherwise looks good.

ntp-proto/src/timetypes.rs
```diff
@@ -1,45 +1,57 @@
...
     }
 }
 
-#[derive(Debug, Copy, Clone, Eq, PartialEq, PartialOrd, Ord)]
+impl Sub for NtpTimestamp {
+    type Output = NtpDuration;
+
+    fn sub(self, rhs: Self) -> Self::Output {
+        NtpDuration {
+            duration: self.timestamp as i64 - rhs.timestamp as i64,
```
djc, line 30: This seems scary. The `u64` range is smaller than `i64` on the low side and larger on the high side. So this should probably be a `checked_sub()` or `saturating_sub()` method instead, depending on what we actually need? (Maybe both?)

PR #7, opened 2022-04-05
Clock trait + unix <-> ntp conversion

folkertdev:


ntp-proto/src/timetypes.rs
```diff
@@ -16,10 +21,48 @@ impl NtpTimestamp {
...
+
+        // alternatively, abuse FP arithmetic to save an instruction
+        // let fraction = (nanos as f64 * 4.294967296) as u64;
+
+        let timestamp = (seconds << 32) + fraction;
+        NtpTimestamp::from_bits(timestamp.to_be_bytes())
+    }
+
+    #[cfg(test)]
+    pub(crate) fn duration_since_unix_epoch(self) -> std::time::Duration {
```
djc, line 45: I feel like there's already too much code here that is `#[cfg(test)] pub(crate)`. It can sometimes be found, but most of the code that is only necessary for tests should live in one of the tests modules, not in the library. If need be, just make the `NtpTimestamp::timestamp` field `pub(crate)` (though it might be better to put tests that need access like that in a test module inside this module, which can access the field even if private).

ntp-proto/src/timetypes.rs
```diff
@@ -16,10 +21,48 @@ impl NtpTimestamp {
...
         NtpTimestamp { timestamp }
     }
+
+    #[cfg(test)]
+    pub(crate) const fn seconds(self) -> u32 {
+        (self.timestamp >> 32) as u32
+    }
+
+    #[cfg(test)]
+    pub(crate) const fn fraction(self) -> u32 {
```
djc, line 63: This could use some documentation on what `fraction` actually is. Probably also a comment on the `timestamp` field to explain what it contains exactly?

PR #9, opened 2022-04-05
Expose clock to users + minor cleanup of tests.

davidv1992:


ntp-proto/src/clock.rs
```diff
@@ -2,11 +2,11 @@ use std::time::SystemTime;
 
 use crate::NtpTimestamp;
 
-trait NtpClock {
+pub trait NtpClock {
     fn now(&self) -> NtpTimestamp;
 }
 
-struct SystemClock;
+pub struct SystemClock;
```
djc, line 9: I would at least start by spelling this as `SystemClock(())` so that this crate controls initialization.

PR #10, opened 2022-04-05
perform a measurement with time.google.com

folkertdev:


ntp-daemon/src/main.rs
```diff
@@ -0,0 +1,53 @@
+use ntp_proto::{NtpClock, NtpHeader, SystemClock};
+use std::env;
+use std::error::Error;
+use tokio::net::UdpSocket;
+
+/// Connect with the `time.google.com` NTP server
+async fn setup_connection() -> std::io::Result<UdpSocket> {
+    let addr = env::args()
```
djc, line 8: I'd bring in a clap (derive) based parser early. It can take a `SocketAddr` directly.

djc: Well, you're actually already passing in command-line arguments. I'd just suggest pulling in clap sooner rather than later since (a) it's pretty easy to do and (b) it'll make the code simpler.

PR #21, opened 2022-04-07
Cleanup packet parsing

davidv1992:


ntp-proto/src/packet.rs
```diff
@@ -134,6 +138,8 @@ impl NtpHeader {
     }
 
     pub fn serialize(&self) -> [u8; 48] {
+        // Version should only ever be set internally in this module, so
+        // violations of this should never happen.
```
djc, line 142: I'd make the assert a `debug_assert!()` here, we probably wouldn't want to blow up a production deployment?

davidv1992: Given what NTP does, especially on a client I think the safer option is to actually blow up in case of a programming error and just not adjust the clock anymore, rather than ignore a potential issue and in turn maybe start steering the clock in the wrong direction, as in that case the failure mode is much more subtle. However, I think this is a discussion we should separate from the actual code, so I have created issue #24 for that.

ntp-proto/src/packet.rs
```diff
@@ -15,7 +15,9 @@ impl NtpLeapIndicator {
             1 => NtpLeapIndicator::Leap61,
             2 => NtpLeapIndicator::Leap59,
             3 => NtpLeapIndicator::Unknown,
-            _ => panic!("Internal error: Trying to interpret more than 2 bits as leap indicator"),
+            _ => unreachable!(),
```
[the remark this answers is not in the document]
davidv1992: Comments reordered/added as suggested. Keeping the unreachable as_is per discussion in #24

djc (approved):

folkertdev (approved):

PR #25, opened 2022-04-07
Clock filter

folkertdev:
The first piece of actual logic. 

I've tried to test the separate components, but am open to suggestions for how to test the whole.


ntp-proto/src/filter.rs
```diff
@@ -0,0 +1,315 @@
...
+        }
+    }
+
+    fn sort_by_delay(&mut self) {
+        self.register.sort_by(|t1, t2| {
+            t1.delay
+                .partial_cmp(&t2.delay)
+                .unwrap_or_else(|| panic!("got a NaN"))
+        });
+    }
```
folkertdev, line 70: can we just ignore `NaN` and pick an ordering here? I don't see downsides

ntp-proto/src/filter.rs
```diff
@@ -0,0 +1,315 @@
...
+        // s.rho expressed in seconds.
+        jitter.max(s.precision)
+    }
+}
+
+pub struct System {
+    precision: f64,
+    poll: NtpDuration,
+    leap_indicator: NtpLeapIndicator,
+}
```
folkertdev, line 135: these structs are not complete. Figuring out what the correct type should be is a matter of trial and error, so I'm adding fields as needed.

ntp-proto/src/filter.rs
```diff
@@ -0,0 +1,315 @@
...
+impl ClockFilterContents {
+    #[allow(dead_code)]
+    fn new() -> Self {
+        Self {
+            register: [FilterTuple::dummy(); 8],
+        }
+    }
+
+    fn shift_and_insert(&mut self, new_tuple: FilterTuple, dispersion_correction: NtpDuration) {
+        let mut current = new_tuple;
```
djc, line 55: Renaming the argument immediately doesn't seem useful?

ntp-proto/src/timetypes.rs
```diff
@@ -125,6 +125,32 @@ impl NtpDuration {
         (((self.duration & 0x0000FFFFFFFF0000) >> 16) as u32).to_be_bytes()
     }
 
+    /// NtpDuration::from_seconds(16.0)
+    pub(crate) const MAXDISP: Self = Self {
```
djc, line 129: Seems like we could just name this `MAX_DISPERSION`?

ntp-proto/src/timetypes.rs
```diff
@@ -125,6 +125,32 @@ impl NtpDuration {
...
+        duration: 68719476736,
+    };
+
+    /// NtpDuration::from_seconds(0.005)
+    #[allow(dead_code)]
+    pub(crate) const MINDISP: Self = Self { duration: 21474836 };
```
folkertdev, line 135: floating-point arithmetic is not allowed in `const` contexts. So I hardcoded these numbers. Not ideal but I don't see a better way

ntp-proto/src/filter.rs
```diff
@@ -0,0 +1,315 @@
...
+#[derive(Debug, Clone, Copy, PartialEq)]
+struct FilterTuple {
+    offset: NtpDuration,
+    delay: NtpDuration,
+    dispersion: NtpDuration,
+    time: NtpTimestamp,
+}
+
+impl FilterTuple {
+    fn dummy() -> Self {
```
djc, line 27: Could should this be an associated `const DUMMY` instead?

ntp-proto/src/timetypes.rs
```diff
@@ -298,4 +324,18 @@ mod tests {
...
+        };
+    }
+
+    #[test]
+    fn duration_seconds_roundtrip() {
+        assert_eq_epsilon!(NtpDuration::from_seconds(0.0).to_seconds(), 0.0, 1e-9);
+        assert_eq_epsilon!(NtpDuration::from_seconds(1.0).to_seconds(), 1.0, 1e-9);
+        assert_eq_epsilon!(NtpDuration::from_seconds(1.5).to_seconds(), 1.5, 1e-9);
+        assert_eq_epsilon!(NtpDuration::from_seconds(2.0).to_seconds(), 2.0, 1e-9);
+    }
```
folkertdev, line 340: we have to deal with precision loss here. We seem to be correct up to `1e-9` seconds, so nanoseconds. That seems acceptable to me

ntp-proto/src/filter.rs
folkertdev: I think the names convey useful information here?

ntp-proto/src/filter.rs
djc: Maybe consider https://crates.io/crates/ordered-float?

ntp-proto/src/filter.rs
djc: It's either `new_tuple` or `current`. Changing the binding name for the value immediately at the start of the function doesn't seem useful.

davidv1992 (changes requested):
Having looked at this, I think this is not yet in the right direction to proceed with. To me it just still reeks to much of the reference C implementation, which gives (in my view) issues with dealing with global variables and global state. Also, as it stands, it makes this code, especially the actual clock filter function, almost impossible to properly test in isolation, something I think we really should want to be able to do.

Perhaps it is an idea to take some time for a meeting early next week to discuss options around organizing this better?

ntp-proto/src/filter.rs
```diff
@@ -0,0 +1,283 @@
...
+//
+// Specifically this is a rust implementation of the `clock_filter()` routine,
+// described in the appendix
+//
+//      https://datatracker.ietf.org/doc/html/rfc5905#appendix-A.5.2
+
+use crate::{packet::NtpLeapIndicator, NtpDuration, NtpTimestamp};
+
+/// frequency tolerance (15 ppm)
+const ONE_OVER_PHI: i64 = 15_000_000;
```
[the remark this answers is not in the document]
folkertdev: ah, yes. So `PHI` itself is 15ppm, but we often need to multiply a `NtpDuration` with that value. By using `ONE_OVER_PHI` we can do an integer division instead. I'll make this more clear

ntp-proto/src/filter.rs
```diff
@@ -0,0 +1,283 @@
...
+        }
+    }
+
+    fn sort_by_delay(&mut self) {
+        self.register.sort_by(|t1, t2| {
+            t1.delay
+                .partial_cmp(&t2.delay)
+                .unwrap_or(std::cmp::Ordering::Less)
+        });
+    }
```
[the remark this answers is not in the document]
folkertdev: hence it's important to see where this function is called. It is used only on the "temporary list" which is the copy
```
    let mut temporary_list = peer.clock_filter;

    temporary_list.sort_by_delay();
```

ntp-proto/src/filter.rs
```diff
@@ -0,0 +1,283 @@
...
+            .sqrt();
+
+        // root mean square average (RMS average). - 1 to exclude the smallest_delay
+        let jitter = root_mean_square / (register.len() - 1) as f64;
+
+        // In order to ensure consistency and avoid divide exceptions in other
+        // computations, the psi is bounded from below by the system precision
+        // s.rho expressed in seconds.
+        jitter.max(s.precision)
+    }
```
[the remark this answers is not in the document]
folkertdev: The `valid_tuples` call on line 111 makes sure just the valid tuples are used in the calculation

ntp-proto/src/filter.rs
```diff
@@ -0,0 +1,283 @@
...
+
+    // Popcorn spike suppressor.  Compare the difference between the
+    // last and current offsets to the current jitter.  If greater
+    // than SGATE (3) and if the interval since the last offset is
+    // less than twice the system poll interval, dump the spike.
+    // Otherwise, and if not in a burst, shake out the truechimers.
+    let too_soon = (smallest_delay.time - peer.t) < (s.poll * 2i64);
+    if (peer.offset - dtemp).to_seconds().abs() > SGATE * peer.jitter && too_soon {
+        return;
+    }
```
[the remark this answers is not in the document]
folkertdev: yeah, there are other sections of the implementation talking about spikes, but the spec itself does not seem to really talk about it. 

So it looks like this is an engineering decision, but that means we should re-evaluate if we want to keep this section

ntp-proto/src/filter.rs
```diff
@@ -0,0 +1,283 @@
...
+
+#[allow(dead_code)]
+pub fn clock_filter(
+    peer: &mut Peer,
+    s: &System,
+    c: &LocalClock,
+    clock_offset: NtpDuration,
+    roundtrip_delay: NtpDuration,
+    dispersion: NtpDuration,
+) {
```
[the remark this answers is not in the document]
folkertdev: I don't disagree, but this still seems like the fastest way forward.

Disentangling the state is tricky. I've been continuing work on clock selection and combining, and it's obvious that you run into issues at some point, where either you need to clone or use lifetimes/RC/... It's also clear that the `struct`s in the reference implementation are way too big and monolithic. 

But, I like getting it down into code where we can have the compiler help us. To me, that is just the easiest way of making progress, rather than studying a diagram. 

There are a lot of moving parts, and we don't need to get this right on the first draft, though it is of course vital we get it right eventually.

ntp-proto/src/filter.rs
```diff
@@ -0,0 +1,283 @@
...
+    // Otherwise, and if not in a burst, shake out the truechimers.
+    let too_soon = (smallest_delay.time - peer.t) < (s.poll * 2i64);
+    if (peer.offset - dtemp).to_seconds().abs() > SGATE * peer.jitter && too_soon {
+        return;
+    }
+
+    peer.t = smallest_delay.time;
+    if peer.burst_counter == 0 {
+        todo!()
+        // clock_select();
```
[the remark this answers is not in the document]
folkertdev: yes, hence the `todo!` because it turns out that we want to do this very differently from what the reference is doing, for the reasons you describe. 

How to do it was harder to see at the time, but now that I've looked at clock selection in more detail we indeed should make this have its own (smaller) state, passing in all of `peer` and `s` is just confusing

ntp-proto/src/filter.rs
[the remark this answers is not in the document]
davidv1992: Ok, in that case I think I would rather split things of into two separate types and make it clearer at the type system level that this is never meant to be called on the actual storage but only on the processing variant. That also deals with the fact that it is not the clearest when the jitter and dispersion functions can be called.

ntp-proto/src/filter.rs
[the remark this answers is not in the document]
davidv1992: ah apologies, missed that.

ntp-proto/src/filter.rs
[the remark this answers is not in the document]
davidv1992: I disagree then with this approach, because starting from the current structure presented will also influence where we end up, and as we already know we want something radically different, I'd far rather take a starting point that is at least a little bit closer to what we eventually want, instead of this starting point where not only is it already very clear that it is not the way forward, but it also pretty much precludes any testing.

ntp-proto/src/filter.rs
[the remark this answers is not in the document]
davidv1992: Ok, but if it is a multiplication by 15*10^-6, shouldn't the constant be different?

ntp-proto/src/filter.rs
```diff
@@ -196,64 +193,61 @@ pub struct LocalClock {
...
-        offset: clock_offset,
-        delay: roundtrip_delay,
-        dispersion,
-        time: c.t,
-    };
-
-    let dispersion_correction = multiply_by_phi(c.t - peer.t);
-    peer.clock_filter
-        .shift_and_insert(new_tuple, dispersion_correction);
+pub struct PeerProcessStatistics {
```
djc, line 162: Naming nit: maybe ditch the `Process` part? It doesn't seem to add value here.

PR #32, opened 2022-04-14
Find interval

folkertdev:


ntp-proto/src/filter.rs
```diff
@@ -359,28 +357,56 @@ fn construct_candidate_list<'a>(
...
+        let mut chime = 0; // variable "c"
 
         // Scan the chime list from lowest to highest to find the lower endpoint.
         for tuple in chime_list {
             chime -= tuple.endpoint_type as i32;
-            if chime >= (n - found) as i32 {
-                low = tuple.edge;
+
+            // the code skeleton uses `n - found` here, which is wrong!
+            if chime >= (n - allow) as i32 {
```
folkertdev, line 408: this is subtle, but using `found` would just find the same result for every iteration. The specification text uses the `f` variable, which we give the name `allow`. 

Another case where you need to read very carefully.

ntp-proto/src/filter.rs
```diff
@@ -393,8 +419,10 @@ fn find_interval(chime_list: &[CandidateTuple]) -> (NtpDuration, NtpDuration) {
         chime = 0;
         for tuple in chime_list.iter().rev() {
             chime += tuple.endpoint_type as i32;
-            if chime >= (n - found) as i32 {
-                high = tuple.edge;
+
+            // the code skeleton uses `n - found` here, which is wrong!
+            if chime >= (n - allow) as i32 {
+                high = Some(tuple.edge);
                 break;
```
folkertdev, line 426: why can't you break from a `for`/`while` loop with a value? Seems to me like that should just work.

ntp-proto/src/filter.rs
```diff
@@ -405,20 +433,19 @@ fn find_interval(chime_list: &[CandidateTuple]) -> (NtpDuration, NtpDuration) {
...
-        //  increment the number of allowed falsetickers and go
-        //  around again.  If not and the intersection is
-        //  non-empty, declare success.
+        //  least one truechimer with no midpoint.  If so, increment
+        //  the number of allowed falsetickers and go around again.
         if found > allow {
             continue;
         }
 
-        if high > low {
```
folkertdev, line 416: this check was done just to verify that low and high were written to

ntp-proto/src/filter.rs
djc: Because `for`/`while` might break out of the loop due to their iterator running out/the loop condition becoming false, in which case you wouldn't have the opportunity to yield a value.

ntp-proto/src/filter.rs
```diff
@@ -359,28 +357,56 @@ fn construct_candidate_list<'a>(
...
+
+/// Collect the candidates within the correctness interval
+#[allow(dead_code)]
+fn construct_survivors<'a>(
+    chime_list: &[CandidateTuple<'a>],
+    local_clock_time: NtpTimestamp,
+) -> Vec<SurvivorTuple<'a>> {
+    let mut survivors = Vec::new();
+
+    if let Some((low, high)) = find_interval(chime_list) {
```
djc, line 374: You could do an early return here, which I think would make this code more readable.

It seems like this function (and potentially other functions in this module) are filtering (and enriching?) an initial list. It might be nice to avoid allocations while doing so and instead spin this as a `filter_survivor(candidate: CandidateTuple<'a>, local_clock_time: NtpTimestamp) -> Option<SurvivorTuple<'a>>`.

ntp-proto/src/filter.rs
folkertdev: ah right, you need something like python's `for`/`else` construct

ntp-proto/src/filter.rs
folkertdev: yes that worked out nicely!

ntp-proto/src/filter.rs
```diff
@@ -366,25 +366,32 @@ struct SurvivorTuple<'a> {
 /// Collect the candidates within the correctness interval
 #[allow(dead_code)]
 fn construct_survivors<'a>(
-    chime_list: &[CandidateTuple<'a>],
+    chime_list: &'a [CandidateTuple<'a>],
```
djc, line 369: So now this is allocating a `Vec<SurvivorTuple<'a>>`. I think you could just skip the `collect()` and return `imp lIterator<SurvivorTuple<'a>> + 'a` instead? (Unless there are other uses for `find_interval() and `filter_survivor()`, I would probably merge all three functions into one.)

folkertdev: we need these in a vector because we'll be removing (iteratively filtering) elements from that vector (see the next PR)

davidv1992 (commented):
Some minor things around comments and the new field in the Peer struct, but other than that the main find_interval implementation looks good.

ntp-proto/src/filter.rs
```diff
@@ -402,26 +437,37 @@ fn find_interval(chime_list: &[CandidateTuple]) -> (NtpDuration, NtpDuration) {
 
         //  If the number of midpoints is greater than the number
         //  of allowed falsetickers, the intersection contains at
-        //  least one truechimer with no midpoint.  If so,
-        //  increment the number of allowed falsetickers and go
-        //  around again.  If not and the intersection is
-        //  non-empty, declare success.
+        //  least one truechimer with no midpoint.  If so, increment
+        //  the number of allowed falsetickers and go around again.
         if found > allow {
```
davidv1992, line 442: According to the spec on page 43 point 5 we should reject even if found < allow (only accept if found == allow). Is this a conscious deviation from that part?

folkertdev: yes, the logic in practice is such that we could jump over a `!=`, so we use `>` to be safe

ntp-proto/src/filter.rs
```diff
@@ -197,6 +197,7 @@ pub struct Peer {
     last_measurements: LastMeasurements,
     last_packet: NtpHeader,
     time: NtpTimestamp,
+    stratum: u8,
```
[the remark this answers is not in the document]
folkertdev: we now use the `last_packet`s stratum

davidv1992:
With the current changes this is at a point where I think it is better to do further iterations on this from main.

PR #33, opened 2022-04-14
Clock select

folkertdev:


ntp-proto/src/filter.rs
```diff
@@ -448,6 +438,91 @@ fn find_interval(chime_list: &[CandidateTuple]) -> Option<(NtpDuration, NtpDurat
...
+    // then don't do a clock hop.  Otherwise, select the first
+    // survivor on the list as the new system peer.
+    let first_survivor = survivors[0].peer;
+
+    // TODO we don't do the "If the old system peer is on the list" check yet!
+    if old_system_peer.stratum == first_survivor.stratum {
+        Some(old_system_peer)
+    } else {
+        Some(first_survivor)
+    }
```
folkertdev, line 523: to discuss: this "keep the old system peer" logic is not described in the spec

ntp-proto/src/filter.rs
```diff
@@ -14,6 +14,9 @@ use crate::{packet::NtpLeapIndicator, NtpDuration, NtpHeader, NtpTimestamp};
 const MAX_STRATUM: u8 = 16;
 const MAX_DISTANCE: NtpDuration = NtpDuration::ONE;
 
+// TODO this should be 4 in production?!
+const NSANE: usize = 1;
```
djc, line 16: This is a very opaque variable name. Maybe expand the name and/or provide a comment about the semantics?

ntp-proto/src/filter.rs
```diff
@@ -367,7 +370,7 @@ fn construct_survivors<'a>(
             let p = tuple.peer;
```
djc, line 368: I'd probably rename this as `peer` as well, which is just clearer.

ntp-proto/src/filter.rs
```diff
@@ -435,6 +438,90 @@ fn find_interval(chime_list: &[CandidateTuple]) -> Option<(NtpDuration, NtpDurat
     None
 }
 
+/// Discard the survivor with maximum selection jitter until a termination condition is met.
+fn cluster_algorithm(candidates: &mut Vec<SurvivorTuple>) {
+    // sort the candidates by increasing lambda_p (the merit factor)
+    candidates.sort_by(|a, b| a.metric.cmp(&b.metric));
+
+    // let NMIN be the minimum required number of survivors.
+    const NMIN: usize = 1;
```
djc, line 445: Same as for `NSANE`. Maybe `MIN_SURVIVORS`?

ntp-proto/src/filter.rs
```diff
@@ -435,6 +438,90 @@ fn find_interval(chime_list: &[CandidateTuple]) -> Option<(NtpDuration, NtpDurat
...
+
+    cluster_algorithm(&mut survivors);
+
+    // Pick the best clock.  If the old system peer is on the list
+    // and at the same stratum as the first survivor on the list,
+    // then don't do a clock hop.  Otherwise, select the first
+    // survivor on the list as the new system peer.
+    let first_survivor = survivors[0].peer;
+
+    if old_system_peer.stratum == first_survivor.stratum {
```
djc, line 517: I'd write this as a `match`, could do as below to also deduplicate the `Some`-wrapping:

```rust
Some(match old_system_peer.stratum == survivors[0].peer.stratum {
    true => old_system_peer,
    false => survivors[0].peer
})
```

folkertdev: based on working with this function, I think this logic should live outside of it. So now we just return the list of survivors

davidv1992:
Changed base to main as it should now be merged directly there to avoid issues.

ntp-proto/src/filter.rs
```diff
@@ -182,16 +185,6 @@ pub struct PeerStatistics {
...
 
-#[allow(dead_code)]
-#[derive(Debug)]
-struct PeerConfiguration {
-    source_address: IpAddr,
-    source_port: u16,
-    destination_address: IpAddr,
-    destination_port: u16,
-    reference_id: ReferenceId,
-}
```
davidv1992, line 193: Why is this removed?

ntp-proto/src/filter.rs
```diff
@@ -7,13 +7,16 @@
...
-
 use crate::{packet::NtpLeapIndicator, NtpDuration, NtpHeader, NtpTimestamp, ReferenceId};
 
 const MAX_STRATUM: u8 = 16;
 const MAX_DISTANCE: NtpDuration = NtpDuration::ONE;
 
+// TODO this should be 4 in production?!
+const MIN_INTERSECTION_SURVIVORS: usize = 1;
+
+const MIN_CLUSTER_SURVIVORS: usize = 3;
```
davidv1992, line 18: Could you add explanations for what these are?

ntp-proto/src/filter.rs
```diff
@@ -451,6 +444,116 @@ fn find_interval(chime_list: &[CandidateTuple]) -> Option<(NtpDuration, NtpDurat
...
+
+    let mut min_peer_jitter = 2.0e9;
+    let mut max_selection_jitter = -2.0e9;
+
+    loop {
+        // for each candidate
+        for (index, candidate) in candidates.iter().enumerate() {
+            let p = candidate.peer;
+
+            min_peer_jitter = p.statistics.jitter.min(min_peer_jitter);
```
davidv1992, line 462: Nit: why not calculate this outside the loop with iterator::min?

ntp-proto/src/filter.rs
```diff
@@ -451,6 +444,116 @@ fn find_interval(chime_list: &[CandidateTuple]) -> Option<(NtpDuration, NtpDurat
...
+
+            // compute the selection jitter
+            let selection_jitter = candidates
+                .iter()
+                .map(|q| p.statistics.offset - q.peer.statistics.offset)
+                .map(|v| v.to_seconds().powi(2))
+                .sum::<f64>()
+                .sqrt();
```
davidv1992, line 470: This seems to be missing a division by the number of candidates remaining less 1.

ntp-proto/src/filter.rs
```diff
@@ -451,6 +444,116 @@ fn find_interval(chime_list: &[CandidateTuple]) -> Option<(NtpDuration, NtpDurat
...
+        }
+
+        // If the maximum selection jitter is less than the
+        // minimum peer jitter, then tossing out more survivors
+        // will not lower the minimum peer jitter, so we might
+        // as well stop.  To make sure a few survivors are left
+        // for the clustering algorithm to chew on, we also stop
+        // if the number of survivors is less than or equal to
+        // NMIN (3).
+        if max_selection_jitter < min_peer_jitter || candidates.len() <= MIN_CLUSTER_SURVIVORS {
```
davidv1992, line 485: Since we know the length condition already when starting the current iteration, couldn't we lift that to the outer loop and convert it to a while?

ntp-proto/src/filter.rs
```diff
@@ -451,6 +444,116 @@ fn find_interval(chime_list: &[CandidateTuple]) -> Option<(NtpDuration, NtpDurat
...
+/// be configured to avoid this algorithm by designating a
+/// preferred peer.
+#[allow(dead_code)]
+fn clock_combine<'a>(
+    survivors: &'a [SurvivorTuple<'a>],
+    local_clock_time: NtpTimestamp,
+) -> ClockCombine {
+    let mut y = 0.0;
+    let mut z = 0.0;
+    let mut w = 0.0;
```
davidv1992, line 539: What are these?

ntp-proto/src/filter.rs
```diff
@@ -451,6 +444,116 @@ fn find_interval(chime_list: &[CandidateTuple]) -> Option<(NtpDuration, NtpDurat
...
+
+    let first_offset = survivors[0].peer.statistics.offset;
```
davidv1992, line 541: Who is responsible for the sorting that needs to be done before this point?

ntp-proto/src/filter.rs
```diff
@@ -451,6 +444,116 @@ fn find_interval(chime_list: &[CandidateTuple]) -> Option<(NtpDuration, NtpDurat
...
+    for tuple in survivors {
+        let peer = tuple.peer;
+        let x = peer.root_distance(local_clock_time).to_seconds();
+        y += 1.0 / x;
+        z += peer.statistics.offset.to_seconds() / x;
+        w += (peer.statistics.offset - first_offset).to_seconds().powi(2) / x;
+    }
+
+    let offset = NtpDuration::from_seconds(z / y);
+    let jitter = NtpDuration::from_seconds((w / y).sqrt());
```
davidv1992, line 552: I don't think this produces the correct system jitter. I would expect an additional RMS operation also combining in the peer jitter of the best peer. (see section 11.2.3 of the standard, the formula for PSI)

davidv1992 (changes requested):
There seem to be a number of errors in the formulas here that really need fixing. Structure-wise it looks decent for now though.

ntp-proto/src/filter.rs
djc: If you're looping anyway, it's probably better to do everything in one loop rather than looping twice.

ntp-proto/src/filter.rs
folkertdev: it's unused. I figure we can add it again when we need it (probably when we start working on milestone 2)

ntp-proto/src/filter.rs
folkertdev: yeah I don't have strong feelings either way, but that was my thinking.

djc: Style suggestion: I often find the asymmetric `.min()` and `.max()` relatively hard to reason about. I've started using `Ord::min()` and `Ord::max()` instead.

folkertdev: wait I see now. This minimum is for the whole list of candidates, so it can be computed outside of both loops

folkertdev: oh, nice! we need to use `f64::min` here because floats are not `Ord`, but yes I like this

ntp-proto/src/filter.rs
folkertdev: I added some comments. But I have to say the spec is particularly vague here 

> The routine processes peer offset and jitter statistics to produce the combined system offset THETA and system peer jitter PSI_p, where each server statistic is weighted by the reciprocal of the root synchronization distance and the result normalized.

jitter is calculated from the offsets in the code skeleton, but I honestly don't understand why that is, and why the peer's `jitter` is not considered.

ntp-proto/src/filter.rs
folkertdev: good point. it is implied that this function only runs after the clustering algorithm (which gives the correctly-sorted output). But the code skeleton does not actually enforce this ...

ntp-proto/src/filter.rs
folkertdev: turns out that the skeleton is taking a shortcut here. We need to return a value, which means we need (effectively) a do-while loop. So sadly we can't move this condition up

ntp-proto/src/filter.rs
davidv1992: Ok, this might be a bug in the code skeleton then.

ntp-proto/src/filter.rs
folkertdev: we are now. Github had some trouble with updating the file it seems

ntp-proto/src/filter.rs
folkertdev: yeah, it says

```
        /*
         * Combine the offsets of the clustering algorithm survivors
         * using a weighted average with weight determined by the root
         * distance.  Compute the selection jitter as the weighted RMS
         * difference between the first survivor and the remaining
         * survivors.  In some cases, the inherent clock jitter can be
         * reduced by not using this algorithm, especially when frequent
         * clockhopping is involved.  The reference implementation can
         * be configured to avoid this algorithm by designating a
         * preferred peer.
         */
        y = z = w = 0;
        for (i = 0; s.v[i].p != NULL; i++) {
                p = s.v[i].p;
                x = root_dist(p);
                y += 1 / x;
                z += p->offset / x;
                w += SQUARE(p->offset - s.v[0].p->offset) / x;
        }
        s.offset = z / y;
        s.jitter = SQRT(w / y);
}
```
so we should turn that into ` w += SQUARE(p->jitter - s.v[0].p->jitter) / x;` then?

davidv1992: No, i think the above just recalculates PSI_s, and we should afterwards combine that with PSI_p per the formula on page 43 to get the total jitter.

folkertdev: most likely, we're both clicking on a link in this comment to view the code, and that takes you to the now-outdated version. Very annoying... So you have to navigate via "files changed"

davidv1992: Sorry, with the above I meant the excerpt of c code you posted. From how I understand it, that just recalculates the selection jitter in w (or at least the formulas seem very similar). From what I understand, PSI_p is just the peer jitter from the peer that we selected as the system peer, that is, the peer which came first in the list after clock_select.

The alternative you suggest (w += SQUARE(p->jitter - s.v[0].p->jitter) / x) just seems downright weird to me, as that is just taking the squared difference of two jitters, which I am not sure what that is. (but probably not a jitter)

ntp-proto/src/filter.rs
davidv1992: Ah yes, thought we still used some of it in checkign whether a peer is acceptable, but was mistaken on that.

ntp-proto/src/filter.rs
folkertdev: yes, not sure why but the skeleton decides to calculate just PSI_p in that function. The current code of this PR calculates the full system jitter (PSI) now.

ntp-proto/src/filter.rs
djc: FWIW, this might be a good use case for wrapping a newtype around it that can be used to assert it's sorted?

PR #34, opened 2022-04-14
Add proper support for loop detection.

davidv1992:
This also needed to refactor how we deal with reference ids a bit.

### DIFF
--- a/ntp-proto/src/identifiers.rs
+++ b/ntp-proto/src/identifiers.rs
@@ -1,0 +1,88 @@
+use std::net::IpAddr;
+
+use md5::{Digest, Md5};
+
+#[derive(Debug, Copy, Clone, PartialEq, Eq)]
+pub struct ReferenceId(u32);
+
+impl ReferenceId {
+    // Note: Names chosen to match the identifiers given in rfc5905
+    const KISS_DENY: u32 = 0x44454E59;
+    const KISS_RATE: u32 = 0x52415445;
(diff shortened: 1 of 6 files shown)


ntp-proto/src/identifiers.rs
```diff
@@ -0,0 +1,88 @@
...
+        self.0 == ReferenceId::KISS_DENY
+    }
+
+    pub(crate) fn is_rate(&self) -> bool {
+        self.0 == ReferenceId::KISS_RATE
+    }
+
+    pub(crate) fn is_rstr(&self) -> bool {
+        self.0 == ReferenceId::KISS_RSTR
+    }
```
folkertdev, line 37: use `Self::CONSTANT_NAME` here?

ntp-proto/src/identifiers.rs
```diff
@@ -0,0 +1,88 @@
...
+
+    pub(crate) fn to_bits(self) -> [u8; 4] {
```
folkertdev, line 39: I'd name this `from_bytes`

ntp-proto/src/identifiers.rs
davidv1992: done

ntp-proto/src/identifiers.rs
davidv1992: done

folkertdev (approved):

PR #35, opened 2022-04-14
System state

folkertdev:


ntp-proto/src/system.rs
```diff
@@ -0,0 +1,127 @@
...
+    ) {
+        let survivors = match clock_select(peers, local_clock_time, system_poll) {
+            None => return,
+            Some(v) => v,
+        };
+
+        // TODO, in the implementation (but not the spec from what I've seen)
+        // the current system peer is kept if it is a survivor and it's stratum
+        // is equal to the best (i.e. index 0) of the survivors
+        let p = survivors[0].peer;
```
folkertdev, line 44: thoughts here?

ntp-proto/src/system.rs
```diff
@@ -0,0 +1,127 @@
...
+        self.jitter = combined.jitter;
+
+        match local_clock(p, self.offset) {
+            LocalClockCode::Panic => {
+                // The offset is too large and probably bogus.  Complain to the
+                // system log and order the operator to set the clock manually
+                // within PANIC range.  The reference implementation includes a
+                // command line option to disable this check and to change the
+                // panic threshold from the default 1000 s as required.
+                panic!("fatal error");
```
folkertdev, line 63: the skeleton just does `exit(0)` here. I think panicking is the right option here (we already panic in some cases where the state is just invalid)

ntp-proto/src/system.rs
```diff
@@ -0,0 +1,127 @@
...
+            LocalClockCode::Step => {
+                // The offset is more than the step threshold (0.125 s by
+                // default).  After a step, all associations now have
+                // inconsistent time values, so they are reset and started
+                // fresh.  The step threshold can be changed in the reference
+                // implementation in order to lessen the chance the clock might
+                // be stepped backwards.  However, there may be serious
+                // consequences, as noted in the white papers at the NTP project site.
+
+                // TODO: reset all associations
```
folkertdev, line 74: this one is a bit weird, and needs us to think about ephemeral and persistent associations.

ntp-proto/src/system.rs
```diff
@@ -0,0 +1,127 @@
...
+enum LocalClockCode {
+    Ignore = 0,
+    Slew = 1,
+    Step = 2,
+    Panic = 3,
+}
+
+fn local_clock(_peer: &Peer, _system_offset: NtpDuration) -> LocalClockCode {
+    todo!()
+}
```
folkertdev, line 127: @davidv1992 I'm guessing you'll be working on this part?

ntp-proto/src/time_types.rs
```diff
@@ -179,6 +179,13 @@ impl NtpDuration {
         Self { duration }
     }
 
+    #[must_use]
+    pub(crate) const fn abs(self) -> Self {
+        Self {
+            duration: self.duration.abs(),
```
folkertdev, line 185: this seems to be correct

ntp-proto/src/filter.rs
```diff
@@ -9,7 +9,7 @@
 
 use crate::{packet::NtpLeapIndicator, NtpDuration, NtpHeader, NtpTimestamp};
 
-const MAX_STRATUM: u8 = 16;
+pub const MAX_STRATUM: u8 = 16;
```
djc, line 12: I'd consider doing `#![warn(unreachable_pub)]` and being more precise about what is meant here, that makes it a lot easier to understand the level of intended visibility -- that is, I think you really meant `pub(crate)` here?

davidv1992:
Structurally this still feels antithetical to what we want, but I suggest we discuss that next week on thursday.

PR #36, opened 2022-04-15
Update a peer with a new packet

folkertdev:


ntp-proto/src/time_types.rs
```diff
@@ -179,12 +179,34 @@ impl NtpDuration {
...
+        Self::from_seconds(log2d(input))
+    }
+
+    #[must_use]
+    pub(crate) const fn divided_by(self, divisor: i64) -> Self {
+        // No overflow risks for division
+        Self {
+            duration: self.duration / divisor,
+        }
+    }
```
folkertdev, line 192: I needed const division.

ntp-proto/src/time_types.rs
```diff
@@ -179,12 +179,34 @@ impl NtpDuration {
         Self { duration }
     }
 
+    pub(crate) fn from_exponent(input: i8) -> Self {
+        Self::from_seconds(log2d(input))
+    }
```
folkertdev, line 184: so technically the exponents are `i8`s. In practice they are usually `u8`s so we may want a custom function for that (which could be `const`)

ntp-proto/src/time_types.rs
davidv1992: I guess this is not fixable by adding const to the div implementation?

davidv1992: I guess this is not fixable by adding const to the div implementation?

ntp-proto/src/filter.rs
```diff
@@ -299,6 +328,124 @@ impl Peer {
...
+            let delay = BROADCAST_DELAY;
+            let dispersion =
+                packet_precision + system_precision + multiply_by_phi(BROADCAST_DELAY * 2i64);
+
+            FilterTuple {
+                offset,
+                delay,
+                dispersion,
+                time: local_clock_time,
+            }
```
davidv1992, line 384: Lets not deal with broadcast yet.

ntp-proto/src/filter.rs
```diff
@@ -299,6 +328,124 @@ impl Peer {
...
+                self.host_poll,
+                NtpDuration::from_exponent(self.last_packet.poll),
+            );
+            self.next_date = self.out_date + offset;
+        }
+
+        if self.next_date < local_clock_time {
+            self.next_date = local_clock_time + NtpDuration::ONE;
+        }
+    }
```
davidv1992, line 440: I don't quite see what this is supposed to do?

ntp-proto/src/filter.rs
folkertdev: finding all the places where we do need to add it later will be a nightmare. When the cost is small, I figure we can just do the implementation. Or leave an explicit `todo!()`

ntp-proto/src/time_types.rs
folkertdev: sadly not, traits cannot have const methods yet.

PR #38, opened 2022-04-19
Added fuzzing for the interval finding algorithm.

davidv1992:


ntp-proto/src/filter.rs
```diff
@@ -451,21 +451,54 @@ fn find_interval(chime_list: &[CandidateTuple]) -> Option<(NtpDuration, NtpDurat
...
+        });
+        candidates.push(CandidateTuple {
+            peer: &peers[i],
+            endpoint_type: EndpointType::Upper,
+            edge: NtpDuration::from_fixed_int((*center).wrapping_add(size as i64)),
+        });
+    }
+    candidates.sort_by(|a, b| a.edge.cmp(&b.edge));
+    let survivors = construct_survivors(&candidates, NtpTimestamp::from_fixed_int(0));
+    assert!(survivors.len() == 0 || 2 * survivors.len() > spec.len());
```
folkertdev, line 495: what property is this testing exactly?  Looks like either there are no survivors, or the number of survivors is double the number of inputs (peers?). That is not intuitive

PR #41, opened 2022-04-21
add a type for reachability

folkertdev:
effectively a cherry-pick from #36 

The `Reach` type is currently unused in the code, but it is tested.

The naming is taken from the spec, but I'm not totally happy with it. But `reachability` in full also doesn't really work (`peer.reachability.received_packet()` just doesn't look right to me). So, suggestions welcome I guess, but `reach` at least mirrors the spec and it's not terrible


davidv1992 (approved):
Looks good to me, can be merged

PR #42, opened 2022-04-21
Create FilterTuple from a packet

folkertdev:
Logic to convert a new packet into a FilterTuple, the starting point of our clock update process

### DIFF
--- a/ntp-proto/src/filter.rs
+++ b/ntp-proto/src/filter.rs
@@ -7,7 +7,10 @@
 //
 //      https://datatracker.ietf.org/doc/html/rfc5905#appendix-A.5.2
 
-use crate::{packet::NtpLeapIndicator, NtpDuration, NtpHeader, NtpTimestamp, ReferenceId};
+use crate::{
+    packet::{NtpAssociationMode, NtpLeapIndicator},
+    NtpDuration, NtpHeader, NtpTimestamp, ReferenceId,
+};
 
 const MAX_STRATUM: u8 = 16;
 const MAX_DISTANCE: NtpDuration = NtpDuration::ONE;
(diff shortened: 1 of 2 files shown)


ntp-proto/src/time_types.rs
```diff
@@ -180,6 +180,19 @@ impl NtpDuration {
...
 
+    /// Interpret an exponent `k` as `2^k` seconds, expressed as an NtpDuration
+    pub(crate) fn from_exponent(input: i8) -> Self {
+        let seconds = if input < 0 {
+            // 1. / (1L << -(a))
+            1.0 / (1i64 << -input) as f64
+        } else {
+            // 1L << (a)
+            (1i64 << input) as f64
+        };
```
djc, line 191: Style nit: I'd write this as:

```rust
Self::from_seconds(match input {
    // 1L << (a)
    0.. => (1i64 << input) as f64,
    // 1. / (1L << -(a))
    _ => 1.0 / (1i64 << -input) as f64,
})
```

folkertdev: actually, do we expect problems with just using 

```
Self::from_seconds(2.0f64.powi(input as i32))
```

My tests don't see any precision loss (makes sense given it's a power of 2) and this code expresses what we're trying to achieve much better than the bitshifting.

djc: Hah, that looks much nicer, and I think it makes sense.

ntp-proto/src/filter.rs
```diff
@@ -58,6 +61,52 @@ impl FilterTuple {
...
+        // | xmt       | T3         | transmit timestamp    |
+        // | dst       | T4         | destination timestamp |
+
+        // for a broadcast association, different logic is used
+        debug_assert_ne!(packet.mode, NtpAssociationMode::Broadcast);
+
+        let packet_precision = NtpDuration::from_exponent(packet.precision);
+
+        // offset is the average of the deltas (T2 - T1) and (T4 - T3)
+        let offset1 = packet.receive_timestamp - packet.origin_timestamp;
```
davidv1992, line 89: We should probably not be using origin timestamp from the packet, but instead keep our sending time locally and using that. This would allow us to randomize origin_timestamp more in the future to make attacks harder, and to use more precise timestamping once we have kernel timestamping working.

folkertdev: we could update the `origin_timestamp` field of the packet. That would be less error-prone (the packet is stored in the peer, so it's very easy to (mis)use the packet's origin timestamp in the logic.) Thoughts?

davidv1992: Fair point, that seems reasonable.

PR #43, opened 2022-04-21
reorganize filter.rs into more files

folkertdev:
`pub(crate)` all the things. 

This seems to generally work well. The one thing I'm not entirely sure on is where `PeerStatistics` should live. I know that in rust you can have circular dependencies between modules, but intuitively I dislike that. Right now there is still a circular dependency between `peer.rs` and `filter.rs`


djc:
Maybe keep `PeerStatistics` directly in the crate root for now?

PR #47, opened 2022-04-26
Made duration::from_exponent use integer math.

davidv1992:


ntp-proto/src/time_types.rs
```diff
@@ -182,7 +182,17 @@ impl NtpDuration {
 
     /// Interpret an exponent `k` as `2^k` seconds, expressed as an NtpDuration
     pub(crate) fn from_exponent(input: i8) -> Self {
-        Self::from_seconds(2.0f64.powi(input as i32))
+        Self {
+            duration: if input > 30 {
```
djc, line 186: I will again suggest writing this as a `match` expression, which seems to end up more readable.

folkertdev (approved):

ntp-proto/src/time_types.rs
```diff
@@ -183,14 +183,11 @@ impl NtpDuration {
...
-            duration: if input > 30 {
-                std::i64::MAX
-            } else if input > 0 {
-                0x1_0000_0000_i64 << input
-            } else if input < -32 {
-                0
-            } else {
-                0x1_0000_0000_i64 >> (-input)
+            duration: match input {
+                exp if exp > 30 => std::i64::MAX,
```
djc, line 187: Note, I think you can do `match input  { 0..=30 => .. }`, which would make this more concise still.

PR #49, opened 2022-04-26
Ensure conversion from float to duration saturates properly.

davidv1992:


ntp-proto/src/time_types.rs
```diff
@@ -164,7 +164,15 @@ impl NtpDuration {
...
-        let duration = (i as i64) << 32 | (f * u32::MAX as f64) as i64;
+        // Ensure proper saturating behaviour
+        let duration = match i as i64 {
+            i if i >= std::i32::MIN as i64 && i <= std::i32::MAX as i64 => {
+                (i << 32) | (f * u32::MAX as f64) as i64
+            }
+            i if i < std::i32::MIN as i64 => std::i64::MIN,
+            i if i > std::i32::MAX as i64 => std::i64::MAX,
+            _ => unreachable!(),
+        };
```
folkertdev, line 175: shame that `foo..` patterns don't work yet

folkertdev (approved):

ntp-proto/src/time_types.rs
djc: Works for me: https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=4ba5a068b3f3598168190e296c6e3ab0

djc: (Not sure why all the `std::` prefixes here, I'm pretty sure those are unnecessary.

Issue #14, opened 2022-04-05
Implement kernel-level software timestamping


davidv1992:
We need some way to have software timestamping work with tokio in a proper way. Would prefer to have thin unsafe wrappers for the system calls in separate crates, prefer also to have few dependencies (this shouldn't be too much code, and it is probably better to own it ourselves than have a dependency on something like nix, libc is acceptable though in my view)

rnijveld:
I've been looking at a way to get this implemented.

Setting the right flags shouldn't be much trouble using the `RawFd` and a few basic libc calls, but reading the ancillary messages for the actual time stamp data is a bigger issue. Tokio doesn't implement it right now, although there have been some mentions (e.g. https://github.com/tokio-rs/tokio/issues/799). The most recent mention I could find (https://github.com/tokio-rs/tokio/issues/2975) suggests waiting for an std implementation (https://github.com/rust-lang/rust/issues/76915), but it looks like that will still take quite a lot of work to get somewhere.

I think our current best bet is to go with implementing our own `recvmsg` with [`AsyncFd`](https://docs.rs/tokio/latest/tokio/io/unix/struct.AsyncFd.html) to detect socket reading readiness. That would require us implementing our own parsing of ancillary control messages (for examples look at [std](https://github.com/rust-lang/rust/blob/master/library/std/src/os/unix/net/ancillary.rs) or the [recvmsg code in nix](https://github.com/nix-rust/nix/blob/c59a8c8cc6cc6922888d6f33635905f9ea3d3205/src/sys/socket/mod.rs#L1552)). That seems doable given our limited use-case, although it will require a good chunk of unsafe code because of all the pointer arithmetic involved.

PR #52, opened 2022-04-28
Initial networking stack with software timestamping

rnijveld:
Right now the receive timestamp is returned via an output parameter. I think it could be made to return as a return value just as easily, but this might feel more natural in places where a normal recv could also be used.

See #14


ntp-udp/src/lib.rs
```diff
@@ -0,0 +1,128 @@
...
+    mhdr.msg_controllen = control_buf.len();
+    mhdr.msg_iov = buf as *mut _ as *mut _;
+    mhdr.msg_iovlen = 1;
+    mhdr.msg_flags = 0;
+}
+
+fn recv(
+    socket: &std::net::UdpSocket,
+    buf: &mut [u8],
+    recv_ts: &mut Option<NtpTimestamp>,
```
djc: Using a `&mut Option` as an outparam seems pretty unidiomatic. I think it would definitely be better to return a tuple (why do we think callers will be interested in the length of the read?).

ntp-daemon/src/main.rs
```diff
@@ -24,7 +24,7 @@ async fn setup_connection() -> std::io::Result<UdpSocket> {
 
 #[tokio::main]
 async fn main() -> Result<(), Box<dyn Error>> {
-    let socket = setup_connection().await?;
+    let socket = ntp_udp::UdpSocket::from_tokio(setup_connection().await?)?;
```
djc, line 27: Why not add a basic constructor to `ntp_udp::UdpSocket` that takes the source and destination addresses, and move that functionality out of this crate's `setup_connection()`? That seems a simpler design. (I would probably skip having the `from_std()` and `from_tokio()` constructors unless they add value for testing.)

ntp-udp/src/lib.rs
rnijveld: Definitely! So the final version already changed that! The length of the read would mostly be important to ensure that the buffer was filled with the expected number of bytes and not fewer than expected.

PR #53, opened 2022-04-28
have peer produce data for system update

folkertdev:
have peer produce the minimal data the system needs to process an update of the peer


ntp-proto/src/peer.rs
```diff
@@ -65,9 +65,17 @@ impl Reach {
     }
 }
 
-pub enum Decision {
-    Ignore,
-    Process,
+#[allow(dead_code)]
+pub(crate) enum MsgForSystem {
```
djc, line 69: Why not just use Option for this?

PR #55, opened 2022-04-28
Initial implementation of peer process.

davidv1992:


ntp-daemon/src/peer.rs
```diff
@@ -0,0 +1,76 @@
...
+use tokio::{
+    net::{ToSocketAddrs, UdpSocket},
+    sync::watch,
+    time::Instant,
+};
+
+fn poll_interval_to_duration(poll_interval: i8) -> Duration {
+    match poll_interval {
+        i if i >= 0 => Duration::from_secs(1),
+        i if i < 64 => Duration::from_secs(1 << i),
```
rnijveld, line 13: It would seem this is only reachable by negative numbers because of the previous match arm, is that right?

PR #68, opened 2022-04-28
Implemented basic multi-peer handling.

davidv1992:


ntp-daemon/src/main.rs
```diff
@@ -1,17 +1,53 @@
...
+use ntp_proto::{filter_and_combine, NtpClock, NtpDuration};
 use peer::start_peer;
 use std::error::Error;
 
 #[tokio::main]
 async fn main() -> Result<(), Box<dyn Error>> {
     let clock = UnixNtpClock::new();
-    let mut channel = start_peer("216.239.35.4:123", clock).await.unwrap();
+    let mut peers = vec![
+        start_peer("0.pool.ntp.org:123", UnixNtpClock::new())
```
djc, line 14: Would be nice to deduplicate this with a loop.

ntp-daemon/src/main.rs
```diff
@@ -1,17 +1,53 @@
...
 
     loop {
-        channel.changed().await.unwrap();
-        println!("{:?}", channel.borrow_and_update().as_ref())
+        let i = {
+            let mut changed: FuturesUnordered<_> = peers
+                .iter_mut()
+                .enumerate()
+                .map(|(i, c)| async move {
+                    c.changed().await.unwrap();
```
djc, line 34: Is unwrapping okay here? Why? Add a comment?

PR #69, opened 2022-04-28
Proto cleanup

folkertdev:


ntp-daemon/src/peer.rs
```diff
@@ -47,21 +47,16 @@ pub async fn start_peer<A: ToSocketAddrs, C: 'static + NtpClock + Send>(
...
                 result = socket.recv(&mut buf) => {
                     // Note: packets are allowed to be bigger when including extensions.
-                    // we don't expect them, but the server may still send them. They
-                    // are guaranteed safe to ignore.
-                    if let Ok((size, Some(timestamp))) = result {
-                        if size >= 48 {
-                            let packet = NtpHeader::deserialize(&buf);
+                    // we don't expect them, but the server may still send them. The
+                    // extra bytes are guaranteed safe to ignore. `recv` truncats the messages
+                    if let Ok((_, Some(timestamp))) = result {
```
djc, line 52: Switch the order so we can early continue? Can also pull the `tx.send()` out of the `if` expression.

ntp-proto/src/peer.rs
```diff
@@ -51,44 +50,43 @@ pub struct Peer {
...
-pub enum MsgForSystem {
-    NoUpdate,
-    PeerUpdated(PeerUpdated),
+pub enum PeerError {
+    /// The association mode is not one that this peer supports
+    InvalidMode,
+    /// The send time on the received packet is not the time we sent it at
+    InvalidPacketTime,
+    /// Peer is in an invalid state
+    Kiss,
```
davidv1992, line 78: Not all of these represent error conditions perse, so calling this an error is a bit weird.

ntp-proto/src/peer.rs
```diff
@@ -219,19 +217,19 @@ impl Peer {
         );
 
         match updated {
-            None => MsgForSystem::NoUpdate,
+            None => Err(PeerError::InvalidPacketTime),
```
davidv1992, line 220: This is not really the correct condition to give back, as it isn't that we got packets out of order here, just that the most recent packet isn't (yet) good enough to consider for the rest of the algorithms.

ntp-daemon/src/peer.rs
davidv1992: I am not sure what you suggest here with switch the order. Also, pulling out the tx.send out of the two if expressions at the end is not really an option, as they don't have an else branch (there is an implicit no-update else there, which semantically matters because calling tx.send itself is the trigger for further updates which are not always needed)

davidv1992 (approved):

PR #70, opened 2022-04-29
Update peer with new system variables

folkertdev:


ntp-daemon/src/main.rs
```diff
@@ -58,5 +62,9 @@ async fn main() -> Result<(), Box<dyn Error>> {
             }
             None => println!("filter and combine did not produce a result"),
         }
+
+        // TODO produce an updated snapshot
+        let system_snapshot = SystemSnapshot::default();
+        system_tx.send(system_snapshot)?;
```
folkertdev, line 68: this depends on the actual clock update, which isn't implemented yet.

ntp-daemon/src/peer.rs
```diff
@@ -55,9 +69,12 @@ pub async fn start_peer<A: ToSocketAddrs, C: 'static + NtpClock + Send>(
                             // TODO log something
                         } else {
                             let packet = NtpHeader::deserialize(&buf);
-                            let result = peer.handle_incoming(packet, timestamp);
 
-                            if peer.accept_synchronization(timestamp, NtpDuration::ZERO).is_err() {
+                            let system_snapshot = *system_snapshots.borrow_and_update();
```
folkertdev, line 73: right now I'm waiting with reading the `system_snapshots` till the last moment. Does that actually matter. If not we could refactor a little bit

ntp-proto/src/peer.rs
```diff
@@ -124,11 +144,10 @@ pub enum AcceptSynchronizationError {
...
 
 impl Peer {
-    pub fn new(
-        our_id: ReferenceId,
-        peer_id: ReferenceId,
-        current_system_time: NtpTimestamp,
-    ) -> Self {
+    pub fn new(our_id: ReferenceId, peer_id: ReferenceId, local_clock_time: NtpTimestamp) -> Self {
+        // we initialize with the current time so that we're in the correct epoch.
+        let time = local_clock_time;
```
folkertdev, line 149: the specification does not initialize the time (it is set before it is read), but using `0` caused issues for us right? because the logic turned out to put us in the wrong epoch.

ntp-proto/src/peer.rs
```diff
@@ -192,15 +212,10 @@ impl Peer {
...
 
-            // TODO: properly fill in system parameters
-            let filter_input = FilterTuple::from_packet_default(
-                &message,
-                NtpDuration::from_seconds(0.0),
-                recv_time,
-                recv_time,
-            );
+            let filter_input =
+                FilterTuple::from_packet_default(&message, system.precision, recv_time, recv_time);
```
folkertdev, line 216: just checking: the repeated `recv_time` is correct here? the spec uses `c.t` for the second one (it is the timestamp that ends up in the filtertuple).

davidv1992: Note: There are issues here, these are preexisting, we will make an issue to address the second timescale this currently uses implicitly

PR #80, opened 2022-05-09
add and use PollInterval type (taken from clock-adjust branch)

folkertdev:


ntp-proto/src/peer.rs
```diff
@@ -152,9 +152,9 @@ impl Peer {
         let time = local_clock_time;
 
         Self {
-            last_poll_interval: 2,
-            next_poll_interval: 2,
-            remote_min_poll_interval: 2,
+            last_poll_interval: PollInterval::MIN,
+            next_poll_interval: PollInterval::MIN,
+            remote_min_poll_interval: PollInterval::MIN,
```
folkertdev, line 157: see below for discussion on what `PollInterval::MIN` should be. I'm not sure where these defaults come from, but `MIN` is , at least at the moment, not `2`

ntp-proto/src/time_types.rs
```diff
@@ -381,6 +382,47 @@ ntp_duration_scalar_div!(u16);
 ntp_duration_scalar_div!(u32);
 // u64 and usize deliberately excluded as they can result in overflows
 
+#[derive(Debug, Copy, Clone, PartialEq, Eq, PartialOrd, Ord)]
+pub struct PollInterval(i8);
```
djc, line 386: Could use a little documentation on what the inner value means.

ntp-proto/src/time_types.rs
```diff
@@ -381,6 +382,47 @@ ntp_duration_scalar_div!(u16);
...
+
+impl PollInterval {
+    pub const MIN: Self = Self(4);
+    pub const MAX: Self = Self(17);
```
folkertdev, line 390: The situation here is interesting. The spec says (section 7.2):

    | MINPOLL   | 4     | minimum poll exponent (16 s)     |
    | MAXPOLL   | 17    | maximum poll exponent (36 h)     |

The code skeleton says

    #define MINPOLL         6       /* % minimum poll interval (64 s)*/
    #define MAXPOLL         17      /* % maximum poll interval (36.4 h) */

finally, the ntp repository 

    #define	NTP_MINPOLL	3	/* log2 min poll interval (8 s) */
    #define NTP_MINDPOLL	6	/* log2 default min poll (64 s) */
    #define NTP_MAXDPOLL	10	/* log2 default max poll (~17 m) */
    #define	NTP_MAXPOLL	17	/* log2 max poll interval (~36 h) */

So nobody agrees. Guess we follow the spec literally here?

ntp-proto/src/time_types.rs
```diff
@@ -381,6 +382,47 @@ ntp_duration_scalar_div!(u16);
...
+
+    pub fn inc(&mut self) {
+        *self = Self(self.0 + 1).min(Self::MAX);
+    }
+
+    #[must_use]
```
djc, line 396: So far `inc()` and `dec()` only appear to be used in tests, whereas `plus_one()` is actually used in the library. Will that change? I would probably only have copying `inc()` and `dec()` and ditch `plus_one()`; having both seems a little weird and modifying by `&mut` isn't actually that valuable for a type that's smaller than a word.

djc (commented):
I think this is a really nice change! Some nits below.

ntp-proto/src/time_types.rs
folkertdev: I pulled this from branch, but turns out inc/dec also were not used there. I've changed the semantics of inc/dec to take/return by-value

ntp-proto/src/time_types.rs
davidv1992: Indeed, decided to go with the main (normative) part of the spec here. We can always decide to deviate here should we want to.

ntp-proto/src/peer.rs
davidv1992: I needed some initial values, and these seemed appropriate. It seems the standard doesn't specify these

ntp-proto/src/time_types.rs
```diff
@@ -381,6 +382,50 @@ ntp_duration_scalar_div!(u16);
...
+    pub fn inc(self) -> Self {
+        Self(self.0 + 1).min(Self::MAX)
+    }
+
+    #[must_use]
+    pub fn dec(self) -> Self {
+        Self(self.0 - 1).max(Self::MIN)
+    }
```
davidv1992, line 406: Not sure I agree with the change in interface here. This is going to result in a lot of code following the pattern `a=a.inc()`, which is why I originally decided on modifying inc/dec

PR #81, opened 2022-05-10
System peer

folkertdev:
Select the system peer and transport its variables to the system task. They will be part of the input for clock adjustment.

### DIFF
--- a/ntp-proto/src/clock_select.rs
+++ b/ntp-proto/src/clock_select.rs
@@ -1,6 +1,6 @@
-use crate::peer::{PeerSnapshot, MAX_DISTANCE};
+use crate::peer::{PeerSnapshot, SystemPeerVariables, MAX_DISTANCE};
 use crate::time_types::NtpInstant;
-use crate::{NtpDuration, PollInterval, ReferenceId};
+use crate::{NtpDuration, PollInterval};
 
 // TODO this should be 4 in production?!
 /// Minimum number of survivors needed to be able to discipline the system clock.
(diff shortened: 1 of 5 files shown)


ntp-proto/src/clock_select.rs
```diff
@@ -23,20 +23,43 @@ const MIN_INTERSECTION_SURVIVORS: usize = 1;
...
 const MIN_CLUSTER_SURVIVORS: usize = 3;
 
+#[derive(Debug)]
+pub struct FilterAndCombine {
+    pub system_offset: NtpDuration,
+    pub system_jitter: NtpDuration,
+    pub system_peer_variables: SystemPeerVariables,
+}
+
 pub fn filter_and_combine(
```
djc, line 33: Maybe rename this and move it into an `impl FilterAndCombine` block?

ntp-proto/src/clock_select.rs
```diff
@@ -23,20 +23,43 @@ const MIN_INTERSECTION_SURVIVORS: usize = 1;
...
     );
 
-    Some(combined)
+    let filter_and_combine = FilterAndCombine {
+        system_offset: combined.system_offset,
+        system_jitter: combined.system_jitter,
+        system_peer_variables: new_system_peer,
+    };
+
+    Some(filter_and_combine)
```
djc, line 62: Just do `Some(FilterAndCombine { .. })`? There's no point in having the variable binding.

folkertdev: I think that looks weird. Rust is normally very "chain" oriented: we prefer `a.map(..).filter(..)` over `filter(map(a, ...), ...)`. But when constructors are involved it's suddenly OK to nest expressions deeply.

anyhow, not here to fight over style, it's just something about rust's formatting that I think is weird.

ntp-proto/src/peer.rs
```diff
@@ -96,12 +96,21 @@ pub enum IgnoreReason {
     TooOld,
 }
 
+#[allow(dead_code)]
```
davidv1992, line 99: What is dead here that requires this? Perhaps we could limit this to a single field, which also helps identify when we can remove it better.

davidv1992 (commented):
Mostly looks good to me, one minor issue around an allow(dead_code) that I think we should resolve before merging.

Issue #60, opened 2022-04-28
Logging


davidv1992:
Setup basics for logging and add some logging in relevant places.

PR #82, opened 2022-05-11
Add logging for UDP layer

rnijveld:
Related to #60

### DIFF
--- a/ntp-daemon/Cargo.toml
+++ b/ntp-daemon/Cargo.toml
@@ -12,5 +12,6 @@ futures = "0.3.21"
 ntp-proto = { path = "../ntp-proto" }
 ntp-os-clock = { path = "../ntp-os-clock" }
 ntp-udp = { path = "../ntp-udp" }
-env_logger = "0.9.0"
 log = "0.4.17"
+tracing = "0.1.34"
+tracing-subscriber = "0.3.11"
--- a/ntp-daemon/Cargo.toml
+++ b/ntp-daemon/Cargo.toml
@@ -12,3 +12,5 @@ futures = "0.3.21"
 ntp-proto = { path = "../ntp-proto" }
 ntp-os-clock = { path = "../ntp-os-clock" }
 ntp-udp = { path = "../ntp-udp" }
+env_logger = "0.9.0"
+log = "0.4.17"
--- a/ntp-udp/src/lib.rs
+++ b/ntp-udp/src/lib.rs
@@ -4,7 +4,7 @@ use std::{
     os::unix::prelude::AsRawFd,
 };
 
-use log::{debug, trace, warn};
+use tracing::{debug, trace, warn, instrument};
 use ntp_proto::NtpTimestamp;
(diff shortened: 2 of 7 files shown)


ntp-udp/src/lib.rs
```diff
@@ -38,8 +41,33 @@ impl UdpSocket {
...
+                    if log::log_enabled!(log::Level::Trace) {
+                        if let Ok(size) = &result {
+                            trace!(
+                                "sent {} bytes from {:?} to peer {:?}",
+                                size,
+                                self.as_ref().local_addr(),
+                                self.as_ref().peer_addr()
+                            );
+                        }
+                    }
```
folkertdev, line 60: we can do a proper match here, dead-code elimination will just remove that if the log level is such that nothing happens in a branch

```rust
match &result {
    Err(e) => { 
        debug!("error sending data: {}", e),
    }
    Ok(size) => {   
        trace!(
            "sent {} bytes from {:?} to peer {:?}",
            size,
            self.as_ref().local_addr(),
            self.as_ref().peer_addr()
        );
    }
}
```

ntp-udp/src/lib.rs
```diff
@@ -48,13 +76,43 @@ impl UdpSocket {
...
+                    if log::log_enabled!(log::Level::Trace) {
+                        if let Ok((size, ts)) = &result {
+                            trace!(
+                                "received message of size {} with receive ts {:?} from {:?}",
+                                size,
+                                ts,
+                                self.as_ref().local_addr()
+                            );
+                        }
+                    }
```
folkertdev, line 95: same thing just writing out the match is fine, DCE will handle it

ntp-udp/src/lib.rs
rnijveld: Are you sure? The log macros shouldn't be removed as dead code, because the log level is determined at runtime

ntp-udp/src/lib.rs
```diff
@@ -48,13 +76,43 @@ impl UdpSocket {
         loop {
             let mut guard = self.io.readable().await?;
             match guard.try_io(|inner| recv(inner.get_ref(), buf)) {
-                Ok(result) => return result,
-                Err(_would_block) => continue,
+                Ok(result) => {
```
djc, line 79: This would be easier to follow if you did an early continue for the `Err` case, reducing rightward drift for all the code in the `Ok` case.

ntp-daemon/Cargo.toml
```diff
@@ -12,3 +12,5 @@ futures = "0.3.21"
 ntp-proto = { path = "../ntp-proto" }
 ntp-os-clock = { path = "../ntp-os-clock" }
 ntp-udp = { path = "../ntp-udp" }
+env_logger = "0.9.0"
+log = "0.4.17"
```
djc, line 16: I think we should use tracing instead of log, this is much more powerful in particular for async code.

rnijveld: Looks way better indeed!

PR #83, opened 2022-05-12
Peer reset infrastructure

folkertdev:
sets up the communication channels and logic for the reset of all peers. It does not yet have logic for triggering a reset, because that depends on the clock adjustment

### DIFF
--- a/ntp-daemon/src/main.rs
+++ b/ntp-daemon/src/main.rs
@@ -6,7 +6,7 @@ use ntp_os_clock::UnixNtpClock;
 use ntp_proto::{
     FilterAndCombine, NtpClock, NtpInstant, PollInterval, SystemConfig, SystemSnapshot,
 };
-use peer::start_peer;
+use peer::{start_peer, PeerChannels};
 use std::error::Error;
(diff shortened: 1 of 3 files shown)


ntp-daemon/src/main.rs
```diff
@@ -28,70 +40,96 @@ async fn main() -> Result<(), Box<dyn Error>> {
 
     let mut snapshots = Vec::with_capacity(peers.len());
 
+    // when we perform a clock jump, all current measurement data is off. We force all associations
+    // to clear their measurement data and get new data. This vector contains associations that
+    // have not yet confirmed that their measurement data has been cleared.
+    let mut waiting_for_reset: Vec<PeerChannels> = Vec::with_capacity(peers.len());
```
folkertdev, line 46: so basically  the reset process is to move all peers into this vector `waiting_for_reset.append(&mut peers)` and triggering the reset of the peers `_reset_tx.send(...)`. The rest (resetting the peer, reinstating the peer as giving valid measurements) is already handled in this PR.

davidv1992:
Looks good

PR #84, opened 2022-05-12
System instant

folkertdev:
use the system instant instead of a wrapped ntp timestamp

### DIFF
--- a/ntp-daemon/src/peer.rs
+++ b/ntp-daemon/src/peer.rs
@@ -1,5 +1,5 @@
 use ntp_proto::{
-    IgnoreReason, NtpHeader, NtpInstant, Peer, PeerSnapshot, ReferenceId, SystemConfig,
+    IgnoreReason, NtpClock, NtpHeader, NtpInstant, Peer, PeerSnapshot, ReferenceId, SystemConfig,
     SystemSnapshot,
 };
 use tokio::{
@@ -19,8 +19,9 @@ pub enum MsgForSystem {
     Snapshot(PeerSnapshot),
 }
 
-pub async fn start_peer<A: ToSocketAddrs>(
+pub async fn start_peer<A: ToSocketAddrs, C: 'static + NtpClock + Send>(
     addr: A,
+    clock: C,
     config: SystemConfig,
     mut system_snapshots: watch::Receiver<SystemSnapshot>,
 ) -> Result<watch::Receiver<MsgForSystem>, std::io::Error> {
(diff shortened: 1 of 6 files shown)


ntp-daemon/src/peer.rs
```diff
@@ -59,13 +65,16 @@ pub async fn start_peer<A: ToSocketAddrs, C: 'static + NtpClock + Send>(
                         .as_mut()
                         .reset(Instant::now() + poll_interval.as_system_duration());
 
-                    // TODO: Figure out proper error behaviour here
-                    let ntp_instant = NtpInstant::from_ntp_timestamp(clock.now().unwrap());
+                    let ntp_instant = NtpInstant::now();
                     let packet = peer.generate_poll_message(ntp_instant);
+
+                    // TODO error handling here; what if we cannot determine the origin timestamp?
+                    last_send_timestamp = clock.now().ok();
```
davidv1992, line 72: If we can't properly stamp a send timestamp, that pretty much inhibits any synchronization, so in my view crashing out here with an expect is acceptable. The only other alternative I see is to log the error and not send the packet, but given the type of causes for the clock failing I don't really see this doing much in terms of resilience (it is pretty much guaranteed to always fail if it fails once)

ntp-proto/src/filter.rs
```diff
@@ -118,7 +126,8 @@ impl LastMeasurements {
         system_precision: NtpDuration,
         frequency_tolerance: FrequencyTolerance,
     ) -> Option<(PeerStatistics, NtpInstant)> {
-        let dispersion_correction = (new_tuple.time - peer_time) * frequency_tolerance;
+        let dispersion_correction =
+            NtpInstant::abs_diff(new_tuple.time, peer_time) * frequency_tolerance;
```
davidv1992, line 130: This is a mild change of behaviour compared to the old situation, although new_tuple.time should never be before peer_time. However, perhaps we should assert this? (at least in debug mode)

ntp-proto/src/time_types.rs
```diff
@@ -1,47 +1,50 @@
...
+    pub fn now() -> Self {
         Self {
-            timestamp: NtpTimestamp::from_bits(bits),
+            instant: Instant::now(),
         }
     }
 
-    pub(crate) const fn to_bits(self) -> [u8; 8] {
-        self.timestamp.to_bits()
+    // used to populate the T1 and T3 fields to ensure the server is doing something sensible
```
davidv1992, line 19: I think this is (or at least should) only ever be used for the T1 timestamp. T3 isn't even ours to control as that is the timestamp the server indicated it sent its packet at.

ntp-proto/src/time_types.rs
```diff
@@ -1,47 +1,50 @@
...
+    type Output = NtpInstant;
 
-    fn sub(self, rhs: Self) -> Self::Output {
-        self.timestamp - rhs.timestamp
+    fn add(mut self, rhs: Duration) -> Self::Output {
+        self.instant += rhs;
+
+        self
     }
 }
```
davidv1992, line 49: I think we only ever need this in testing code, so we should probably mark it with the appropriate cfg marker.

davidv1992 (commented):
Very nice PR, left a few minor comments.

ntp-proto/src/time_types.rs
folkertdev: exactly, so we populate it, and then we check that the server has overwritten what we wrote

davidv1992: Sorry, but I think I am missing the point here. The server doesn't write our memory right, so I am not getting what you are implying here

PR #86, opened 2022-05-13
Clock adjustment

davidv1992:


ntp-os-clock/src/lib.rs
```diff
@@ -117,29 +122,74 @@ impl NtpClock for UnixNtpClock {
...
-
-        // gradual adjustment if less than half a second off, otherwise jump
-        if (offset_seconds == -1 && offset_nanos >= 500_000_000)
-            || (offset_seconds == 0 && offset_nanos <= 500_000_000)
-        {
-            ntp_kapi_timex.modes |= libc::ADJ_OFFSET;
-            ntp_kapi_timex.offset =
-                (offset_nanos as i32 + offset_seconds * 1_000_000_000) as libc::c_long;
+        ntp_kapi_timex.modes = libc::MOD_FREQUENCY;
+        ntp_kapi_timex.freq = (freq * 65536e6) as libc::c_long;
```
folkertdev, line 128: what is this magic number?

ntp-os-clock/src/lib.rs
```diff
@@ -117,29 +122,74 @@ impl NtpClock for UnixNtpClock {
...
+        if unsafe { libc::clock_gettime(libc::CLOCK_REALTIME, &mut tp as *mut _) } == -1 {
+            return Err(convert_errno());
+        }
+
+        tp.tv_sec += offset_secs as libc::time_t;
+        tp.tv_nsec += offset_nanos as libc::c_long;
+        while tp.tv_nsec >= 1_000_000_000 {
+            tp.tv_nsec -= 1_000_000_000;
+            tp.tv_sec += 1;
+        }
```
folkertdev, line 158: we expect this to usually do 0 or 1 iterations? when can this happen? maybe add a comment?

folkertdev: because you can make this branchless but maybe there is no point?

ntp-proto/src/clock.rs
```diff
@@ -8,5 +10,397 @@ pub trait NtpClock {
...
+    Panic,
+}
+
+impl<C: NtpClock> ClockController<C> {
+    pub fn new(clock: C) -> Self {
+        clock.set_freq(0.).expect("Unable to set clock frequency");
+        Self {
+            clock,
+            state: ClockState::StartupBlank,
+            last_update_time: NtpInstant::now(),
```
folkertdev, line 63: this default is not technically correct (no update was done now). But that doesn't matter for the logic?

ntp-proto/src/clock.rs
```diff
@@ -8,5 +10,397 @@ pub trait NtpClock {
...
+                        // Filter out short spikes
+                        return ClockUpdateResult::Ignore;
+                    }
+
+                    // Seems that the large difference reflects reality, since
+                    // it persisted for a significant amount of time. So step
+                    // the clock
+                    return self.do_step(offset, last_peer_update);
+                }
+                _ => {
```
folkertdev, line 132: I'm personally a fan of making pattern matches exhaustive when you can

folkertdev: with an or-pattern in this case, presumably

ntp-proto/src/clock.rs
```diff
@@ -8,5 +10,397 @@ pub trait NtpClock {
...
+        //  - Stepping the clock if a large difference persists long
+        //    enough
+        //  - Ensuring a proper initial frequency measurement on startup
+        //  - Making small (gradual) adjustments to the clock when we
+        //    only have a small error
+        if offset.abs() > NtpDuration::STEP_THRESHOLD {
+            // Large spikes are filtered initialy (to handle weird but temporary network issues)
+            // and then handled by stepping if they persist.
+            match self.state {
+                ClockState::Sync => {
```
folkertdev, line 104: `use ClockState::*` would shorten these matches a bit

ntp-proto/src/peer.rs
```diff
@@ -115,10 +115,9 @@ pub struct PeerSnapshot {
     pub(crate) reference_timestamp: NtpTimestamp,
     #[allow(dead_code)]
     pub(crate) poll_interval: PollInterval,
```
folkertdev, line 117: these `dead_code` fields are not needed after all?

ntp-proto/src/clock.rs
```diff
@@ -8,5 +10,397 @@ pub trait NtpClock {
...
+    use core::cell::RefCell;
+    use std::time::Duration;
+
+    #[derive(Debug, Clone, Default)]
+    struct TestClock {
+        last_freq: RefCell<Option<f64>>,
+        last_offset: RefCell<Option<NtpDuration>>,
+    }
+
+    impl NtpClock for TestClock {
```
folkertdev, line 262: this is very nice!

ntp-proto/src/clock.rs
davidv1992: I prefer being explicit with these and not polluting the namespace

ntp-proto/src/clock.rs
davidv1992: Done

ntp-os-clock/src/lib.rs
davidv1992: Anyway, made it an if to indicate intent more clearly

ntp-proto/src/clock.rs
folkertdev: for testing code I think this is totally acceptable

ntp-proto/src/peer.rs
davidv1992: probably a good idea, I will make a PR for that

ntp-os-clock/src/lib.rs
folkertdev: unless I'm missing something
```
tp.tv_sec += ((tp.tv_nsec / 1_000_000_000) == 0) as _;
tp.tv_nsec %= 1_000_000_000;
```

now, I'm not sure it's actually better, but here it is

davidv1992: I dont think it is better, for one thing, it uses modulo and division, which is slow. Anyway, hopefully clang is smart enough to recognize that it can convert the if into a number of cmovs in release builds, which should be the fastest solution.

ntp-proto/src/peer.rs
davidv1992: Removed by #92

PR #91, opened 2022-05-13
stop sending messages to servers that send the DENY code

folkertdev:
we did previously remove a peer that receives DENY from the list of peers, but did not cancel the task that keeps sending messages to the server. Also the ordering of checks was such that the KISS code would never be processed if other fields of the packet are also invalid. But the spec notes that if a KISS code is sent, the rest of the packet may be invalid anyway. So we should check for kiss codes first when validating a packet.

### DIFF
--- a/ntp-daemon/src/system.rs
+++ b/ntp-daemon/src/system.rs
@@ -63,7 +63,9 @@ async fn run(
     let mut snapshots = Vec::with_capacity(peers.len());
 
     while let Some(msg_for_system) = msg_for_system_rx.recv().await {
-        peers.receive_update(msg_for_system, reset_epoch);
+        if let NewMeasurement::No = peers.receive_update(msg_for_system, reset_epoch) {
+            continue;
+        }
 
         // remove snapshots from previous iteration
         snapshots.clear();
@@ -150,6 +152,11 @@ struct Peers {
     peers: Box<[PeerStatus]>,
 }
 
+enum NewMeasurement {
+    Yes,
+    No,
+}
+
 impl Peers {
     fn new(length: usize) -> Self {
         Self {
@@ -170,18 +177,27 @@ impl Peers {
             })
     }
 
-    fn receive_update(&mut self, msg: MsgForSystem, current_reset_epoch: ResetEpoch) {
+    fn receive_update(
+        &mut self,
+        msg: MsgForSystem,
+        current_reset_epoch: ResetEpoch,
+    ) -> NewMeasurement {
         match msg {
             MsgForSystem::MustDemobilize(index) => {
                 self.peers[index.index] = PeerStatus::Demobilized;
+                NewMeasurement::No
             }
             MsgForSystem::Snapshot(index, msg_reset_epoch, snapshot) => {
                 if current_reset_epoch == msg_reset_epoch {
                     self.peers[index.index] = PeerStatus::Valid(snapshot);
+                    NewMeasurement::Yes
+                } else {
+                    NewMeasurement::No
                 }
             }
             MsgForSystem::PeerNotFit(index) => {
                 self.peers[index.index] = PeerStatus::NoMeasurement;
+                NewMeasurement::No
             }
         }
     }


ntp-daemon/src/peer.rs
```diff
@@ -137,27 +135,30 @@ pub async fn start_peer<A: ToSocketAddrs + std::fmt::Debug, C: 'static + NtpCloc
...
-                                config.frequency_tolerance,
-                                config.distance_threshold,
-                                system_poll,
-                            );
-
-                            if accept.is_err() {
-                                let _ = tx.send(MsgForSystem::NoMeasurement);
-                            } else  {
-                                match result {
-                                    Ok(update) => {
```
folkertdev, line 152: the diff here is not great. what I'm doing is inverting the control flow: we now check the `result` before checking whether we should `accept_synchronization`. In particular that means the DENY kiss code is now correctly propagated

ntp-daemon/src/peer.rs
```diff
@@ -137,27 +135,30 @@ pub async fn start_peer<A: ToSocketAddrs + std::fmt::Debug, C: 'static + NtpCloc
...
+                                        system_poll,
+                                    );
+
+                                    if accept.is_err() {
+                                        let _ = tx.send(MsgForSystem::NoMeasurement);
+                                    } else  {
                                         let _ = tx.send(MsgForSystem::Snapshot(update));
                                     }
-                                    Err(IgnoreReason::KissDemobilize) => {
-                                        let _ = tx.send(MsgForSystem::MustDemobilize);
```
folkertdev, line 156: this was unreachable before because a packet with a kiss code would never be accepeted

ntp-proto/src/peer.rs
```diff
@@ -237,12 +237,7 @@ impl Peer {
         // the origin_timestamp changed; the server is not allowed to modify this field
         let origin_changed = Some(message.origin_timestamp) != self.next_expected_origin;
 
-        if message.mode != NtpAssociationMode::Server {
-            // we currently only support a client <-> server association
-            Err(IgnoreReason::InvalidMode)
-        } else if transmit_unchanged || origin_changed {
-            Err(IgnoreReason::InvalidPacketTime)
-        } else if message.is_kiss_rate() {
+        if message.is_kiss_rate() {
```
folkertdev, line 240: messages with a kiss code make no guarantees about the other fields, but we still must respect the kiss codes. So, check the kiss codes first

davidv1992:
The merge conflict here are such that it is hard to currently judge the desired end state. Could you resolve these?

ntp-proto/src/identifiers.rs
```diff
@@ -20,7 +20,7 @@ impl ReferenceId {
         }
     }
 
-    pub(crate) fn from_int(value: u32) -> ReferenceId {
+    pub fn from_int(value: u32) -> ReferenceId {
```
davidv1992, line 23: Given that we make the constants public, why does this then also need to be pub?

ntp-proto/src/identifiers.rs
```diff
@@ -7,9 +7,9 @@ pub struct ReferenceId(u32);
 
 impl ReferenceId {
     // Note: Names chosen to match the identifiers given in rfc5905
-    pub(crate) const KISS_DENY: ReferenceId = ReferenceId(0x44454E59);
-    pub(crate) const KISS_RATE: ReferenceId = ReferenceId(0x52415445);
-    pub(crate) const KISS_RSTR: ReferenceId = ReferenceId(0x52535452);
+    pub const KISS_DENY: Self = ReferenceId(0x44454E59);
+    pub(crate) const KISS_RATE: Self = ReferenceId(0x52415445);
+    pub(crate) const KISS_RSTR: Self = ReferenceId(0x52535452);
```
davidv1992, line 12: I would suggest that if we make one public, we make em all public

README.md
```diff
@@ -6,6 +6,21 @@ This project is intended to become a memory-safe implementation of NTP in Rust.
...
+cargo run
+```
+
+runs the NTP daemon
+
+```
+cargo run --bin demobilize
+```
+
+runs an NTP server on port `8080` that always responds with the DENY kiss code. It is used to test
```
davidv1992, line 21: Nit: I would formulate this as it can be used instead of is used, since we don't use it in automated tests.

davidv1992 (changes requested):
Mostly looks good, and it is important that we start testing these sort of things. However, I am not convinced we should make the test server part of the main ntp-daemon package. I think it is cleaner from an auditability standpoint to isolate these in their own package, despite the ergonomic cost.

ntp-daemon/src/peer.rs
```diff
@@ -142,27 +142,30 @@ where
...
-                Ok(update) => {
-                    info!("packet accepted");
-                    let msg = MsgForSystem::Snapshot(self.index, self.reset_epoch, update);
-                    self.channels.msg_for_system_sender.send(msg).await.ok();
-                }
-                Err(IgnoreReason::KissDemobilize) => {
-                    info!("peer must demobilize");
-                    let msg = MsgForSystem::MustDemobilize(self.index);
-                    self.channels.msg_for_system_sender.send(msg).await.ok();
+        match result {
```
folkertdev, line 145: bit of a crappy diff, but this inverts the matches: now we first check the result, do something with the kiss code if needed, then look at whether the message is accepted.

ntp-proto/src/identifiers.rs
```diff
@@ -7,9 +7,9 @@ pub struct ReferenceId(u32);
 
 impl ReferenceId {
     // Note: Names chosen to match the identifiers given in rfc5905
-    pub(crate) const KISS_DENY: ReferenceId = ReferenceId(0x44454E59);
-    pub(crate) const KISS_RATE: ReferenceId = ReferenceId(0x52415445);
-    pub(crate) const KISS_RSTR: ReferenceId = ReferenceId(0x52535452);
+    pub const KISS_DENY: ReferenceId = ReferenceId(u32::from_be_bytes(*b"DENY"));
+    pub const KISS_RATE: ReferenceId = ReferenceId(u32::from_be_bytes(*b"RATE"));
+    pub const KISS_RSTR: ReferenceId = ReferenceId(u32::from_be_bytes(*b"RSTR"));
```
folkertdev, line 12: makes this much more readable I think

ntp-proto/src/peer.rs
```diff
@@ -222,15 +222,11 @@ impl Peer {
         send_time: NtpTimestamp,
         recv_time: NtpTimestamp,
     ) -> Result<PeerSnapshot, IgnoreReason> {
-        if message.mode != NtpAssociationMode::Server {
-            // we currently only support a client <-> server association
-            warn!("Received packet with invalid mode");
-            Err(IgnoreReason::InvalidMode)
-        } else if message.is_kiss_rate() {
-            warn!("Peer requested rate limit");
+        if message.is_kiss_rate() {
```
folkertdev, line 225: kiss codes must be handled first, because there are no guarantees about any of the other fields in the packet if a kiss code is set.

test-binaries/src/bin/rate-limit-server.rs
```diff
@@ -0,0 +1,45 @@
...
+            packet.reference_id = ReferenceId::KISS_RATE;
+            packet.stratum = 0;
+        } else {
+            let parsed = NtpHeader::deserialize(&buf);
+
+            packet.mode = NtpAssociationMode::Server;
+            packet.stratum = 1;
+            packet.origin_timestamp = parsed.transmit_timestamp;
+            packet.receive_timestamp = ntp_receive;
+            packet.transmit_timestamp = clock.now().unwrap();
```
folkertdev, line 39: NTP server: done

PR #95, opened 2022-05-17
Added a lot of tracing to protocol implementation.

davidv1992:


ntp-proto/src/clock.rs
```diff
@@ -196,17 +203,31 @@ impl<C: NtpClock> ClockController<C> {
...
+        );
+
         // If our preference becomes strong enough, adjust poll interval
         // and reset. The hysteresis here ensures we aren't constantly flip-flopping
         // between different preferred interval lengths.
         if self.poll_interval_counter > Self::POLL_ADJUST {
             self.poll_interval_counter = 0;
             self.preferred_poll_interval = self.preferred_poll_interval.inc();
+            debug!(
+                poll_interval = debug(self.preferred_poll_interval),
```
djc, line 218: You can use `?self.preferred_poll_interval`, which is a good deal more concise than wrapping `debug()` around it.

ntp-proto/src/clock.rs
```diff
@@ -240,6 +262,13 @@ impl<C: NtpClock> ClockController<C> {
     }
 
     fn set_freq(&mut self, offset: NtpDuration, last_peer_update: NtpInstant) {
+        info!(
+            freq = display(
```
djc, line 266: Not sure how it interacts with grouping but you can use a `%` prefix instead of `display()`.

folkertdev (approved):

PR #100, opened 2022-05-19
Switch to SO_TIMESTAMPNS as SO_TIMESTAMPNS_NEW is not defined on all

davidv1992:
platforms.

### DIFF
--- a/ntp-udp/src/lib.rs
+++ b/ntp-udp/src/lib.rs
@@ -105,7 +105,7 @@ fn init_socket(socket: &std::net::UdpSocket) -> io::Result<()> {
         libc::setsockopt(
             fd,
             libc::SOL_SOCKET,
-            libc::SO_TIMESTAMPNS_NEW,
+            libc::SO_TIMESTAMPNS,
             &enable_ts as *const _ as *const libc::c_void,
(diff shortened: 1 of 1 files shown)


davidv1992:
Please ignore the codecov failure for now, we still need to configure some better criteria for it.

folkertdev:
sure, so are there any potential downsides to using this (older?) variant?

davidv1992:
@rnijveld If I remember our conversation correctly there is not really a downside to using the more general SO_TIMESTAMPNS, but could you confirm this?

rnijveld:
@davidv1992 It should be fine, I believe it depends on the size of libc::time_t whether `SO_TIMESTAMPNS` refers to `*_OLD` or `*_NEW`. This does change the struct that is used for the control message, but in practice the resulting struct definition is exactly the same.

rnijveld:
Just for extra reference, see: https://www.kernel.org/doc/Documentation/networking/timestamping.txt

One thing to take care of is that `SO_TIMESTAMPNS_OLD` will return invalid timestamps on 32-bit machines after 2038

davidv1992:
Given the timeline for that I am not too worried about that personally, and suggest we move forward with merging this.

PR #102, opened 2022-05-19
Disable codecov patch coverage check, as we have some code not really

davidv1992:
amenable to coverage testing.


davidv1992:
This is primarily to deal with the fact that we don't want coverage checks to cause failures when modifying platform glue code, not all of which can be tested automatically.

PR #105, opened 2022-05-19
async architecture overhaul

folkertdev:
further refactoring suggestions welcome


ntp-daemon/src/main.rs
```diff
@@ -1,169 +1,165 @@
...
+                info!("filter and combine did not produce a result");
+                continue;
+            }
+        };
+
+        let offset_ms = clock_select.system_offset.to_seconds() * 1000.0;
+        let jitter_ms = clock_select.system_jitter.to_seconds() * 1000.0;
+        info!(offset_ms, jitter_ms, "system offset and jitter");
+
+        let adjust_type = controller.update(
```
djc, line 107: Why not just pass `&clock_select` to `update()`? This looks silly.

ntp-daemon/src/main.rs
```diff
@@ -1,169 +1,165 @@
...
+
+                reset_epoch = reset_epoch.inc();
+                reset_tx.send_replace(reset_epoch);
+            }
+            _ => {}
+        }
+
+        // Handle updating system snapshot
+        match adjust_type {
+            ClockUpdateResult::Ignore => {}
```
djc, line 138: Could do `if let ClockUpdateResult::Ignore = adjust_type { continue; }` here.

ntp-daemon/src/main.rs
```diff
@@ -1,169 +1,165 @@
...
+                global.leap_indicator = clock_select.system_peer_snapshot.leap_indicator;
+            }
+        }
+    }
+
+    // the channel closed and has no more messages in it
+    Ok(())
+}
+
+fn receive_msg_for_system(
```
djc, line 151: I think the `MsgForSystem` name is a little unclear, with the usage of `Msg` being unidiomatic (Rust type names usually avoid abbreviations). I could probably call it `PeerUpdate`. I would also suggest making the outer message type a named field struct so that `sender` (to replace `index`) can be a field you can use without matching first. Could potentially be a `struct PeerUpdate<T> { sender: usize, message: T }` as well if there are multiple kinds of channels (but that doesn't seem to apply here/yet).

Personally I prefer inlining relatively short single-use functions like this, since I find code easier to follow if I don't have to jump around. On the other hand, if you end up defining a `Peers` type as suggested above, this could plausibly be a method on that.

ntp-daemon/src/main.rs
```diff
@@ -1,169 +1,165 @@
...
+        .unwrap();
+    }
 
-    let mut peers = vec![
-        new_peer("0.pool.ntp.org:123").await.unwrap(),
-        new_peer("1.pool.ntp.org:123").await.unwrap(),
-        new_peer("2.pool.ntp.org:123").await.unwrap(),
-        new_peer("3.pool.ntp.org:123").await.unwrap(),
-    ];
+    let mut peers = vec![PeerState::AwaitingReset; peer_addresses.len()];
```
djc, line 78: It feels to me like this would be cleaner more like this:

- Rename `PeerState` to `PeerStatus`
- Define `struct PeerState { status: PeerStatus, last: PeerSnapshot }`
- Define `struct Peers { peers: Vec<PeerState> }`

In general I would encourage making the system's state an explicit `struct`, which I think will make it clearer what state is involved and how that relates to each other.

ntp-daemon/src/peer.rs
```diff
@@ -1,177 +1,272 @@
...
+}
 
 #[derive(Debug, Clone, Copy)]
 pub enum MsgForSystem {
     /// Received a Kiss-o'-Death and must demobilize
-    MustDemobilize,
-    /// There is no measurement available, either because no
-    /// packet has been received yet, or because synchronization was rejected
-    NoMeasurement,
+    MustDemobilize(usize),
```
djc, line 37: I suggest `PeerIndex` would also make for a good newtype wrapper.

ntp-daemon/src/peer.rs
```diff
@@ -1,177 +1,272 @@
...
     /// Received an acceptable packet and made a new peer snapshot
-    Snapshot(PeerSnapshot),
+    Snapshot(usize, ResetEpoch, PeerSnapshot),
 }
 
-pub struct PeerChannels {
-    pub peer_snapshot: watch::Receiver<MsgForSystem>,
-    pub peer_reset: watch::Receiver<u64>,
+struct PeerProcess<C> {
```
djc, line 42: I'd probably call this `PeerTask` instead.

ntp-daemon/src/peer.rs
```diff
@@ -1,177 +1,272 @@
...
+        );
+
+        if let Some(msg) = prepare_msg_for_system(self.index, self.reset_epoch, accept, result) {
+            self.msg_for_system_sender.send(msg).await.ok();
+        }
+    }
 }
 
 #[instrument(skip(clock, config, system_snapshots, reset))]
 pub async fn start_peer<A: ToSocketAddrs + std::fmt::Debug, C: 'static + NtpClock + Send>(
```
djc, line 143: I'd suggest turning this into a `spawn()` static method on the `PeerProcess` type.

ntp-daemon/src/peer.rs
```diff
@@ -1,177 +1,272 @@
...
+            system_snapshots,
+            reset,
+            socket,
+            peer,
+            last_send_timestamp: None,
+            last_poll_sent: Instant::now(),
+            reset_epoch: ResetEpoch::default(),
+        };
 
         loop {
```
djc, line 177: I'd move the `loop` here into a `run()` method on the `PeerProcess` type.

If I seem to be contradicting myself here re inlining single-use functions, I think that's warranted here because the behavior for the first part here (initialize a tokio task) is quite different from the second part (continually keep running the peer's workload).

ntp-daemon/src/main.rs
folkertdev: there is some stuff I like in here, but one thing I'm strongly against: making invalid states representable. What you are proposing with splitting the `PeerStatus` from the `last: PeerSnapshot` is effectively creating an untagged union. That's ... bad.

ntp-daemon/src/main.rs
folkertdev: > with the usage of Msg being unidiomatic 

this is just not true. See `Vec`, `RwLock`, `Cow`. I generally like writing out names but especially in type names abbreviations are commonly used in practice.

Anyhow, the problem I'm hitting here is that most names don't communicate direction: does the system want the peer to update, or has the peer updated, which should cause changes to the system. "ForSystem" makes clear who needs to deal with this message.

ntp-daemon/src/main.rs
```diff
@@ -1,169 +1,165 @@
...
 
-    // channel to send the reset signal to all peers
-    let (reset_tx, reset_rx) = watch::channel::<u64>(0);
-    let mut last_reset_index: u64 = 0;
+    // send the reset signal to all peers
+    let mut reset_epoch: ResetEpoch = ResetEpoch::default();
+    let (reset_tx, reset_rx) = watch::channel::<ResetEpoch>(reset_epoch);
+
+    // receive peer snapshots from all peers
+    let (msg_for_system_tx, mut msg_for_system_rx) = tokio::sync::mpsc::channel::<MsgForSystem>(32);
```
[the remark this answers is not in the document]
djc: Sounds good to me. (Consider importing `tokio::sync::mpsc` and then calling `mpsc::channel` from there.)

ntp-daemon/src/main.rs
djc: I commonly use acronyms in type names, but not abbreviations. `Vec`, in my mind, is an exception because it's extremely common. I read `PeerUpdate` as a "peer update", which I think has a clear direction. That said, I'm not that interested in bikeshedding, so feel free to use my suggestions or ignore them.

ntp-daemon/src/main.rs
djc: I'm not sure what you mean, since in your current iteration you have `snapshots` and `peers` as separate `Vec`s, so my proposal doesn't change the design in that regard. (Unless I used the wrong type name for whatever thing `snapshots` contains.)

ntp-daemon/src/peer.rs
```diff
@@ -1,177 +1,281 @@
...
-pub struct PeerChannels {
-    pub peer_snapshot: watch::Receiver<MsgForSystem>,
-    pub peer_reset: watch::Receiver<u64>,
+    Snapshot(PeerIndex, ResetEpoch, PeerSnapshot),
 }
 
-#[instrument(skip(clock, config, system_snapshots, reset))]
-pub async fn start_peer<A: ToSocketAddrs + std::fmt::Debug, C: 'static + NtpClock + Send>(
-    addr: A,
+struct PeerTask<C> {
```
folkertdev, line 47: I'm bothered by this struct. It's just a random collection of fields, not really an abstraction. It is hard to figure out when certain pieces of state are updated. I also don't really see a good way to fix that though, so I'd love your suggestions

ntp-daemon/src/main.rs
folkertdev: I believe the idea here was to not pass anything into the clock that it doesn't need, to make the data flow easier (and e.g. to spot that certain fields are never read). @davidv1992 do you have an opinion here. It does look silly, but it's also harmless so I don't have strong feelings either way

ntp-daemon/src/peer.rs
```diff
@@ -1,177 +1,277 @@
...
-            tokio::select! {
-                () = &mut poll_wait => {
-                    let system_snapshot = *system_snapshots.borrow_and_update();
+    async fn handle_poll(&mut self, poll_wait: &mut Pin<&mut Sleep>) {
+        let system_snapshot = *self.channels.system_snapshots.read().await;
+        let packet = self.peer.generate_poll_message(system_snapshot);
 
+        // Sent a poll, so update waiting to match deadline of next
+        self.last_poll_sent = Instant::now();
+        self.update_poll_wait(poll_wait, system_snapshot);
```
davidv1992, line 98: This was probably a pre-existing bug, but we really should also check peer acceptability here and communicate it if the peer has become unnacceptable to use for synchronization. If you dont want to fix that in this PR then we should make an issue for this.

ntp-daemon/src/main.rs
davidv1992: I think this is a remnant from an earlier stage in development as this logic was developed partially in parallel with the clock_select struct, and have no problem just passing the clock select input directly, as it indeed looks somewhat silly

ntp-daemon/src/system.rs
```diff
@@ -0,0 +1,184 @@
...
+    // the channel closed and has no more messages in it
+    Ok(())
+}
+
+#[derive(Debug, Clone, Copy)]
+enum PeerStatus {
+    Demobilized,
+    AwaitingReset,
+    Valid(PeerSnapshot),
+}
```
davidv1992, line 136: Again, missing status for when the peer is just not acceptable for synchronization.

ntp-daemon/src/system.rs
```diff
@@ -0,0 +1,184 @@
...
+            }
+        }
+    }
+
+    fn reset_all(&mut self) {
+        for peer_status in self.peers.iter_mut() {
+            *peer_status = match peer_status {
+                PeerStatus::Demobilized => PeerStatus::Demobilized,
+                PeerStatus::AwaitingReset => PeerStatus::AwaitingReset,
+                PeerStatus::Valid(_) => PeerStatus::AwaitingReset,
```
davidv1992, line 180: Minor nit, although this helps more when the nodata state is also there: I would suggest using the _ matcher here as this really is just set to awaitingreset unless we are/have demobilized.

ntp-daemon/src/main.rs
davidv1992: Djc, I think folkert meant your suggestion for the struct peerstate, which seems to suggest separating the peersnapshot from the peerstatus. As a peersnapshot is really only useful for one specific status (the valid status) doing that would result in a datastructure that allows values to be created that are logically invalid, in effect the untagged union folkert is talking about.

ntp-daemon/src/system.rs
```diff
@@ -0,0 +1,184 @@
...
+
+    let mut peers = Peers::new(peer_addresses.len());
+
+    run(
+        config,
+        &mut peers,
+        reset_epoch,
+        global_system_snapshot,
+        msg_for_system_rx,
+        reset_tx,
```
davidv1992, line 49: Minor, however if we want to change it now is the time, but I think it might be nicer to not pass these as parameters to run, but instead have a daemon struct that contains most of these, with run a method on it.

davidv1992 (changes requested):
Structurally it reads much nicer already. However, there are some (in my view severe) logic problems that we really should fix before merging this.

ntp-daemon/src/main.rs
folkertdev: so, i tried this and the alternative is worse; we still need to unpack all those fields. We also have a bunch of tests that supply all these arguments. This does not really spark joy
```rust
let &FilterAndCombine {
    system_offset: offset,
    system_jitter: jitter,
    system_peer_snapshot: snapshot,
} = filter_and_combine;

let PeerSnapshot {
    time: last_peer_update,
    leap_indicator: leap_status,
    root_delay,
    root_dispersion,
    ..
} = snapshot;
```

ntp-daemon/src/peer.rs
```diff
@@ -1,177 +1,277 @@
...
     /// Received a Kiss-o'-Death and must demobilize
-    MustDemobilize,
-    /// There is no measurement available, either because no
-    /// packet has been received yet, or because synchronization was rejected
-    NoMeasurement,
+    MustDemobilize(PeerIndex),
     /// Received an acceptable packet and made a new peer snapshot
-    Snapshot(PeerSnapshot),
+    Snapshot(PeerIndex, ResetEpoch, PeerSnapshot),
 }
```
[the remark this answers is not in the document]
folkertdev: this state is now implicit. 

We start accepting messages from a peer when it sends a message in the correct reset epoch. That epoch is initially 0, and the first message that a peer will send will also have that reset epoch. Thus the message is accepted and the peer is considered valid from that point onward.

So we start out with marking all peers as "awaiting reset" (see `Peers::new()`).

This saves a state that would just have the same code everywhere as `AwaitingReset`

ntp-daemon/src/main.rs
davidv1992: Ok, in that case I suggest we just keep it as is.

ntp-daemon/src/system.rs
folkertdev: maybe I'm something here, but this is not where the system applies filtering. This is just whatever the peers report. The `snapshots` contain valid peer snapshots (so any peer that has a measurement taken in the current reset epoch), and only then do we actually inspect the content to see if we should use this peer's snapshot. 

The initial state of a peer is the `AwaitingReset`, which really means that the system is waiting for a valid message from this peer (valid meaning from withing the current reset epoch).

ntp-daemon/src/system.rs
davidv1992: Ok, then we will leave it as is.

ntp-daemon/src/peer.rs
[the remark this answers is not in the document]
folkertdev: what we are looking at here is the system keeping track of its peers at the highest level. Whether it actually sends snapshots with information that we think is acceptable is handled much later. Here we just record whether it's sending snapshots at all.

ntp-daemon/src/peer.rs
folkertdev: why is that needed here? can't we wait till the next packet we receive? What do we do when the peer is not acceptable any more (my guess is we just wait till the next packet)

ntp-daemon/src/main.rs
folkertdev: yes, exactly. Anyway I think the rest of the suggestions here have worked well and this is now outdated

ntp-daemon/src/peer.rs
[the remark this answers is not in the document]
folkertdev: you are correct: I dropped a `send` when we don't accept a packet. This has now been added.

ntp-daemon/src/system.rs
```diff
@@ -0,0 +1,199 @@
...
+    global_system_snapshot: Arc<tokio::sync::RwLock<SystemSnapshot>>,
+    mut msg_for_system_rx: mpsc::Receiver<MsgForSystem>,
+    reset_tx: watch::Sender<ResetEpoch>,
+) -> Result<(), Box<dyn Error>> {
+    let mut controller = ClockController::new(UnixNtpClock::new());
+    let mut snapshots = Vec::with_capacity(peers.len());
+
+    while let Some(msg_for_system) = msg_for_system_rx.recv().await {
+        peers.receive_update(msg_for_system, reset_epoch);
+
```
davidv1992, line 67: I am missing the early out here for messages that should not cause a recalculation

ntp-daemon/src/system.rs
```diff
@@ -0,0 +1,199 @@
...
+    /// Demobilized peers are kept because our logic is built around using indices,
+    /// and removing a peer would mess up the indexing.
+    Demobilized,
+    /// We are waiting for the first snapshot from this peer _in the current reset epoch_.
+    /// This state is the initial state for all peers (when the system is spawned), and also
+    /// entered when the system performs a clock jump and forces all peers to reset, or when a peer
+    /// indicates that it is no longer fit for synchronization (e.g. root distance became too big)
+    ///
+    /// A peer can leave this state by either becoming demobilized, or by sending a snapshot that
+    /// is within the system's current reset epoch.
```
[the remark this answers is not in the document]
folkertdev: which part?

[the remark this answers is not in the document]
davidv1992: The current text still suggests this only happens in response to resets, even though there are other reasons this can happen (which are probably just as common)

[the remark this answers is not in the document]
davidv1992: this state i mean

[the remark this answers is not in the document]
davidv1992: o never mind, i cannot read

PR #107, opened 2022-05-24
Implemented basic support for sentry.

davidv1992:


davidv1992:
Although sentry is compatible with opentelemetry, this compatibility is not feature-complete, and in particular is limited around crash reporting, which is the primary use we have for this integration. In short, we want this sentry integration to provide good error tracing when starting to test statime on internal machines, which we intend to start doing shortly.

djc:
Fair enough!

PR #113, opened 2022-05-27
Initial configuration setup

rnijveld:
This adds command line arguments and toml configuration parsing.

Some notes:

- A few options are available as command line arguments, mostly a way to specify the configuration file and an option to override the log filter. If a `NTP_LOG` environment variable is available, that is used as well for configuration the log filter.
- It is also possible to set the peers directly via the command line arguments, this will override the peers from the config.
- In the configuration each peer can be represented as a string or as a map. The map syntax can be used in the future for peer specific configuration (such as enabling initial burst mode, or defining peers with other discovery mechanisms).
- When we want the peerconfig to implement different discovery mechanisms we probably want to turn it into an enum instead.
- Currently we ignore unknown keys in the config, but I think we might add a flag to the serde deserialize config to disable that, as it may prevent typos etc
- Asside from peers we can currently also set the system configuration, but all values are currently just simple integers, I think for parsing of the `NtpDuration` for the distance threshold and `FrequencyTolerance` we may want to add a little more additional parsing in the future to make it a little easier to use for humans.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -57,6 +57,17 @@ dependencies = [
 ]
 
 [[package]]
+name = "atty"
+version = "0.2.14"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "d9b39be18770d11421cdb1b9947a45dd3f37e93092cbf377614828a319d5fee8"
+dependencies = [
+ "hermit-abi",
+ "libc",
+ "winapi",
+]
+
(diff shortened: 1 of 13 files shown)


davidv1992 (commented):
Looks strong in general, one short discussion point with regard to deserializers for peer_config.

folkertdev (approved):

PR #123, opened 2022-05-31
Move accept logic to system

folkertdev:
when a peer has a snapshot, it will now always send it to the system. The system is now resonsible for doing the acceptance check.

The upside of this change is that the system has a more complete picture of what the peers are doing. This is nice for observability.

### DIFF
--- a/ntp-daemon/src/peer.rs
+++ b/ntp-daemon/src/peer.rs
@@ -3,8 +3,8 @@ use std::{ops::ControlFlow, pin::Pin, sync::Arc};
 use tracing::warn;
 
 use ntp_proto::{
-    AcceptSynchronizationError, IgnoreReason, NtpClock, NtpHeader, NtpInstant, NtpTimestamp, Peer,
-    PeerSnapshot, PollInterval, ReferenceId, SystemConfig, SystemSnapshot,
+    IgnoreReason, NtpClock, NtpHeader, NtpInstant, NtpTimestamp, Peer, PeerSnapshot, ReferenceId,
+    SystemConfig, SystemSnapshot,
 };
 use ntp_udp::UdpSocket;
 use tracing::{info, instrument};
@@ -42,9 +42,6 @@ pub enum MsgForSystem {
     MustDemobilize(PeerIndex),
     /// Received an acceptable packet and made a new peer snapshot
     Snapshot(PeerIndex, ResetEpoch, PeerSnapshot),
-    /// Received a packet, but its content is not acceptable for synchronization
-    /// (e.g. invalid stratum, large root distance, forms a loop or peer is unreachable)
-    PeerNotFit(PeerIndex),
 }
 
 pub(crate) struct PeerChannels {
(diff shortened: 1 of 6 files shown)


ntp-daemon/src/peer.rs
```diff
@@ -92,36 +89,15 @@ where
...
         self.update_poll_wait(poll_wait, system_snapshot);
 
-        let accept = self.fitness(ntp_instant, system_snapshot.poll_interval);
-
-        if let Err(accept_error) = accept {
-            info!(?accept_error, "peer is not fit for use in synchronization");
-            let msg = MsgForSystem::PeerNotFit(self.index);
-            self.channels.msg_for_system_sender.send(msg).await.ok();
-        }
+        // NOTE: fitness check is not performed here, but by System
```
davidv1992, line 100: This now removes the entire step of sending here, reintroducing the bug of us considering unreachable peers.

ntp-daemon/src/system.rs
```diff
@@ -152,6 +160,7 @@ enum PeerStatus {
     NoMeasurement,
     /// This peer has sent snapshots taken in the current reset epoch. We store the most recent one
     Valid(PeerSnapshot),
+    NotFit(PeerSnapshot),
```
davidv1992, line 163: Fitness is a time-dependent state, I think we should step away from encoding fitness in the peerstatus enum here and just check it upon use.

ntp-proto/src/peer.rs
```diff
@@ -113,13 +113,17 @@ pub struct PeerSnapshot {
     pub(crate) stratum: u8,
     pub(crate) peer_id: ReferenceId,
 
+    pub(crate) reference_id: ReferenceId,
+    pub(crate) our_id: ReferenceId,
+    pub(crate) reach: Reach,
```
davidv1992, line 118: Why do these need to be pub(crate)?

ntp-proto/src/peer.rs
```diff
@@ -51,15 +51,15 @@ pub struct Peer {
 /// If the register contains any nonzero bits, the server is considered reachable;
 /// otherwise, it is unreachable.
 #[derive(Debug, Default, Clone, Copy)]
-struct Reach(u8);
+pub(crate) struct Reach(u8);
```
davidv1992, line 54: why does this need to be pub(crate)?

davidv1992 (changes requested):
This still has a few issues with reintroducing old issues (we really should building more tests for the daemon code itself).

ntp-daemon/src/system.rs
folkertdev: like above, this does mean that an invalid snapshot can trigger a clock select. That is always correct, but could be inefficient.

ntp-daemon/src/peer.rs
davidv1992: That would be best. I would consider just having two snapshot-sending messages, one where we have an actual new measurement, and one where we do not. And then even on the new measurement one system should probably only do clock selection and such if the peer is acceptable after this new measurement.

ntp-daemon/src/system.rs
davidv1992: see above

davidv1992 (approved):
Looks good, one inacurrate comment, but if that is fixed this is good to go.

PR #131, opened 2022-06-02
expose the peer state via a unix socket

folkertdev:


ntp-daemon/src/main.rs
```diff
@@ -23,5 +27,66 @@ async fn main() -> Result<(), Box<dyn Error>> {
...
+
+    std::fs::remove_dir_all("/run/ntpd-rs")?;
+    std::fs::create_dir("/run/ntpd-rs")?;
+
+    let peers_listener = UnixListener::bind("/run/ntpd-rs/observe")?;
+
+    // this binary needs to run as root to be able to adjust the system clock
+    // directories inherit root permissions, but the client should not need elevated
+    // permissions to read from the socket. So we explicitly set the permissions
+    let permissions: std::fs::Permissions = PermissionsExt::from_mode(0o777);
```
folkertdev, line 41: so maybe this makes perfect sense, but I needed `7` (so read/write/execute) as the permission here. intuitively 6 (read/write) should be enough but that did not work.

ntp-daemon/src/main.rs
```diff
@@ -23,5 +27,66 @@ async fn main() -> Result<(), Box<dyn Error>> {
...
     let _guard = finish_tracing_init(&mut config, has_log_override)?;
 
-    ntp_daemon::spawn(&config.system, &config.peers).await
+    let peers_reader = Arc::new(tokio::sync::RwLock::new(Vec::new()));
+    let peers_writer = peers_reader.clone();
+
+    std::fs::remove_dir_all("/run/ntpd-rs")?;
+    std::fs::create_dir("/run/ntpd-rs")?;
+
+    let peers_listener = UnixListener::bind("/run/ntpd-rs/observe")?;
```
folkertdev, line 36: we can make this path configurable, though that does mean that the client should use the same config.

ntp-daemon/src/main.rs
```diff
@@ -23,5 +27,66 @@ async fn main() -> Result<(), Box<dyn Error>> {
...
+
+    // this binary needs to run as root to be able to adjust the system clock
+    // directories inherit root permissions, but the client should not need elevated
+    // permissions to read from the socket. So we explicitly set the permissions
+    let permissions: std::fs::Permissions = PermissionsExt::from_mode(0o777);
+
+    std::fs::set_permissions("/run/ntpd-rs", permissions.clone())?;
+    std::fs::set_permissions("/run/ntpd-rs/observe", permissions)?;
```
folkertdev, line 44: we assume we're the only `ntpd-rs` instance that is running. That seems like a fair assumption to me

ntp-daemon/src/system.rs
```diff
@@ -162,7 +176,20 @@ enum PeerStatus {
     Measurement(PeerSnapshot),
 }
 
-struct Peers {
+#[derive(Debug, Serialize, Deserialize)]
+pub enum ObservablePeerState {
+    Nothing,
+    Observable {
```
folkertdev, line 182: we could retain the demobilized/NoMeasurement distinction here

ntp-proto/src/time_types.rs
```diff
@@ -321,6 +325,16 @@ impl NtpDuration {
...
+impl Serialize for NtpDuration {
+    fn serialize<S>(&self, serializer: S) -> Result<S::Ok, S::Error>
+    where
+        S: serde::Serializer,
+    {
+        let seconds = self.to_seconds();
+        seconds.serialize(serializer)
+    }
+}
+
```
folkertdev, line 337: mirrors the `Deserialize` implementation below

test-binaries/src/bin/peer-state.rs
```diff
@@ -0,0 +1,37 @@
+use ntp_daemon::ObservablePeerState;
+use std::error::Error;
+
+#[tokio::main]
+async fn main() -> Result<(), Box<dyn Error>> {
```
folkertdev, line 5: just a test application for now. We can expand this into the full client but I didn't want to do command-line argument parsing in this PR

ntp-daemon/src/main.rs
```diff
@@ -23,5 +27,66 @@ async fn main() -> Result<(), Box<dyn Error>> {
...
+
+                let buffer = {
+                    let state = peers_reader.read().await;
+
+                    serde_json::to_vec(state.deref()).unwrap()
+                };
+
+                write_to_unix_socket(stream, &buffer).await?;
+            }
+        }
```
davidv1992, line 69: I think this may become a little bit nicer if we spawn the observability handler as its own task. At that point we could wait for either to finish, or both with simpler constructs like join.

ntp-daemon/src/main.rs
```diff
@@ -23,5 +27,66 @@ async fn main() -> Result<(), Box<dyn Error>> {
     // this away when not using sentry.
     let _guard = finish_tracing_init(&mut config, has_log_override)?;
 
-    ntp_daemon::spawn(&config.system, &config.peers).await
+    let peers_reader = Arc::new(tokio::sync::RwLock::new(Vec::new()));
+    let peers_writer = peers_reader.clone();
+
+    std::fs::remove_dir_all("/run/ntpd-rs")?;
+    std::fs::create_dir("/run/ntpd-rs")?;
```
davidv1992, line 34: Removing and recreating the directory feels like work that shouldn't be the daemons responsibility. Also, if we do this I would strongly suggest we make the path configurable.

ntp-daemon/src/main.rs
davidv1992: this path should be configurable. Client can accept the path on the commandline or via configuration if it is non-default.

ntp-daemon/src/main.rs
davidv1992: not sure the permissions make sense to me either. However, I would suggest we also make this configurable so that a system administrator can decide who should be able to access the socket.

ntp-daemon/src/main.rs
davidv1992: agree that that assumption is reasonable, we need it elsewhere as well (the clock api doesn't take it well when two programs start to try to steer the same clock and may produce undesirable results for both in that case, so this doesn't seem horrible to me)

ntp-daemon/src/main.rs
```diff
@@ -23,5 +27,66 @@ async fn main() -> Result<(), Box<dyn Error>> {
...
+
+    let peers_listener = UnixListener::bind("/run/ntpd-rs/observe")?;
+
+    // this binary needs to run as root to be able to adjust the system clock
+    // directories inherit root permissions, but the client should not need elevated
+    // permissions to read from the socket. So we explicitly set the permissions
+    let permissions: std::fs::Permissions = PermissionsExt::from_mode(0o777);
+
+    std::fs::set_permissions("/run/ntpd-rs", permissions.clone())?;
```
davidv1992, line 43: We may want to make permissions on the directory separately configurable from the socket itself. Also am wondering a bit if creating, then setting permissions isn't a security risk, as it could potentially lead to a permission race on startup.

davidv1992 (changes requested):
This is missing some configurability which I think we should have before merging this. Also, in preparation for further work building on this, I would strongly suggest we already move the socket-handling to its own file, instead of keeping it in main (this will also help future testing).

I also personally am a bit hesitant on the way we generate the observability stuff. This is currently done in the main daemon loop continuously, from the snapshots we get for internal use. However, it seems to me more useful to just directly share the snapshots with the observability client, and instead do the conversion to the observability subset only there. Then we would only need to do that step when actually asked for it. What was your reason for going with the currently implemented approach?

ntp-daemon/src/main.rs
folkertdev: I played around with this a bit, but got nowhere. 

The `bind` call must create the "file". Trying to create the file with the normal rust functions results in a 

```
Error: Os { code: 98, kind: AddrInUse, message: "Address already in use" }
```

when you try `bind` (I made sure the file is closed, so I think it just bases this on the file existing)

ntp-daemon/src/main.rs
davidv1992: Ok, having looked into the options around security some more, I am going to take the stronger position that we should leave creating the /run directory to systemd, as this makes running with minimal permissions a lot easier.

ntp-daemon/src/main.rs
folkertdev: I think we want `either` behavior here. Certainly, when the system task is done, the application should shut down. Any of the other observing tasks should only exit when they hit an error, which I think should also just bring the app down (this most likely happens on startup, when some path is not right or some configuration setting is incorrect/missing)

ntp-daemon/src/main.rs
folkertdev: and we panic when the path doesn't exist?

ntp-daemon/src/main.rs
```diff
@@ -11,17 +16,88 @@ async fn main() -> Result<(), Box<dyn Error>> {
...
+            }
+            Err(ref e) if e.kind() == std::io::ErrorKind::WouldBlock => {
+                continue;
+            }
+            Err(e) => {
+                return Err(e);
+            }
+        }
+    }
 }
```
davidv1992, line 103: Why do we need this utility function, isn't write() in the AsyncWriteExt trait sufficient for our purposes?

davidv1992 (changes requested):
Still has some code-structure issues, and I want to reiterate that I really think we should move the observability part of main.rs into its own file, especially given future intentions around further functionality.

ntp-daemon/src/main.rs
folkertdev: I can't figure this out. But the above is what the documentation on `UnixStream` suggests. As far as I can see, the `write()` function is still susceptible to the readiness event being a false positive.

ntp-daemon/src/main.rs
davidv1992: That seems to me the only reasonable option.

ntp-daemon/src/main.rs
davidv1992: Perhaps we can avoid this by using umask just before creating the socket.

ntp-daemon/src/main.rs
folkertdev: this case is handled deep inside the stack, so using the `AsyncWriteExt` methods gives the same functionality.

ntp-daemon/src/config/mod.rs
```diff
@@ -57,6 +57,60 @@ pub struct Config {
...
+    pub mode: u32,
+}
+
+fn default_configure_path() -> PathBuf {
+    PathBuf::from("/run/ntpd-rs/configure")
+}
+
+const fn default_configure_permissions() -> u32 {
+    0o777
+}
```
folkertdev, line 88: this is convenient for testing, but maybe not the best default in practice? thoughts?

davidv1992: Small and need not be fixed before this gets merged, but I would change the configure default to limit that one to the user/group of the daemon.

davidv1992 (approved):
Looks good, can be merged.

PR #134, opened 2022-06-03
Disable panic in dev and release profiles to ensure proper termination on panic.

### LINKED ISSUE #132: Exceeding panic threshold does not cause proper termination
When trigering the panic threshold, the ntp-daemon goes into an unresponsive state but does not properly terminate. It should just terminate abnormally.

davidv1992:
Fixes #132


folkertdev (approved):
good catch!

PR #135, opened 2022-06-03
Changed systemd configuration to run with low privileges.

### LINKED ISSUE #118: Run the program in an as unpriviliged environment as possible
This would probably involve setting up capabilities (CAP_SYS_TIME specifically for our case) and setuid mechanics (to switch away to a user with no meaningful permissions on the system).

davidv1992:
fixes #118 without the need for any changes to our code, which seems to me the best solution to that problem.

### DIFF
--- a/CONFIGURATION.md
+++ b/CONFIGURATION.md
@@ -106,7 +106,7 @@ Furthermore, if at all possible, rebooting should be limited to only those exit 
 
 ## Systemd configuration
 
-To run NTPD-rs as the system NTP service, the following systemd service definition can be used. Note that this service definition assumes that the ntp-daemon binary has been installed to `/usr/local/bin`, and that the configuration is stored in the default `/etc/ntp.toml` location.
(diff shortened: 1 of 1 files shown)


CONFIGURATION.md
```diff
@@ -105,7 +105,7 @@ Furthermore, if at all possible, rebooting should be limited to only those exit
 
 ## Systemd configuration
 
-To run NTPD-rs as the system ntp service, the following systemd service definition can be used. Note that this service definition assumes that the ntp-daemon binary has bin installed to /usr/local/bin, and that the configuration is stored in the default /etc/ntp.toml location.
+To run NTPD-rs as the system ntp service, the following systemd service definition can be used. Note that this service definition assumes that the ntp-daemon binary has bin installed to /usr/local/bin, and that the configuration is stored in the default /etc/ntp.toml location. Furthermore, it assumes the existence of a low-privileged `ntpd-rs` group and user.
```
folkertdev, line 108: this should document how to create such a group/user

davidv1992: I don't think this is the right place to document that, as it is both a fairly standard system administrative task, and the procedure for doing so is not entirely uniform across the Linux ecosystem, depending on which utilities are included in the distribution. I have added a short sentence referring the user to their distributions documentation on this.

folkertdev (approved):

PR #138, opened 2022-06-07
ntp client

folkertdev:
a very rough draft for the client (but everything seems to work)


test-binaries/src/bin/client.rs
```diff
@@ -1,19 +1,79 @@
...
+pub const CMD_PEERS: &str = "peers";
+pub const CMD_SYSTEM: &str = "system";
+pub const CMD_CONFIG: &str = "config";
+
+pub const FLAG_LOG_FILTER: &str = "log-filter";
+
+fn build_app<'a>() -> Command<'a> {
+    Command::new("client")
+        // .version(concatcp!(VERSION, "\n"))
+        // .about("Runs the given .roc file, if there are no compilation errors.\nUse one of the SUBCOMMANDS below to do something else!")
```
davidv1992, line 13: This seems off, also this might be a license issue, please check you own all the code submitted here!

test-binaries/src/bin/client.rs
```diff
@@ -1,19 +1,79 @@
...
+use clap::{Arg, Command};
+use ntp_daemon::ObservableState;
+
+pub const CMD_PEERS: &str = "peers";
+pub const CMD_SYSTEM: &str = "system";
+pub const CMD_CONFIG: &str = "config";
+
+pub const FLAG_LOG_FILTER: &str = "log-filter";
+
+fn build_app<'a>() -> Command<'a> {
```
davidv1992, line 10: Nit: Why not just use the derive interface of clap like the daemon?

ntp-daemon/src/tracing.rs
```diff
@@ -60,6 +70,11 @@ pub fn init(
             }
         }
 
-        Ok(guard)
+        let state = TracingState {
+            guard,
+            reload_handle: fmt_handle,
+        };
+
+        Ok(state)
```
davidv1992, line 78: NIT: prefer Ok(TracingState {}) instead of a named intermediate, as it makes the code more compact and readable.

test-binaries/src/bin/client.rs
```diff
@@ -1,19 +1,79 @@
...
+            Command::new(CMD_SYSTEM).about("Information about the state of the daemon itself"),
+        )
+        .subcommand(
+            Command::new(CMD_CONFIG)
+                .about("Adjust configuration (e.g. loglevel) of the daemon")
+                .arg(
+                    Arg::new(FLAG_LOG_FILTER)
+                        .long(FLAG_LOG_FILTER)
+                        .help("Change the log filter")
+                        .default_value("error")
```
davidv1992, line 28: Suggest changing default to info, to match daemon's default

davidv1992 (changes requested):
Mostly minor comments, though the licensing issue will need fixing.

Other than those, this may also be the time to move the client out of the test-binaries into its own crate. What is your view on that?

davidv1992 (approved):
Looks good, nice set of changes

PR #139, opened 2022-06-08
Test for 1.51 MSRV

djc:


folkertdev:
is there a specific reason for picking 1.51? What are the considerations here?

folkertdev:
in particular because ultimately this is a binary, not a library, why does the MSRV matter?

djc:
Because people will want to compile this on Linux distributions using the compiler provided by the system package manager.

djc:
I don't expect compiling from source to ever be niche. My server runs Gentoo, and the expectation there is definitely that packages get compiled from source on a machine under the administrator's control. Same for Arch IIRC.

Personally I don't think the changes here have very significant downsides compared to the upside of allowing a substantially larger range of compiler versions.

davidv1992:
We have discussed this internally, and for now have decided to not take a strict policy on the MSRV. Our rationale for this comes down to the following:
 - We for now value the flexibility of being able to move into new rust features more than the advantages of supporting more versions.
 - The rust foundation itself only actively supports the latest stable version. As such, it is very much advisable for users to stick to that version anyway, as it is the only version that will receive any security updates.
 
What we will do for now is setup the CI such that we can track what our current minimum version number is, so that we at least notice it when we start using features requiring a newer rust version and update our documentation accordingly. Furthermore, we are considering to start providing signed precompiled binaries for a number of major distributions once we have a more stable client, which should alleviate the issue for most end users.

We will reconsider this policy if and when it becomes problematic for for example inclusion into distributions. As for locally compiling distros such as arch and gentoo, to the extent I have been able to verify they seem to do a good job of keeping up to date with the latest rust release so this should not pose a problem.

PR #153, opened 2022-06-16
Added configuration sanity checking.

### LINKED ISSUE #150: BUG: No warning when started with empty peer list
When started with empty peer list, no indication of an issue is given, instead the daemon stays very silent.

davidv1992:
Fixes #150

### DIFF
--- a/ntp-daemon/src/main.rs
+++ b/ntp-daemon/src/main.rs
@@ -18,6 +18,9 @@ async fn main() -> Result<(), Box<dyn Error>> {
 
     let mut config = Config::from_args(args.config, args.peers).await?;
 
+    // Warn/error if the config is unreasonable.
+    config.check();
+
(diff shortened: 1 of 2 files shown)


ntp-daemon/src/main.rs
```diff
@@ -18,6 +18,9 @@ async fn main() -> Result<(), Box<dyn Error>> {
 
     let mut config = Config::from_args(args.config, args.peers).await?;
 
+    // Warn/error if the config is unreasonable.
+    config.check();
```
rnijveld, line 22: Kind of thinking this check should be split/also/only happen after tracing is fully initialized, if your config has a valid sentry config, I'm assuming that the current no peers setup warning wouldn't show up in there right now, only in local logs (as long as you configured the command line log level high enough).

davidv1992: Fair point, fixed.

PR #169, opened 2022-06-23
test clap argument parsing

folkertdev:
- test clap argument parsing
- validate peer addresses and fail early  when invalid


davidv1992 (approved):
Looks good, the TryFrom is definitely an improvement over the old setup.

PR #174, opened 2022-06-28
Add ability to configure log output format

rnijveld:


ntp-daemon/src/main.rs
```diff
@@ -4,23 +4,32 @@ use clap::Parser;
...
-    let finish_tracing_init = ntp_daemon::tracing::init(log_filter);
+    let finish_tracing_init =
+        ntp_daemon::tracing::init(log_filter, args.log_format.unwrap_or_default());
 
-    let mut config = Config::from_args(args.config, args.peers).await?;
+    let mut config = match Config::from_args(args.config, args.peers).await {
+        Ok(c) => c,
+        Err(e) => {
+            error!("There was an error in the configuration: {}", e);
+            std::process::exit(78);
```
folkertdev, line 26: why 78? maybe document the magic number?

rnijveld: Hmm I shouldn't have pushed that code with this pull request, I'll split it out into a separate one. The number 78 is from `/usr/include/sysexits.h` btw, but that should definitely be documented.

folkertdev (approved):

davidv1992 (approved):

PR #175, opened 2022-06-28
Configuration changes client and sockets.

davidv1992:


ntp-daemon/src/main.rs
```diff
@@ -46,20 +46,14 @@ async fn main() -> Result<(), Box<dyn Error>> {
...
     )
     .await;
 
-    // exit if any of the tasks has completed
-    tokio::select! {
-        done = main_loop_handle => Ok(done??),
-        done = peer_state_handle => Ok(done??),
-        done = dynamic_config_handle => Ok(done??),
-    }
+    Ok(main_loop_handle.await??)
```
folkertdev, line 58: so now a failure in the observer/dynamic config will no longer crash the daemon, right? Is that actually what we want?

davidv1992: That is correct. I believe that is what we should want as crashes there shouldn't affect our ability to correctly steer time.

folkertdev: sure, fair enough. It is a bit of a tradeoff but that seems reasonable

PR #186, opened 2022-07-01
Added version and stratum checks for incoming packets.

### LINKED ISSUE #194: Validation of Parsed Packets
Reported by jsha

In ntp-proto/src/packet.rs, there are some input fields that could be validated against [limits in the specification](https://datatracker.ietf.org/doc/html/rfc5905#section-7.2) but do not appear to be:

- Version should be 4
- Stratum should be less than or equal to MAXSTRAT
- Poll should be validated against MINPOLL and MAXPOLL

davidv1992:
Fixes #194

### DIFF
--- a/fuzz/fuzz_targets/packet_parsing_sound.rs
+++ b/fuzz/fuzz_targets/packet_parsing_sound.rs
@@ -3,7 +3,11 @@ use libfuzzer_sys::fuzz_target;
 use ntp_proto::NtpHeader;
 
 fuzz_target!(|data: [u8;48]| {
-    let a = NtpHeader::deserialize(&data);
-    let b = a.serialize();
-    assert_eq!(data, b);
+    match NtpHeader::deserialize(&data) {
+        Ok(a) => {
+            let b = a.serialize();
+            assert_eq!(data, b);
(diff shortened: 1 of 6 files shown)


djc:
Should we instead make NtpHeader::deserialize() fallible?

davidv1992:
I don't think that that is a proper solution, packets with a stratum outside of the specified ranges still have a valid interpretation, they just shouldn't be used to synchronize with. The same holds for non-supported versions.

djc:
@jsha any thoughts on that?

rnijveld:
Personally I would prefer to fail early (i.e. during parsing) rather than late. It would add additional value to the type in that it always represents a somewhat usable packet, instead of one where we basically already know that we shouldn't use it (but where a programmer might forget to check).

djc:
> Personally I would prefer to fail early (i.e. during parsing) rather than late. It would add additional value to the type in that it always represents a somewhat usable packet, instead of one where we basically already know that we shouldn't use it (but where a programmer might forget to check).

Yes, exactly.

folkertdev (approved):

PR #189, opened 2022-07-01
Dynamic peer management

### LINKED ISSUE #111: Implement dynamic management of peers
In preparation for pools we should have tools to allow dynamic adding and removing of peers. (e.g. at runtime)

davidv1992:
Closes #111

### DIFF
--- a/ntp-daemon/src/peer_manager.rs
+++ b/ntp-daemon/src/peer_manager.rs
@@ -23,22 +23,34 @@ pub enum PeerStatus {
     Measurement(PeerSnapshot),
 }
 
-#[derive(Debug, Clone, Copy)]
+#[derive(Debug, Clone, Copy, Hash)]
 pub struct PeerIndex {
-    pub index: usize,
+    index: usize,
+}
+
+impl PeerIndex {
+    #[cfg(test)]
+    pub fn from_inner(index: usize) -> Self {
+        PeerIndex { index }
+    }
 }
 
 #[derive(Debug, Default)]
 pub struct Peers {
-    peers: Box<[PeerStatus]>,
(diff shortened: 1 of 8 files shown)


ntp-daemon/src/system.rs
```diff
@@ -364,7 +371,35 @@ mod tests {
         let (msg_for_system_tx, msg_for_system_rx) = mpsc::channel::<MsgForSystem>(32);
         let global_system_snapshot = Arc::new(tokio::sync::RwLock::new(SystemSnapshot::default()));
 
-        let peers_rwlock = Arc::new(tokio::sync::RwLock::new(Peers::new(&test_peer_configs(4))));
+        // TODO: Fix this
```
folkertdev, line 374: ?

davidv1992: This was a leftover from a refactoring step which I forgot to remove. It is gone now.

ntp-daemon/src/peer_manager.rs
```diff
@@ -23,30 +22,92 @@ pub enum PeerStatus {
...
+        let index = self.next;
+        self.next += 1;
+        PeerIndex { index }
+    }
+}
+
+#[derive(Debug)]
+struct PeerData {
+    status: PeerStatus,
+    config: Arc<PeerConfig>,
```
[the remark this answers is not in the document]
davidv1992: I made this an Arc in preparation for implementing things like pools. In situations like that, the configuration may be shared between multiple peers. It is not strictly necessary right now, but hopefully saves some conflicts in merges down the line.

folkertdev (approved):

PR #203, opened 2022-07-07
Limited window during which responses to poll are accepted.

### LINKED ISSUE #199: Make next_expected_origin valid for limited amount of time.
In ntp-proto/src/peer.rs, the next_expected_origin is kept live until either (a) a packet with that timestamp is accepted, or (b) the next poll interval arrives. Since the poll interval can be quite long at the maximum (many hours), this gives a long window for attackers to send packets guessing origin timestamp values. I recommend having a maximum time (on the order of seconds) past which reply packets won't be accepted, independent of the poll interval.

split off from #193

davidv1992:
Closes #199

### DIFF
--- a/ntp-proto/src/peer.rs
+++ b/ntp-proto/src/peer.rs
@@ -9,6 +9,7 @@ use serde::{Deserialize, Serialize};
 use tracing::{debug, info, instrument, trace, warn};
 
 const MAX_STRATUM: u8 = 16;
+const POLL_WINDOW: std::time::Duration = std::time::Duration::from_secs(5);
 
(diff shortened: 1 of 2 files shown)


ntp-proto/src/peer.rs
```diff
@@ -9,6 +9,7 @@ use serde::{Deserialize, Serialize};
 use tracing::{debug, info, instrument, trace, warn};
 
 const MAX_STRATUM: u8 = 16;
+const POLL_WINDOW: std::time::Duration = std::time::Duration::from_secs(5);
```
rnijveld, line 12: Maybe this is something that should be configurable, such that it can be set lower on high performance networks and higher on a mobile device on a slow connection?

PR #205, opened 2022-07-08
Implemented asymmetric panic thresholds.

### LINKED ISSUE #196: Forward vs Backward Movement
Reported by jsha

The most common need for a large step is when a system starts up and has no idea what time it is, or hasn't been keeping track of time while it was turned off. This should result in a forward step.

By contrast, the biggest risk to security is a large backward step in time. For instance, if an attacker is in possession of a misissued, revoked certificate, and an OCSP response or CRL from the past that says it's still good, rewinding a system clock could cause that system to accept the certificate even though it is revoked. By contrast, moving a clock forward in time will almost never cause a certificate to be accepted that would otherwise be rejected.

This means there's an asymmetry between forward and backward movement. I recommend ntp-daemon allow a different size of forward step than backward step, since forward steps are less risky.

davidv1992:
Closes #196

### DIFF
--- a/MANAGEMENT_CLIENT.md
+++ b/MANAGEMENT_CLIENT.md
@@ -19,7 +19,7 @@ The current client exposes 3 different commands:
 
 ## Available configuration parameters
 
-Currently, only the `log-level` and `panic-threshold` configuration parameters can be set dynamically, through the `--log-level` and `--panic-threshold` command line parameters respectively. For information on the allowed values for these, see [the configuration documentation](CONFIGURATION.md).
+Currently, only the `log-level` and `panic-threshold` configuration parameters can be set dynamically, through the `--log-level` and `--panic-threshold` command line parameters respectively. For information on the allowed values for these, see [the configuration documentation](CONFIGURATION.md). Note that for the panic threshold, only symmetric thresholds can be configured through the management client.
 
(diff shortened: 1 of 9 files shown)


MANAGEMENT_CLIENT.md
```diff
@@ -19,7 +19,7 @@ The current client exposes 3 different commands:
 
 ## Available configuration parameters
 
-Currently, only the `log-level` and `panic-threshold` configuration parameters can be set dynamically, through the `--log-level` and `--panic-threshold` command line parameters respectively. For information on the allowed values for these, see [the configuration documentation](CONFIGURATION.md).
+Currently, only the `log-level` and `panic-threshold` configuration parameters can be set dynamically, through the `--log-level` and `--panic-threshold` command line parameters respectively. For information on the allowed values for these, see [the configuration documentation](CONFIGURATION.md). Note that for the panic threshold, only symmetric thresholds can be configured through the managment client.
```
rnijveld, line 22: manag**e**ment

rnijveld (approved):

MANAGEMENT_CLIENT.md
davidv1992: Fixed

PR #215, opened 2022-07-15
Fixed way we add port to ipv6 addresses.

davidv1992:
This fix has the (intended) side effect of also removing the
need for network in parsing the configuration.


folkertdev:
yes, but now the host is not checked up-front. That means that an individual peer now has to deal with invalid hosts. I personally prefer checking that up-front, even if it requires network access.

folkertdev:
or does https://github.com/memorysafety/ntpd-rs/pull/216 handle invalid hosts?

folkertdev (approved):

PR #218, opened 2022-07-19
send timestamps

folkertdev:
finally worked this out. Two links that helped me crack this

- https://stackoverflow.com/questions/47313383/linux-udp-datagrams-and-kernel-timestamps-lots-of-examples-and-stackoversflow
- https://pastebin.com/qd0gspRc

the thing that nobody tells you is that these timestamping messages come in on the `MSG_ERRQUEUE`

### DIFF
--- a/ntp-proto/src/packet.rs
+++ b/ntp-proto/src/packet.rs
@@ -125,6 +125,8 @@ pub struct NtpHeader {
 }
 
 impl NtpHeader {
+    pub const WIRE_SIZE: usize = 48;
+
     /// A new, empty NtpHeader
     pub fn new() -> Self {
         Self {
--- a/ntp-proto/src/packet.rs
+++ b/ntp-proto/src/packet.rs
@@ -125,8 +125,6 @@ pub struct NtpHeader {
 }
 
 impl NtpHeader {
-    pub const WIRE_SIZE: usize = 48;
-
     /// A new, empty NtpHeader
     pub fn new() -> Self {
         Self {
--- a/ntp-proto/src/packet.rs
+++ b/ntp-proto/src/packet.rs
@@ -256,7 +256,7 @@ mod tests {
 
     #[test]
     fn roundtrip_bitrep_leap() {
-        for i in 0..4 as u8 {
+        for i in 0..4u8 {
             let a = NtpLeapIndicator::from_bits(i);
             let b = a.to_bits();
             let c = NtpLeapIndicator::from_bits(b);
@@ -267,7 +267,7 @@ mod tests {
 
     #[test]
     fn roundtrip_bitrep_mode() {
-        for i in 0..8 as u8 {
+        for i in 0..8u8 {
             let a = NtpAssociationMode::from_bits(i);
             let b = a.to_bits();
             let c = NtpAssociationMode::from_bits(b);
@@ -358,13 +358,11 @@ mod tests {
         for i in 0..=0xFF {
             let mut packet = base;
             packet[0] = i;
-            match NtpHeader::deserialize(&packet) {
-                Ok(a) => {
-                    let b = a.serialize();
-                    assert_eq!(packet, b);
-                }
-                _ => {}
-            };
+
+            if let Ok(a) = NtpHeader::deserialize(&packet) {
+                let b = a.serialize();
+                assert_eq!(packet, b);
+            }
         }
     }
 }
--- a/ntp-proto/src/packet.rs
+++ b/ntp-proto/src/packet.rs
@@ -1,6 +1,25 @@
+use std::fmt::Display;
+
 use serde::{Deserialize, Serialize};
 
 use crate::{NtpDuration, NtpTimestamp, ReferenceId};
+
+#[derive(Debug)]
+pub enum PacketParsingError {
+    InvalidVersion(u8),
+}
+
+impl Display for PacketParsingError {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        match self {
+            Self::InvalidVersion(version) => {
+                f.write_fmt(format_args!("Invalid version {}", version))
+            }
+        }
+    }
+}
+
+impl std::error::Error for PacketParsingError {}
 
 #[derive(Debug, Copy, Clone, PartialEq, Eq, Serialize, Deserialize)]
 pub enum NtpLeapIndicator {
@@ -84,10 +103,11 @@ impl NtpAssociationMode {
     }
 }
 
+pub const NTP_VERSION: u8 = 4;
+
 #[derive(Debug, Copy, Clone, PartialEq, Eq)]
 pub struct NtpHeader {
     pub leap: NtpLeapIndicator,
-    version: u8,
     pub mode: NtpAssociationMode,
     pub stratum: u8,
     pub poll: i8,
@@ -109,7 +129,6 @@ impl NtpHeader {
     pub fn new() -> Self {
         Self {
             leap: NtpLeapIndicator::NoWarning,
-            version: 4,
             mode: NtpAssociationMode::Client,
             stratum: 0,
             poll: 0,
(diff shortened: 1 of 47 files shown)


ntp-udp/src/lib.rs
```diff
@@ -98,27 +104,57 @@ impl AsRef<std::net::UdpSocket> for UdpSocket {
...
+    }
+}
+
 fn recv(socket: &std::net::UdpSocket, buf: &mut [u8]) -> io::Result<(usize, Option<NtpTimestamp>)> {
     let mut buf_slice = IoSliceMut::new(buf);
 
     // could be on the stack if const extern fn is stable
     let control_size =
-        unsafe { libc::CMSG_SPACE(std::mem::size_of::<libc::timespec>() as _) } as usize;
+        unsafe { libc::CMSG_SPACE((3 * std::mem::size_of::<libc::timespec>()) as _) } as usize;
```
folkertdev, line 157: SO_TIMESTAMPING sends more data than the SO_TIMESTAMPNS that we used before

ntp-udp/src/lib.rs
```diff
@@ -47,18 +49,22 @@ impl UdpSocket {
...
+                Ok(result) => match result {
+                    Err(e) => {
+                        debug!(error = debug(&e), "error sending data");
+                        return Err(e);
                     }
-                    return result;
-                }
+                    Ok(size) => {
+                        trace!(sent = size, "sent bytes");
+                        let send_timestamp = fetch_send_timestamp(self.io.get_ref())?;
```
rnijveld, line 64: I feel like it would be better to be a little more careful when retrieving the send timestamp. I'm unsure about the blocking behavior here (what if the kernel/driver responds too late and the recvmsg call is too early), I would prefer not to error on EWOULDBLOCK but instead have the same behavior as with our other recvmsg call. Secondly there could be other errors in the error queue, I think it would be best to keep to the timestamping docs (https://www.kernel.org/doc/Documentation/networking/timestamping.txt) and look for a message with the correct control message set. I'm unsure what to do about any other errors (or what could even appear here), I feel that ignoring other messages here is fine since we weren't looking at them before. But better to be safe here, or if we don't need that make sure we really document that it isn't needed.

folkertdev:
I have a rust implementation of that SO answer. But, what do we actually do with it? We already don't rely on that timestamp (an Option is returned). You would prefer to not even ask for it if we know beforehand that it is not supported?

ntp-udp/src/lib.rs
folkertdev: > I would prefer not to error on EWOULDBLOCK but instead have the same behavior as with our other recvmsg call

done

> Secondly there could be other errors in the error queue, I think it would be best to keep to the timestamping docs (https://www.kernel.org/doc/Documentation/networking/timestamping.txt, see 2.1.1) and look for a message with the correct control message set. 

so that would mean continue to receive until we find a message with the timestamping control message? and skip (but report) any other messages in that queue?

ntp-udp/src/lib.rs
```diff
@@ -98,27 +104,57 @@ impl AsRef<std::net::UdpSocket> for UdpSocket {
...
     Ok(())
 }
 
+fn cvt(t: libc::c_int) -> crate::io::Result<libc::c_int> {
+    match t {
+        -1 => Err(std::io::Error::last_os_error()),
+        _ => Ok(t),
+    }
+}
+
```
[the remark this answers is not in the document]
djc: The use of `crate::io` seems really strange, not least mixed with `std::io` in the function body. Could just use `io` directly for both, right?

I'd also recommend giving this an actually meaningful name instead of `cvt`.

ntp-proto/src/packet.rs
```diff
@@ -125,6 +125,8 @@ pub struct NtpHeader {
 }
 
 impl NtpHeader {
+    pub const WIRE_SIZE: usize = 48;
+
```
davidv1992, line 129: What is this needed for?

ntp-udp/src/lib.rs
```diff
@@ -47,18 +49,34 @@ impl UdpSocket {
...
+        let send_size = self.send_help(buf).await?;
+
+        // the send timestamp may never come (when the driver does not support send timestamping)
+        // set a very short timeout to prevent hanging forever. We automatically fall back to a
+        // less accurate timestamp when this function returns None
+        let timeout = std::time::Duration::from_millis(10);
+        match tokio::time::timeout(timeout, self.fetch_send_timestamp()).await {
+            Err(_) => Ok((send_size, None)),
+            Ok(send_timestamp) => Ok((send_size, send_timestamp?)),
+        }
```
davidv1992, line 62: Why not generate the fallback timestamp here? That would save us from returning an option and simplifies callers.

davidv1992 (commented):
This mostly looks good, but I am missing handling of 2 edge cases:
 - We don't obtain certainty that the received timestamp is from the last message sent (and not for example a wildly delayed message from an earlier poll).
 - What happens when a message from the server is received before the send timestamp?

folkertdev:
> What happens when a message from the server is received before the send timestamp?

that should be fine, because that message does not have the `MSG_ERRQUEUE` bit set. If somehow something incorrect does end up in the `MSG_ERRQUEUE` then we just don't find a timestamp and use the "in program" timestamp

folkertdev:
> We don't obtain certainty that the received timestamp is from the last message sent (and not for example a wildly delayed message from an earlier poll).

We very deliberately don't send the message body along with the send timestamp: you get an ethernet frame, which I don't really want to parse here.

I think it would be sufficient to check the send timestamp reported by the kernel against the timestamp that we create in the program. the kernel one must be later than our program one.

ntp-udp/src/lib.rs
```diff
@@ -67,6 +85,23 @@ impl UdpSocket {
         }
     }
 
+    async fn fetch_send_timestamp(&self) -> io::Result<Option<NtpTimestamp>> {
+        trace!("waiting for socket to become readable");
+        loop {
+            let mut guard = self.io.readable().await?;
+            match guard.try_io(|inner| fetch_send_timestamp_help(inner.get_ref())) {
+                Ok(send_timestamp) => {
+                    dbg!(&send_timestamp);
```
davidv1992, line 94: This should be replaced with proper logging

ntp-udp/src/lib.rs
```diff
@@ -47,18 +49,34 @@ impl UdpSocket {
...
         buf_size = buf.len(),
     ))]
-    pub async fn send(&self, buf: &[u8]) -> io::Result<usize> {
+    pub async fn send(&self, buf: &[u8]) -> io::Result<(usize, Option<NtpTimestamp>)> {
+        let send_size = self.send_help(buf).await?;
+
+        // the send timestamp may never come (when the driver does not support send timestamping)
+        // set a very short timeout to prevent hanging forever. We automatically fall back to a
+        // less accurate timestamp when this function returns None
+        let timeout = std::time::Duration::from_millis(10);
```
davidv1992, line 58: This timeout turns out to be a bit too low on my machine, maybe we should just make this 100 milliseconds.

ntp-udp/src/lib.rs
folkertdev: should it? it was there for development, to see that we actually ended up in that branch. Do you think it's valuable in production?

ntp-udp/src/lib.rs
folkertdev: sounds good

folkertdev: we'll need to look at this in more detail. bumping that timeout makes various tests fail

davidv1992: This seems to have been a symptom of waiting for the wrong queue (both of these actually), so don't think this needs further attention for now.

ntp-udp/src/lib.rs
```diff
@@ -221,14 +282,93 @@ fn recv(socket: &std::net::UdpSocket, buf: &mut [u8]) -> io::Result<(usize, Opti
...
+) -> io::Result<Option<NtpTimestamp>> {
+    // we get back two control messages: one with the timestamp (just like a receive timestamp),
+    // and one error message with no error reason. The payload for this second message is kind of
+    // undocumented.
+    //
+    // section 2.1.1 of https://www.kernel.org/doc/Documentation/networking/timestamping.txt says that
+    // a `sock_extended_err` is returned, but in practice we also see a socket address. The linux
+    // kernel also has this https://github.com/torvalds/linux/blob/master/tools/testing/selftests/net/so_txtime.c#L153=
+    const CONTROL_SIZE: usize = control_message_space::<[libc::timespec; 3]>()
+        + control_message_space::<(libc::sock_extended_err, libc::sockaddr_storage)>();
```
davidv1992, line 297: This seems to me to be somewhat weird, and doesn't entirely match with what the selftests do either. Are we sure this is correct? I would expect more something along the line of:
`control_message_space::<libc::timespec>` for the timestamp
`+control_message_space::<libc::sock_extended_err>` for the error part
`+control_message_space::<libc::sockaddr_storage>` for the sockaddr

folkertdev: I've adjusted this to 
```
    const CONTROL_SIZE: usize = control_message_space::<[libc::timespec; 3]>()
        + control_message_space::<(libc::sock_extended_err, libc::sockaddr)>();
```

the value of that is 112, which is exactly how many bytes are used (based on mhdr.msg_controllen after the receive_message. I checked this with a bigger buffer: 112 is what is actually used).

given these sizes in practice

```
[ntp-udp/src/lib.rs:301] control_message_space::<[libc::timespec; 3]>() = 64
[ntp-udp/src/lib.rs:301] control_message_space::<[libc::timespec; 1]>() = 32
[ntp-udp/src/lib.rs:301] control_message_space::<libc::sock_extended_err>() = 32
[ntp-udp/src/lib.rs:301] control_message_space::<libc::sockaddr>() = 32
```

I think my code is now correct and a tight bound.

ntp-udp/src/lib.rs
```diff
@@ -296,14 +388,162 @@ fn recv(
...
+) -> io::Result<Option<NtpTimestamp>> {
+    // we get back two control messages: one with the timestamp (just like a receive timestamp),
+    // and one error message with no error reason. The payload for this second message is kind of
+    // undocumented.
+    //
+    // section 2.1.1 of https://www.kernel.org/doc/Documentation/networking/timestamping.txt says that
+    // a `sock_extended_err` is returned, but in practice we also see a socket address. The linux
+    // kernel also has this https://github.com/torvalds/linux/blob/master/tools/testing/selftests/net/so_txtime.c#L153=
+    const CONTROL_SIZE: usize = control_message_space::<[libc::timespec; 3]>()
+        + control_message_space::<(libc::sock_extended_err, libc::sockaddr)>();
```
folkertdev, line 403: as a note: `sockaddr` gives a tight bound for current test cases. `sockaddr_storage` is much bigger, but I'm not sure that is ever really needed (sockaddr should cover ipv6 already)

davidv1992 (approved):
This is starting to look strong, I think we should merge this, but with a minor change that disables using transmit timestamping by default (but in such a way that we can keep tests and such)

PR #227, opened 2022-07-29
readme: the foundation is separate from the project

tshepang:
Note also that the project does all the decisions regarding Rust itself, including what toolchain versions are supported.


rnijveld:
Thanks for fixing the mistake!

PR #246, opened 2022-08-18
Added timing and system peer parameters needed for server to systemsnapshot.

### LINKED ISSUE #241: add extra clock functionality
in the header: what fields can we already fill in, what extra state do we need?

add config for how to deal with stratums: we should be able to say which stratum the system clock is; it is an upper limit on servers that we accept. 

stratum (according to the spec) is the stratum of the remote server that you pick plus one

davidv1992:
Closes #241

### DIFF
--- a/ntp-proto/src/peer.rs
+++ b/ntp-proto/src/peer.rs
@@ -94,11 +94,19 @@ impl Reach {
 
 #[derive(Debug, Clone, Copy, Serialize, Deserialize)]
 pub struct SystemSnapshot {
-    /// May be updated by local_clock
+    /// Desired poll interval
     pub poll_interval: PollInterval,
-    /// A constant at runtime
+    /// Precision of the local clock
     pub precision: NtpDuration,
-    /// May be updated by clock_update
+    /// Log of the precision of the local clock
+    pub stratum: u8,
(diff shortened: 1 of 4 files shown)


ntp-proto/src/peer.rs
```diff
@@ -94,11 +94,21 @@ impl Reach {
...
 pub struct SystemSnapshot {
-    /// May be updated by local_clock
+    /// Desired poll interval
     pub poll_interval: PollInterval,
-    /// A constant at runtime
+    /// Precision of the local clock
     pub precision: NtpDuration,
-    /// May be updated by clock_update
+    /// Log of the precision of the local clock
+    pub log_precision: i8,
```
folkertdev, line 102: maybe rename to `clock_log_position` or `clock_position_log`?

folkertdev: also, how should this value be interpreted? why are negative numbers used in the tests?

davidv1992: On second thought, I think it is a premature optimization to add this right now. For future reference, it is the 2log (rounded to integer) of the previous argument, though I made a small mistake in filling it in in the tests.

PR #248, opened 2022-08-18
clippy fixes in tests

folkertdev:


ntp-daemon/src/observer.rs
```diff
@@ -285,7 +285,6 @@ mod tests {
         let mut buf = [0_u8; 12];
         let mut bufref: &mut [u8] = &mut buf;
         reader.read_buf(&mut bufref).await.unwrap();
-        drop(bufref);
```
folkertdev, line 288: dropping a reference is pointless

ntp-daemon/src/peer.rs
```diff
@@ -723,7 +723,7 @@ mod tests {
 
         tokio::select! {
             _ = tokio::time::sleep(Duration::from_millis(10)) => {/*expected */},
-            _ = socket.recv(&mut buf) => { assert!(false); }
+            _ = socket.recv(&mut buf) => { panic!("unexpected"); }
```
djc, line 726: Maybe `unreachable!()`?

PR #250, opened 2022-08-18
use ifaddr to get actual timestamping support info

folkertdev:


ntp-udp/src/lib.rs
```diff
@@ -361,6 +384,89 @@ fn fetch_send_timestamp_help(
...
+
+            let name = interface.interface_name;
+            let length = Ord::min(name.len(), ifrn_name.len());
+            ifrn_name[0..length].copy_from_slice(&name.as_bytes()[0..length]);
+
+            Ok(Some(ifrn_name))
+        } else {
+            Ok(None)
+        }
+    }
```
davidv1992, line 467: This seems to return the first interface name given by the system, which we have no guarantee that it is actually the interface we are actually using with the socket.

folkertdev: yes, just pushed some code that actually finds the right name

ntp-udp/src/ifaddrs.rs
```diff
@@ -0,0 +1,115 @@
...
+                let ipv6 = std::net::Ipv6Addr::new(a, b, c, d, e, f, g, h);
+                let socketaddr = std::net::SocketAddrV6::new(
+                    ipv6,
+                    inaddr.sin6_port,
+                    inaddr.sin6_flowinfo,
+                    inaddr.sin6_scope_id,
+                );
+
+                Some(std::net::SocketAddr::V6(socketaddr))
+            }
```
folkertdev, line 60: is this realistically going to come up? In any case, I did not test this, just made the types/names work

ntp-udp/src/ifaddrs.rs
```diff
@@ -0,0 +1,115 @@
...
+            _ => None,
+        };
+
+        let addr = InterfaceAddress {
+            interface_name: ifname.to_string_lossy().to_string(),
```
folkertdev, line 65: has an allocation, but we run this once per socket so I think this is OK

ntp-udp/src/ifaddrs.rs
```diff
@@ -0,0 +1,113 @@
...
+    let mut addrs = mem::MaybeUninit::<*mut libc::ifaddrs>::uninit();
+    let errorcode = unsafe { libc::getifaddrs(addrs.as_mut_ptr()) };
+
+    match errorcode {
+        -1 => Err(std::io::Error::last_os_error()),
+        _ => Ok(InterfaceAddressIterator {
+            base: unsafe { addrs.assume_init() },
+            next: unsafe { addrs.assume_init() },
+        }),
+    }
```
davidv1992, line 112: nit: why not just use cerr here?

PR #258, opened 2022-08-25
get the interface name (using ifaddr under the hood)

folkertdev:
extracted from #250 to minimize merge conflicts


davidv1992 (commented):
One minor unsafe marking issue, otherwise good to merge.

PR #268, opened 2022-08-31
use epoll to await when the send timestamp is available

### LINKED ISSUE #261: Wait in proper way for timestamp events
Tokio only supports readable and writeable events for file descriptors, but we need a third. A workaround for this is to use a self-managed epoll fd, which can then be given to tokio and becomes readable when it has an event for us. We may still need to figure out which of the specific events we really need, but a simple test program should help us do that. See for documentation:
https://man7.org/linux/man-pages/man7/epoll.7.html
https://github.com/tokio-rs/tokio/issues/4885

folkertdev:
fix #261

This works reliably on my machine, but will need extra verification. In particular, we need to make sure this interacts well with tokio

### DIFF
--- a/ntp-udp/src/lib.rs
+++ b/ntp-udp/src/lib.rs
@@ -5,7 +5,7 @@ use std::{
     io::{ErrorKind, IoSliceMut},
     mem::{size_of, MaybeUninit},
     net::{Ipv4Addr, SocketAddr},
-    os::unix::prelude::AsRawFd,
+    os::unix::prelude::{AsRawFd, RawFd},
 };
 
 use ntp_proto::NtpTimestamp;
@@ -25,6 +25,7 @@ enum Timestamping {
 
 pub struct UdpSocket {
     io: AsyncFd<std::net::UdpSocket>,
+    exceptional_condition: AsyncFd<RawFd>,
     send_counter: u32,
     timestamping: TimestampingConfig,
 }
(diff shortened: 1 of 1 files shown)


ntp-udp/src/lib.rs
```diff
@@ -160,16 +163,15 @@ impl UdpSocket {
...
-            // becomes `writable()` when it has sent all its packets. So in practice this is also
-            // the moment when the timestamp message is available in the error queue.
-            let mut guard = self.io.writable().await?;
-            match guard.try_io(|inner| fetch_send_timestamp_help(inner.get_ref(), expected_counter))
-            {
+            // Therefore, we manually configure an extra file descriptor to listen for POLLPRI on
+            // the main udp socket. This `exceptional_condition` file descriptor becomes readable
+            // when there is something in the error queue.
+            let mut guard = self.exceptional_condition.readable().await?;
+            match guard.try_io(|_| fetch_send_timestamp_help(self.io.get_ref(), expected_counter)) {
```
folkertdev, line 174: here we use a guard on the `exceptional_condition` to perform io on the main `io` udp socket. I'm not 100% sure that this is correct if the `io` udp socket were to e.g. block.

davidv1992: From how I understand how tokio works internally, this is the correct approach here.

davidv1992 (approved):
Looks good, nice comments.

PR #269, opened 2022-09-01
refactor udp code

folkertdev:


davidv1992 (approved):
This is a really nice improvement. There are 3 points were safety still relies on the crate as a whole, I have marked these with comments, but this is a large enough improvement that I would prefer to merge this now.

Note: although the notes mark some internal functions as should be unsafe, having gone through the entire crate I am fairly certain the external interface of the crate is safe.

PR #271, opened 2022-09-01
add pool

folkertdev:


davidv1992 (changes requested):
This effectively removes support for individual peers, and allows only to configure pools. That seems like a bad idea to me. Furthermore, I would like to be able to configure how many peers I want from a pool on a per-pool basis. Finally, I think we want slightly different restart behaviour for a pool vs an individual peer, in that for an individual peer I would expect the client to always do a resolve of the domain name if it is forced to restart due to network issues, whilst for a pool, I would be comfortable with it using an address it fetched earlier but did not yet use.

PR #272, opened 2022-09-01
Implemented allow and deny lists.

davidv1992:
This does not yet implement changing these lists add runtime. I suggest we do that later given the size of these changes.

### DIFF
--- a/ntp-daemon/src/config/server.rs
+++ b/ntp-daemon/src/config/server.rs
@@ -9,9 +9,21 @@ use serde::{
     Deserialize, Deserializer,
 };
 
+use crate::{config::subnet::IpSubnet, ipfilter::IpFilter};
+
+#[derive(Debug, PartialEq, Eq, Copy, Clone, Deserialize)]
+pub enum FilterAction {
+    Ignore,
+    Deny,
+}
+
 #[derive(Debug, PartialEq, Eq, Clone)]
 pub struct ServerConfig {
     pub addr: SocketAddr,
+    pub denylist: IpFilter,
+    pub denylist_action: FilterAction,
+    pub allowlist: IpFilter,
+    pub allowlist_action: FilterAction,
 }
 
 impl TryFrom<&str> for ServerConfig {
@@ -20,6 +32,10 @@ impl TryFrom<&str> for ServerConfig {
     fn try_from(value: &str) -> Result<Self, Self::Error> {
         Ok(ServerConfig {
             addr: SocketAddr::from_str(value)?,
+            denylist: IpFilter::none(),
+            denylist_action: FilterAction::Ignore,
+            allowlist: IpFilter::all(),
+            allowlist_action: FilterAction::Ignore,
         })
     }
 }
@@ -46,18 +62,43 @@ impl<'de> Deserialize<'de> for ServerConfig {
 
             fn visit_map<M: MapAccess<'de>>(self, mut map: M) -> Result<ServerConfig, M::Error> {
                 let mut addr = None;
+                let mut allowlist = None;
+                let mut allowlist_action = None;
+                let mut denylist = None;
+                let mut denylist_action = None;
                 while let Some(key) = map.next_key::<&str>()? {
                     match key {
                         "addr" => {
                             if addr.is_some() {
                                 return Err(de::Error::duplicate_field("addr"));
                             }
-                            let raw: &str = map.next_value()?;
-
-                            let parsed_addr =
-                                SocketAddr::from_str(raw).map_err(de::Error::custom)?;
-
-                            addr = Some(parsed_addr);
+                            addr = Some(map.next_value::<SocketAddr>()?);
+                        }
+                        "allowlist" => {
+                            if allowlist.is_some() {
+                                return Err(de::Error::duplicate_field("allowlist"));
+                            }
+                            let list: Vec<IpSubnet> = map.next_value()?;
+                            allowlist = Some(IpFilter::new(&list));
+                        }
+                        "allowlist-action" => {
+                            if allowlist_action.is_some() {
+                                return Err(de::Error::duplicate_field("allowlist-action"));
+                            }
+                            allowlist_action = Some(map.next_value::<FilterAction>()?);
+                        }
+                        "denylist" => {
+                            if denylist.is_some() {
+                                return Err(de::Error::duplicate_field("denylist"));
+                            }
+                            let list: Vec<IpSubnet> = map.next_value()?;
+                            denylist = Some(IpFilter::new(&list));
+                        }
+                        "denylist-action" => {
+                            if denylist_action.is_some() {
+                                return Err(de::Error::duplicate_field("denylist-action"));
+                            }
+                            denylist_action = Some(map.next_value::<FilterAction>()?);
                         }
                         _ => {
                             return Err(de::Error::unknown_field(key, &["addr"]));
@@ -66,7 +107,29 @@ impl<'de> Deserialize<'de> for ServerConfig {
                 }
 
                 let addr = addr.ok_or_else(|| de::Error::missing_field("addr"))?;
-                Ok(ServerConfig { addr })
+                let (allowlist, allowlist_action) = match allowlist {
+                    Some(allowlist) => (
+                        allowlist,
+                        allowlist_action
+                            .ok_or_else(|| de::Error::missing_field("allowlist-action"))?,
+                    ),
+                    None => (IpFilter::all(), FilterAction::Ignore),
+                };
+                let (denylist, denylist_action) = match denylist {
+                    Some(denylist) => (
+                        denylist,
+                        denylist_action
+                            .ok_or_else(|| de::Error::missing_field("denylist-action"))?,
+                    ),
+                    None => (IpFilter::none(), FilterAction::Ignore),
+                };
+                Ok(ServerConfig {
+                    addr,
+                    allowlist,
+                    allowlist_action,
+                    denylist,
+                    denylist_action,
+                })
             }
         }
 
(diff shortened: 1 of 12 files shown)


ntp-daemon/src/ipfilter.rs
```diff
@@ -0,0 +1,328 @@
...
+        // Ensure values are sorted by value and then by length
+        data.sort_by(
+            |(v1, l1), (v2, l2)| {
+                if *v1 != *v2 {
+                    v1.cmp(v2)
+                } else {
+                    l1.cmp(l2)
+                }
+            },
+        );
```
folkertdev, line 82: isn't this the normal behavior of `PartialOrd` on tuples?

ntp-daemon/src/ipfilter.rs
```diff
@@ -0,0 +1,328 @@
...
+
+        let mut result = BitTree {
+            nodes: [Treenode::default()].into(),
```
folkertdev, line 85: eh, why not `vec![Treenode::default()]`?

ntp-daemon/src/ipfilter.rs
```diff
@@ -0,0 +1,328 @@
...
+/// It matches the integer a 4-bit segment at a time
+/// recording at each level whether for a given symbol
+/// all integers with the prefix extended with that
+/// symbol are either in or outside of the set.
+struct BitTree {
+    nodes: Vec<Treenode>,
+}
+
+fn top_quad(v: u128) -> u8 {
+    ((v & (0xF << 124)) >> 124) as u8
```
folkertdev, line 30: does `(v >> 124) & 0b1111` have problems? saves an instruction (or multiple because 128 bits do not fit in a single register)

also, why `quad` in the name? 4 bits are usually called a [nibble](https://en.wikipedia.org/wiki/Nibble)

ntp-daemon/src/ipfilter.rs
```diff
@@ -0,0 +1,328 @@
...
+        let mut last = 0;
+        for (idx, (val, _)) in data.iter().enumerate() {
+            let cur = top_quad(*val);
+            if cur != last {
+                for i in last + 1..=cur {
+                    first_idx[i as usize] = idx
+                }
+                last = cur;
+            }
+        }
```
folkertdev, line 107: maybe note that this loop relies heavily on the input data being sorted?

folkertdev: also my intuition says that the `leading_zeros` function could come in handy here

ntp-daemon/src/ipfilter.rs
```diff
@@ -0,0 +1,328 @@
...
+        result
+    }
+
+    /// Create the substructure for a node, recursively.
+    /// Max recursion depth is maximum value of data[i].1/4
+    /// for any i
+    fn fill_node(&mut self, mut data: &mut [(u128, u8)], node_idx: usize) {
+        // Figure out how the data splits over the 16 segments
+        // in this node.
+        let mut first_idx = [0; 16];
```
folkertdev, line 97: initialize this with `[data.len(); 16]`, saves a loop below

ntp-daemon/src/ipfilter.rs
davidv1992: fixed

ntp-daemon/src/ipfilter.rs
davidv1992: fixed

ntp-daemon/src/ipfilter.rs
davidv1992: fixed

ntp-daemon/src/ipfilter.rs
```diff
@@ -0,0 +1,336 @@
+use std::net::{IpAddr, Ipv4Addr, Ipv6Addr};
+
+use crate::config::subnet::IpSubnet;
+
+/// One part of a BitTree
+#[derive(Debug, Copy, Clone, Default, PartialEq, Eq)]
+struct Treenode {
```
djc, line 7: Should probably be called `TreeNode`.

ntp-daemon/src/ipfilter.rs
```diff
@@ -0,0 +1,336 @@
...
+            }
+            self.fill_node(seg, child_offset);
+            child_offset += 1;
+        }
+    }
+}
+
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct IpFilter {
+    ipv4filter: BitTree,
```
djc, line 172: `ipv4_filter`

ntp-daemon/src/config/server.rs
```diff
@@ -20,6 +32,10 @@ impl TryFrom<&str> for ServerConfig {
     fn try_from(value: &str) -> Result<Self, Self::Error> {
         Ok(ServerConfig {
             addr: SocketAddr::from_str(value)?,
+            denylist: IpFilter::none(),
```
djc, line 35: Quite a bit of duplication here, maybe add some abstraction?

ntp-daemon/src/ipfilter.rs
```diff
@@ -0,0 +1,336 @@
...
+    // Where in the array the child nodes of this
+    // node are located. A child node is only
+    // generated if the symbol cannot be used to
+    // make a final decision at this level
+    child_offset: u32,
+    inset: u16,
+    outset: u16,
+}
+
+#[derive(Debug, Clone, PartialEq, Eq)]
```
djc, line 17: I do wonder whether this is all worth it compared to using BTreeMap with (IpAddr, u8).

davidv1992: I initially rejected BTree because I didn't see how to get it working (this requires O(n) time with a BTree if you just put in all the ranges, because there can be overlap), but thinking about it now you can do prefiltering on the ranges to get a BTree working as well. It will probably work, but my estimate would be that it would be less efficient, though I am not sure how. Given the need for special construction of the BTree I am not sure this is worth investigating.

ntp-daemon/src/config/server.rs
davidv1992: Maybe I am just too used to the code as it is, but I am not really getting what duplication you are getting at, could you clarify?

ntp-daemon/src/ipfilter.rs
davidv1992: fixed.

ntp-daemon/src/ipfilter.rs
davidv1992: fixed

ntp-daemon/src/config/server.rs
djc: You have multiple pairs of an `IpFilter` and a `FilterAction`. It might be better to group them in a single type.

ntp-daemon/src/ipfilter.rs
djc: Presumably if you just need a subnets (that is, the bit filter is only a prefix) than it seems pretty straightforward, I think I've used it before (not sure what you mean by prefiltering).

davidv1992:
DO NOT MERGE: something is wrong with CI and I am using this to actively debug that currently.

ntp-daemon/src/ipfilter.rs
djc: I guess that's true. Adding a check on insertion still seems quite a bit simpler than including your own `BitTree` implementation, though.

davidv1992: I agree, and had I thought of that before implementing the current solution, that would have been the route to take. However, I think there is no strong reason at this time to go back and implement another approach given the current state of the code.

ntp-daemon/src/config/server.rs
davidv1992: I have put some thought into this, and I disagree since although there is superficial duplication, the interpretation of both fields is quite significantly different between the allow and deny lists, and therefore combining these does in my opinion not really give much space for code reuse, but does give space for confusion on behavior.

ntp-daemon/src/ipfilter.rs
djc: I could hardly disagree more. @rnijveld any thoughts?

rnijveld: I think there's definitely some reasons to go for this rather than something else, but the main one for me would be that we can fully control the efficiency with this implementation, which is something that is definitely important for a deny/allow list given that we want to do that as quickly as possible while using as few resources as possible. Building upon a libstd datastructure cannot give us that level of control.

I think right now we have three options to resolve this

* Complete the Vec-based implementation that @folkertdev started
* Create a BTree based implementation
* Use the implementation from this pull request

I feel that the Vec-based implementation is a long way from being completed, and Folkert already took quite some time to get it to where it is right now. I'm also not too convinced, with all the exceptions that need to be there, it will be much more readable than the implementation in this PR.

Using a BTree would help a little with worst case scenario's compared to the Vec I guess, but I don't see right now how to use the BTreeMap/BTreeSet APIs as efficiently as possible. @djc You wouldn't happen to have any example we could take a look at here? That might help!

The current implementation we have right here might use a custom datastructure, but I feel that the extra control we have here does add some value. Other than that this implementation is already documented, tested, and pretty much is confirmed to be working correctly. The custom datastructure code does take a little while to read, but I don't think it's too excessive to understand, and the API is just as easy as any other implementation would be.

Given all that I think that we should just go ahead and merge this one for now, and maybe at a later point (or if someone really wants to right now) replace this implementation with another one if need be.

@davidv1992 could you do a rebase/merge of main? I don't think there's anything else left to do before a merge, code looks great!

davidv1992: @rnijveld Did the rebase

PR #285, opened 2022-09-15
use an ip filter based on a sorted vector

folkertdev:
so that we can properly contrast the approaches: an implementation of an ip filter using a sorted vector.

The `BTreeMap` provides sorted keys, but we don't really get any other benefit, and assume that storing the values in a `Vec` will be better versus a tree (e.g. scanning can be vectorized and data locality is better)

This implementation is much smaller than the bitset implementation, and has very little complex logic (some bit shifts, that's it really). A clear win.

But, the time complexity of a lookup (the important operation here, construction is performed once and hence I don't think its complexity is relevant) degrades from `O(1)` to `O(n)`, where `n` is the number of IPs in the filter. That is unfortunate, but

- we expect `n` to be small (couple dozen, at most)
- the implementation can be readily vectorized
- technically we can get it down to `O(log n)` by using a binary search instead of a linear scan. But I predict this is actually slower for small n

So overall I'm in favor of this simpler, `Vec`-based approach

### DIFF
--- a/ntp-daemon/src/ipfilter.rs
+++ b/ntp-daemon/src/ipfilter.rs
@@ -2,236 +2,128 @@ use std::net::{IpAddr, Ipv4Addr, Ipv6Addr};
 
 use crate::config::subnet::IpSubnet;
 
-/// One part of a BitTree
-#[derive(Debug, Copy, Clone, Default, PartialEq, Eq)]
-struct TreeNode {
-    // Where in the array the child nodes of this
-    // node are located. A child node is only
-    // generated if the symbol cannot be used to
-    // make a final decision at this level
-    child_offset: u32,
-    inset: u16,
-    outset: u16,
+pub trait IpAddrContains: Ord + Copy + std::fmt::Debug {
+    fn contains(haystack_ip: Self, haystack_mask: u8, needle_ip: Self) -> bool;
+}
+
+impl IpAddrContains for Ipv4Addr {
+    fn contains(haystack_ip: Self, haystack_mask: u8, needle_ip: Self) -> bool {
+        let shifted = |ip: Self| {
+            u32::from_be_bytes(ip.octets())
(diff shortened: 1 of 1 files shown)


folkertdev:
@djc is this what you had in mind? Anything we can simplify further?

folkertdev:
the fuzz tests expose some real problems with this approach. Consider

```rust
        let nets = [
            IpSubnet {
                addr: "0.7.45.10".parse().unwrap(),
                mask: 17,
            },
            IpSubnet {
                addr: "255.255.98.98".parse().unwrap(),
                mask: 3,
            },
        ];

        let needle = ["255.33.0.0".parse().unwrap()];
```

There is no overlap between the patterns: the first bit is different. But the first ip that is smaller than our needle is the `0.7.45.10` and that does not match so we conclude the needle is not in the filter. That is wrong, because the first 3 bits of `255` overlap with the needle. 

So picking the first IP that is smaller does not work, and hence I now don't see how we could do better than a linear scan. Given that `n` is small that may actually be a fine solution, but I fail to see how a `BTree` or sorting the keys is advantageous: the sorting does not consider the mask and is therefore misleading.

djc:
If the mask is 3 you should store `224.0.0.0` as the addr, then this should work just fine?

folkertdev:
That does not just work. Assume we have an address

    124.999.0.0

and masks

    124.888.0.0
    124.0.0.0

Then when we just sort and pick the "first" that is lower, we pick `124.888.0.0` which does not match.

Is there some algorithmic trick I'm missing here?

I've been told that converting ipv4 to ipv6 is not a good idea, maybe @rnijveld  can elaborate?

djc:
124.999.0.0 and 124.888.0.0 are clearly nonsensical, do you have a realistic counterexample? It seems to me that if you actually apply the subnet mask to the IP than the resulting IP will always be lower than or equal to any IP that is expected to match it.

PR #286, opened 2022-09-15
config for pools

folkertdev:
breaking pools into a couple of smaller pieces. Hopefully this is uncontroversial.

### DIFF
--- a/ntp-daemon/src/config/peer.rs
+++ b/ntp-daemon/src/config/peer.rs
@@ -7,7 +7,9 @@ use serde::{
 
 #[derive(Deserialize, Debug, PartialEq, Eq, Clone, Copy)]
 pub enum PeerHostMode {
+    #[serde(alias = "server")]
     Server,
+    #[serde(alias = "pool")]
     Pool,
 }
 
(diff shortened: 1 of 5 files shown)


ntp-daemon/src/config/peer.rs
```diff
@@ -16,14 +16,52 @@ impl Default for PeerHostMode {
     }
 }
 
-#[derive(Debug, PartialEq, Eq, Clone)]
-pub struct PeerConfig {
+#[derive(Deserialize, Debug)]
+pub enum PeerType {
```
rnijveld, line 20: I believe `PeerHostMode` was originally added as the discriminator for the same purpose, so I think we should remove that one. I would also add lowercase aliases for the variants in `PeerType` and/or `PeerHostMode`.

ntp-daemon/src/config/peer.rs
```diff
@@ -16,14 +16,52 @@ impl Default for PeerHostMode {
...
+
+#[derive(Debug, PartialEq, Eq, Clone)]
+pub enum PeerConfig {
+    Standard(StandardPeerConfig),
+    Pool(PoolPeerConfig),
+    // Consul(ConsulPeerConfig),
+}
+
+impl PeerConfig {
+    pub fn addr(&self) -> &str {
```
rnijveld, line 50: Are you sure we want to add these kinds of methods on PeerConfig? It looks to me like they quickly become irrelevant for other types of configurations. Instead maybe match on the PeerConfig variant (and add a little `unimplemented!` for variants we don't support right now)?

PR #287, opened 2022-09-15
ntp-udp safety

davidv1992:
Various improvements to the analysis and isolation of unsafe code within the ntp-udp crate

Note: This PR does 3 things:
 - It isolates each section of unsafe code in its own module such that that module has a safe interface.
 - It reworks some code to ensure the above is actually possible
 - It adds the argumentation why each of the unsafe code blocks has safe behaviour.

The commits below are split consciously to make the review process hopefully easier. Each commit either moves stuff around, or reworks a specific bit for proving safety.


davidv1992:
@djc I think it would be very useful if you also have some time to look at this.

ntp-udp/src/lib.rs
```diff
@@ -1,225 +1,5 @@
 mod interface_name;
+mod raw_socket;
 mod socket;
 
 pub use socket::UdpSocket;
-use std::os::unix::prelude::{AsRawFd, RawFd};
```
djc, line 5: IMO splitting this in multiple tiny modules just makes it harder to understand how these pieces fit together. Also not clear if this commit is strictly moving code around or if it also changes the code in other ways?

ntp-udp/src/raw_socket/exceptional_condition_fd.rs
```diff
@@ -7,6 +7,7 @@ use super::cerr;
 pub(crate) fn exceptional_condition_fd(
     socket_of_interest: &std::net::UdpSocket,
 ) -> std::io::Result<AsyncFd<RawFd>> {
+    // Safety:
```
djc, line 10: Missed one?

ntp-udp/src/raw_socket/set_timestamping_options.rs
```diff
@@ -35,13 +35,18 @@ pub(crate) fn set_timestamping_options(
             | SOF_TIMESTAMPING_OPT_ID;
     }
 
+    // Safety:
+    // we have a reference to the socket, so fd is a valid file descriptor for the duration of the call
+    // SOL_SOCKET + SO_TIMESTAMPING expect a *u32 as value, which &options is. Furthermore, we own options
```
djc, line 40: I did not find documentation for `SO_TIMESTAMPING` easily on DuckDuckGo. Maybe provide a link here with what's required? It's surprising to me that this would be defined as `u32` rather than something like `c_uint`.

ntp-udp/src/raw_socket/recv_message.rs
```diff
@@ -1,21 +1,111 @@
...
+
+use tracing::warn;
+
+use crate::interface_name::sockaddr_storage_to_socket_addr;
 
 use super::cerr;
 
-/// Receive a message on a socket (retry if interrupted)
-pub(crate) fn receive_message(
+pub(crate) enum MessageQueue {
```
djc, line 9: This commit has too much going on for me to review efficiently. I'd recommend cutting this up in smaller pieces which should make it easier to verify each change in isolation.

ntp-udp/src/interface_name.rs
```diff
@@ -48,10 +48,10 @@ impl InterfaceAddress {
     ///
     /// assumes a valid `libc::ifaddrs`
     unsafe fn from_libc_ifaddrs(info: &libc::ifaddrs) -> InterfaceAddress {
```
djc, line 50: At this point, should this function no longer be declared `unsafe`? It seems that the required invariant that `&libc::ifaddrs` contains a vali `libc::ifaddrs` is quite natural and something that should be enforced at creation time.

davidv1992: Unfortunately, due to libc being a direct binding to the c library, it is possible to create invalid libc::ifaddrs values in safe rust code, hence the unsafe here.

ntp-udp/src/lib.rs
davidv1992: Added more text to the PR description explaining what this PR does.

As to the tiny module sizes, this was an explicit choice. Due to how visibility works in rust, all code in a module and its submodules must be considered when deciding whether an unsafe statement occuring in that module is safe. This makes it hard to reason about safety in larger modules.

Because of the potential impact of unsafety, I decided here to prioritize the ease of reasoning about safety over the clarity of the larger scale structure. I will add additional comments to the modules to explain the structure and the reasoning behind it.

djc: That's fair, but in that case I would advocate using inline modules (`mod foo { }`) instead of modules defined through a separate file.

ntp-udp/src/interface_name.rs
djc: Out of curiosity, can you give me an example of safe Rust code that will generate an invalid `libc::ifaddrs` value?

davidv1992:
```
let mut buf: [i8;2] = [1,2];
let a = libc::ifaddrs {
    ifa_next: std::ptr::null_mut(),
    ifa_name: buf.as_mut_ptr(),
    ifa_flags: 0,
    ifa_addr: std::ptr::null_mut(),
    ifa_netmask: std::ptr::null_mut(),
    ifa_ifu: std::ptr::null_mut(),
    ifa_data: std::ptr::null_mut(),
};
```
The resulting a is not a valid ifaddr (ifa_name is not a valid C string), and will in fact trigger UB when passed to `from_libc_ifaddrs`.

djc: Fair enough!

ntp-udp/src/lib.rs
folkertdev: seconding this. I personally dislike having many files (makes navigating harder to me), and the only goal here is namespacing/isolation so inline modules seem like a good fit.

ntp-udp/src/raw_socket/exceptional_condition_fd.rs
davidv1992: Yes, added.

ntp-udp/src/raw_socket/set_timestamping_options.rs
davidv1992: added documentation link.

ntp-udp/src/raw_socket/recv_message.rs
davidv1992: Split this commit into multiple

ntp-udp/src/lib.rs
davidv1992: I have done the flattening into a single file as a final commit on the end. If it is highly desirable for review to have this flattening applied throughout the history please let me know.

ntp-udp/src/raw_socket.rs
```diff
@@ -0,0 +1,386 @@
...
+mod exceptional_condition_fd {
+    use std::os::unix::prelude::{AsRawFd, RawFd};
+
+    use tokio::io::unix::AsyncFd;
+
+    use super::cerr;
+
+    pub(crate) fn exceptional_condition_fd(
+        socket_of_interest: &std::net::UdpSocket,
+    ) -> std::io::Result<AsyncFd<RawFd>> {
```
folkertdev, line 25: maybe add some docs here on what an this is (with some links to linux docs). It also would be really nice if we could upstream this, maybe also link the tokio issue about it that we have somewhere but that I can't find right now

ntp-udp/src/raw_socket.rs
```diff
@@ -0,0 +1,386 @@
...
+        rx_filters: u32,
+        rx_reserved: [u32; 3],
+    }
+
+    #[repr(C)]
+    #[allow(non_camel_case_types)]
+    struct ifreq {
+        ifrn_name: [u8; 16],
+        ifru_data: *mut libc::c_void,
+        __empty_space: [u8; 40 - 8],
```
folkertdev, line 324: just as a note: this ties us to 64-bit operating systems, and also probably to linux. The union that libc uses here is extremely nasty. We should maybe look into upstreaming this?

ntp-udp/src/raw_socket.rs
```diff
@@ -0,0 +1,386 @@
...
+pub(crate) use set_timestamping_options::set_timestamping_options;
+pub(crate) use timestamping_config::TimestampingConfig;
+
+/// Turn a C failure (-1 is returned) into a rust Result
+pub(crate) fn cerr(t: libc::c_int) -> std::io::Result<libc::c_int> {
+    match t {
+        -1 => Err(std::io::Error::last_os_error()),
+        _ => Ok(t),
+    }
+}
```
folkertdev, line 386: any particular reason that this is not at the top of the file? seem like the more natural place, since `cerr` is used in the inner modules.

folkertdev (approved):

ntp-udp/src/raw_socket.rs
davidv1992: Not sure what you mean here. I did not change anything here so maybe we should fix this separately?

folkertdev: I know. It's just that with our eye on supporting more platforms, this is one place where I know we make an assumption about running on a 64-bit OS

ntp-udp/src/raw_socket.rs
davidv1992: No, so moved it.

ntp-udp/src/raw_socket.rs
davidv1992: added addition comment

ntp-udp/src/lib.rs
davidv1992: Reordered to keep the original ordering in that file. There still is one ordering change, namely that the cerr function moved to the top, but hopefully that is not too distracting. Please let me know if this flattening is highly desirable to be done throughout the commit history for this PR.

ntp-udp/src/lib.rs
```diff
@@ -1,219 +1,424 @@
...
+                    Err(e) if std::io::ErrorKind::Interrupted == e.kind() => {
+                        // retry when the recv was interrupted
+                        continue;
+                    }
+
+                    other => break other,
+                }
+            }?;
+
+            if mhdr.msg_flags & libc::MSG_TRUNC > 0 {
```
djc, line 155: Is it valid to compare `msg_flags` if `recvmsg()` returned anything other than `std::io::ErrorKind::Interrupted`? It feels like we should only run these blocks if `sent_bytes` is `Ok`?

ntp-udp/src/lib.rs
```diff
@@ -1,219 +1,424 @@
...
-    rx_filters: u32,
-    rx_reserved: [u32; 3],
-}
+        impl<'a> Iterator for ControlMessageIterator<'a> {
+            type Item = ControlMessage;
+
+            fn next(&mut self) -> Option<Self::Item> {
+                // Safety:
+                // CMSG_FIRSTHDR and CMSG_NXTHDR only return valid pointers or NULL when given valid input
+                let current_msg = unsafe { self.current_msg.as_ref() };
```
djc, line 220: Could return `None` here by using the `?` operator to early return `None`, reducing indentation for the rest of the function.

ntp-udp/src/lib.rs
```diff
@@ -1,219 +1,424 @@
...
+                if let Some(current_msg) = current_msg {
+                    // Safety:
+                    // New ensure mhdr is valid
+                    // CMSG_FIRSTHDR and CMSG_NXTHDR only return valid pointers or NULL when given valid input
+                    self.current_msg = unsafe { libc::CMSG_NXTHDR(&self.mhdr, self.current_msg) };
+
+                    match (current_msg.cmsg_level, current_msg.cmsg_type) {
```
djc, line 227: Style nit: can wrap the `Some` around this `match` instead of repeating it in each arm.

ntp-udp/src/lib.rs
```diff
@@ -1,219 +1,424 @@
...
+                        (libc::SOL_SOCKET, libc::SO_TIMESTAMPING) => {
+                            // Safety:
+                            // New ensures we have valid control messages
+                            // SO_TIMESTAMPING always has a timespec in the data
+                            let cmsg_data =
+                                unsafe { libc::CMSG_DATA(current_msg) } as *const libc::timespec;
+                            let timespec = unsafe { std::ptr::read_unaligned(cmsg_data) };
```
djc, line 234: Why/how do we know we need unaligned reads here?

ntp-udp/src/lib.rs
```diff
@@ -1,219 +1,424 @@
...
+        impl<'a> Iterator for ControlMessageIterator<'a> {
+            type Item = ControlMessage;
+
+            fn next(&mut self) -> Option<Self::Item> {
+                // Safety:
+                // CMSG_FIRSTHDR and CMSG_NXTHDR only return valid pointers or NULL when given valid input
+                let current_msg = unsafe { self.current_msg.as_ref() };
+                if let Some(current_msg) = current_msg {
+                    // Safety:
+                    // New ensure mhdr is valid
```
djc, line 223: It's not clear to me what is meant by "New ensure" at different levels.

ntp-udp/src/lib.rs
```diff
@@ -1,219 +1,424 @@
...
+            rx_filters: u32,
+            rx_reserved: [u32; 3],
+        }
 
-            let support = Self {
-                rx_software: tsi.so_timestamping & libc::SOF_TIMESTAMPING_RX_SOFTWARE != 0,
-                tx_software: tsi.so_timestamping & libc::SOF_TIMESTAMPING_TX_SOFTWARE != 0,
-            };
+        #[repr(C)]
+        union ifr_ifru {
```
djc, line 300: Where is this defined?

ntp-udp/src/lib.rs
```diff
@@ -1,219 +1,424 @@
...
-            //
-            // > Request rx timestamps when data enters the kernel. These timestamps are generated
-            // > just after a device driver hands a packet to the kernel receive stack.
-            //
-            // the linux kernal should always support receive software timestamping
-            assert!(support.rx_software);
+        #[repr(C)]
+        #[allow(non_camel_case_types)]
+        #[derive(Clone, Copy)]
+        struct ifmap {
```
djc, line 319: And this?

ntp-udp/src/lib.rs
folkertdev: rather, we don't for sure know that the pointer is aligned. We could assume it, but why take the risk. `read_unaligned` will have the same performance as a `read` when the address does turn out to be aligned.

ntp-udp/src/lib.rs
folkertdev: https://github.com/torvalds/linux/blob/master/include/uapi/linux/if.h

I'm trying to get that into the `libc` crate, because these types  are a bit nasty and we only use tiny parts of these types, yet have to implement them fully so  they are ABI-compatible.

ntp-udp/src/lib.rs
djc: Fair enough.

ntp-udp/src/lib.rs
djc: As long as it's not in the libc crate (though probably insofar as you're trying to upstream it), would be good to add a link to that near the definition.

ntp-udp/src/lib.rs
davidv1992: msg_flags is only guaranteed to be valid when recvmsg returns no error. However, this is guaranteed by the question mark operator on line 153.

ntp-udp/src/lib.rs
```diff
@@ -1,219 +1,424 @@
...
+                    | libc::SOF_TIMESTAMPING_OPT_TSONLY
+                    | libc::SOF_TIMESTAMPING_OPT_ID;
             }
 
-            other => return other,
+            // for documentation on SO_TIMESTAMPING see
+            // https://www.kernel.org/doc/Documentation/networking/timestamping.txt
+            // Safety:
+            // we have a reference to the socket, so fd is a valid file descriptor for the duration of the call
+            // SOL_SOCKET + SO_TIMESTAMPING expect a *u32 as value, which &options is. Furthermore, we own options
```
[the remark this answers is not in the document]
davidv1992: Added reference in the safety comment. Unfortunately this is not that well documented actually, so that section (and its code sample) really is all we base this on.

ntp-udp/src/lib.rs
davidv1992: done

ntp-udp/src/lib.rs
davidv1992: done

ntp-udp/src/lib.rs
davidv1992: I agree that the previous explanation was unclear, have rewritten the entire safety argument around a set of two invariants holding for the `ControlMessageIterator` struct.

ntp-udp/src/lib.rs
djc: Ahh -- sorry, I missed the question mark. Might consider moving the question mark to the `break other` expression? Honestly not sure if that would be an improvement.

ntp-udp/src/lib.rs
```diff
@@ -1,219 +1,447 @@
...
-    (unsafe { libc::CMSG_SPACE((std::mem::size_of::<T>()) as _) }) as usize
-}
+        // Invariants:
+        // self.mhdr points to a valid libc::msghdr with a valid control
+        // message region.
+        // self.current_msg points to one of the control messages
+        // in the region described by self.mhdr or is NULL
+        //
+        // These invariants are guaranteed from the safety conditions on
+        // callingControlMessageIterator::new, the fact that next preserves
```
djc, line 190: Nit: "callingControlMessageIterator::new".

djc (approved):

ntp-udp/src/lib.rs
davidv1992: Hmm, you have a point that this was getting somewhat confusing to read. Decided to resolve that by refactoring the loop back out into its own function, which makes the question of where to put the question mark moot.

davidv1992: Do you agree this is an improvement @djc?

ntp-udp/src/lib.rs
davidv1992: fixed

ntp-udp/src/lib.rs
djc: IMO it's better to keep the loop inline, just adding indirection (which also means you have to scroll down to be able to get the full picture) doesn't make it easier. There are probably other ways to make it more obvious (for example, you could just explicit arms for the remaining `Err(_)` and `Ok(_)` cases).

davidv1992: done

folkertdev:
I added some links to `ifreq` and friends (linking to their definition in https://github.com/torvalds/linux/blob/master/include/uapi/linux/if.h). I think those got lost in the force-pushing. Can you add them back?

ntp-udp/src/lib.rs
```diff
@@ -1,219 +1,447 @@
...
+        // self.current_msg points to one of the control messages
+        // in the region described by self.mhdr or is NULL
+        //
+        // These invariants are guaranteed from the safety conditions on
+        // calling ControlMessageIterator::new, the fact that next preserves
+        // these invariants and that the fields of ControlMessageIterator
+        // are not modified outside these two functions.
+        struct ControlMessageIterator<'a> {
+            mhdr: libc::msghdr,
+            current_msg: *const libc::cmsghdr,
```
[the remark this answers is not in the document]
davidv1992: Done

davidv1992:
> I added some links to `ifreq` and friends (linking to their definition in https://github.com/torvalds/linux/blob/master/include/uapi/linux/if.h). I think those got lost in the force-pushing. Can you add them back?

Done, couldn't find the original pr though so recreated them, can you check they are ok in this form @folkertdev

folkertdev:
yes looks good

folkertdev (approved):

PR #290, opened 2022-09-20
use constants that are now in libc

folkertdev:
we've upgraded libc, which now includes some constants that we previously defined ourselves


davidv1992 (approved):
Good spot

PR #291, opened 2022-09-20
Server filter tweaks

folkertdev:


ntp-daemon/src/ipfilter.rs
```diff
@@ -83,57 +89,45 @@ impl BitTree {
...
-        // Actually split into the relevant subsegments
-        let mut segs = <[&mut [(u128, u8)]; 16]>::default();
-        for i in 0..15 {
-            (segs[i], data) = data.split_at_mut(first_idx[i + 1] - first_idx[i]);
+        // Actually split into the relevant subsegments, relies on the input being sorted.
+        let mut subsegments: [&mut [(u128, u8)]; 16] = Default::default();
+        for (i, start) in counts.iter().enumerate() {
+            (subsegments[i], data) = data.split_at_mut(*start);
         }
-        segs[15] = data;
```
folkertdev, line 108: not the nices diff, but I was able to simplify the logic a bit here

ntp-daemon/src/ipfilter.rs
```diff
@@ -143,25 +137,33 @@ impl BitTree {
...
-        // allocate nodes
-        self.nodes.resize(
-            self.nodes.len() + (!known_bitmap).count_ones() as usize,
-            TreeNode::default(),
-        );
+
+        // allocate additional empty nodes
+        let unknown_count = known_bitmap.count_zeros() as usize;
+        self.nodes
+            .extend(std::iter::repeat(TreeNode::default()).take(unknown_count));
```
folkertdev, line 151: I think `extend` captures the idea here better than `resize`

davidv1992: Hmm, think this might lead to worse performance in the generated code, but since this data structure already chooses to sacrifice performance at create time and this will be at most 16 nodes, I think the tradeoff is worth it.

PR #296, opened 2022-09-22
pool attempt 2

folkertdev:


ntp-daemon/src/peer_manager.rs
```diff
@@ -124,8 +170,54 @@ impl<C: NtpClock> Peers<C> {
...
+    async fn add_peer_internal(&mut self, config: PeerConfig) -> Vec<JoinHandle<()>> {
+        let new_active = match &config {
+            PeerConfig::Standard(StandardPeerConfig { addr, .. }) => {
+                // unwrap is safe because of the peek() in `socket_addresses`
+                let address = socket_addresses(addr).await.next().unwrap();
+
+                vec![address]
+            }
+            PeerConfig::Pool(PoolPeerConfig { hosts, max_peers }) => {
+                let pool_status = self.pools.entry(hosts.clone()).or_default();
```
davidv1992, line 182: Why is it not a problem if the config doesn't match to a pool, that seems like a critical failure to me?

davidv1992 (commented):
The code has a few quite large gotchas which just need better documentation.

Also in general the design to me doesn't feel natural to the problem domain. I still think it feels more natural to have the pool config far more in charge of managing how the pool's peers are instantiated, instead of having that spread across the manager and a separate state object.

davidv1992:
Also, I am missing the documentation updates on how to use pools. At least a minimal example config would be very useful as part of this PR

PR #297, opened 2022-09-22
rate limiting

folkertdev:


ntp-daemon/src/server.rs
```diff
@@ -160,3 +154,86 @@ fn accept_packet(
...
+#[derive(Debug)]
+struct TimestampedCache<T> {
+    elements: Vec<Option<(T, std::time::Instant)>>,
+}
+
+impl<T: std::hash::Hash + Eq> TimestampedCache<T> {
+    fn new(length: usize) -> Self {
+        Self {
+            // looks a bit odd, but prevents a `Clone` constraint
+            elements: std::iter::repeat_with(|| None).take(length).collect(),
```
djc, line 167: Try `vec![None; length]` or `[None; length].into_iter().collect()`?

ntp-daemon/src/server.rs
```diff
@@ -160,3 +154,86 @@ fn accept_packet(
         }
     }
 }
+
+#[derive(Debug)]
+struct TimestampedCache<T> {
+    elements: Vec<Option<(T, std::time::Instant)>>,
```
djc, line 160: All the qualified types look pretty unidiomatic to me, any reason for that?

Also, this could use some documentation on what the design is here. Is it a fixed-length compact hash map of some sort? What is it's purpose? What are we trying to optimize for?

ntp-daemon/src/server.rs
```diff
@@ -160,3 +154,86 @@ fn accept_packet(
...
+
+    fn is_allowed(
+        &mut self,
+        item: T,
+        timestamp: std::time::Instant,
+        cutoff: std::time::Duration,
+    ) -> bool {
+        match self.get(&item) {
+            None => {
+                self.insert(item, timestamp);
```
djc, line 204: Deduplicate the `insert()` call?

ntp-daemon/src/server.rs
```diff
@@ -160,3 +154,86 @@ fn accept_packet(
...
+
+    fn insert(&mut self, item: T, timestamp: std::time::Instant) {
+        let index = self.index(&item);
+        self.elements[index] = Some((item, timestamp));
+    }
+
+    fn get(&self, item: &T) -> Option<std::time::Instant> {
+        let (existing, timestamp) = self.elements[self.index(item)].as_ref()?;
+
+        if existing == item {
```
djc, line 189: Alternatively, `(existing == item).then_some(*timestamp)`.

ntp-daemon/src/server.rs
folkertdev: yeah, I thought that would work but it doesn't . `length` is not a constant so the array does not work. The vec just does not work. I tried adding a `    const EMPTY: Option<(T, Instant)> = None;` to the impl block but no luck.

ntp-daemon/src/server.rs
folkertdev: ah, forgot about that one

ntp-daemon/src/server.rs
folkertdev: that didn't quite work (because of T not being clone) but I did some refactoring

ntp-daemon/src/server.rs
folkertdev: The planned use is in rate limiting: we keep track of when a peer last checked in. If it checks in too often, we issue a rate limiting KISS code.

A standard hashmap could become very large when we have many connections or when the server has been running for a long time (or you would have to garbage collect it periodically, which has its own complications). We thought about an adapter type with a custom hash instance (that takes the modulo), but this seemed simpler.

djc:
Hmm, I'm still uncomfortable with this data structure. It seems like an optimization (making the code harder to follow/less robust in order to improve some aspect of performance) without clear rationale (why this needs to be optimized -- ideally including benchmarks).

For example, on collisions, the comment just mentions that "collisions should not be a big problem", without explaining why. I'm worried that, if collisions happen, it will be relatively hard to detect and/or remedy the problem.

Approaches that I think would be better:

* Use a real `HashMap` and define a maximum size; if the maximum size is reached, iterate over the items and eject the oldest half of the items.
* If the cardinality of T is small, it might make sense to just use a `Vec` and just do a linear scan for each key on every check.
* There are perfectly serviceable LRU maps in external crates.

folkertdev:
our goal here is rate limiting. With possibly thousands of requests per second, we need both lookups and inserts to be fast, ideally constant time.

Hash collisions are not a big deal because a problematic client will show up again. We can vary the size of the data structure to ensure that the number of hash collisions is acceptable (we have our server running in the NTP pool right now to see what traffic is like in practice).

So then, regarding your alternative approaches

- Hashmap dynamically grows/reallocates. Not something you want in a hot loop
- the cardinality is not small, so a plain vector won't do.
- LRU cache would be an additional dependency. Why do that if ~30 lines of logic on top of a `Vec` suffice?

Just a vector and some custom insert/lookup logic seems by far the easiest. With all your alternatives, we would still need this custom logic.

djc:
> * Hashmap dynamically grows/reallocates. Not something you want in a hot loop

A `Vec` also dynamically grows, but you've made an implementation where it shouldn't. My proposal is similar in that the code would have control over when you reallocate.

Also, reallocation would be amortized over the course of many inserts. Again, have you done benchmarks to verify that your expected performance impact is correct?

folkertdev:
the implementation never (re)allocates the data structure. Its API does not support that. With a hashmap, you cannot prevent it because you can't statically know in which buckets the elements will end up. 

I need convincing that a hashmap is better here. I don't see why it would be, and I don't think the code would actually be simpler or shorter.

djc:
Okay, never mind.

davidv1992 (commented):
The datastructure part looks good to me, and I think the custom choice here is the better choice as we then have good certainty that we don't accidentally insert any potential for large reallocations at runtime (which would pose DOS risks).

I am however not entirely convinced with how this integrates with the server code, especially around when and how it now incorporates the rate limiting. For one, this is taking quite a different aproach to the where and when from the Block/allow list stuff, and I am not entirely sure why. Also, it seems to me we are doing the rate limiting rather late, when only the response generation is still needed, whilst we might save some processing when pulling it forward (again, see also the block/allow list stuff).

ntp-daemon/src/server.rs
davidv1992: Agreed, bit better documentation of that would be better.

ntp-proto/src/packet.rs
```diff
@@ -242,6 +242,36 @@ impl NtpHeader {
...
+        }
+    }
+
+    pub fn rate_limit_response() -> Self {
+        Self {
+            stratum: 0, // indicates a kiss code
+            reference_id: ReferenceId::KISS_RATE,
+            ..Self::new()
+        }
+    }
```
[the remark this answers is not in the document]
folkertdev: counterpoint: these functions are about data, not about logic. Therefore I think they belong in `proto`. The peer/server to me are about orchestrating the various connections. Exactly how to build a piece of data distracts from the logic that is happening there.

ntp-proto/src/packet.rs
```diff
@@ -242,6 +242,36 @@ impl NtpHeader {
...
+            transmit_timestamp: clock.now().expect("Failed to read time"),
+            ..Self::new()
+        }
+    }
+
+    pub fn rate_limit_response() -> Self {
+        Self {
+            stratum: 0, // indicates a kiss code
+            reference_id: ReferenceId::KISS_RATE,
+            ..Self::new()
```
davidv1992, line 272: Missing the origin timestamp setting here. Note that this is required to be done otherwise the remote will ignore the rate as being not a response to its packet.

ntp-proto/src/packet.rs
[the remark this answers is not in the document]
davidv1992: ok

davidv1992 (commented):
Mostly looks good, one minor comment below.

Also, given the move of packet generation logic, I would expect generate_deny (ntp-daemon/src/server.rs lines 68-76) to also be moved.

folkertdev:
all fixed

PR #303, opened 2022-09-29
Added observability for the peer measurement buffer.

davidv1992:


davidv1992:
Given recent experience with the pool server, I don't think we need this.

PR #317, opened 2022-10-06
parse, don't validate normalized addresses

folkertdev:


ntp-daemon/src/config/peer.rs
```diff
@@ -47,34 +43,55 @@ impl PeerConfig {
...
-                // Could be because of no port, add one and see
-                addr = format!("[{addr}]:123");
-                if addr.parse::<SocketAddr>().is_ok() {
-                    Ok(addr)
-                } else {
-                    Err(std::io::Error::new(std::io::ErrorKind::Other, e))
+/// A normalized address has a host and a port part. However, the host may be
+/// invalid, we didn't yet perform a DNS lookup.
+#[derive(Deserialize, Debug, PartialEq, Eq, Clone)]
+pub struct NormalizedAddress(String);
```
davidv1992, line 49: Since this purports to ensure some properties are checked, I think we should use a private field here (and hence not a tuple struct).

davidv1992 (commented):
This is a very good idea, one minor comment on data hiding in the type,

PR #327, opened 2022-10-12
Update dependabot.yml

jauderho:
Keep GitHub Actions up to date.


davidv1992 (approved):
This seems like a good idea.

PR #336, opened 2022-10-14
Pool attempt 3

folkertdev:


ntp-daemon/src/peer_manager.rs
```diff
@@ -78,41 +153,45 @@ impl<C: NtpClock> Peers<C> {
...
+                } else {
+                    None
                 }
             }
+        });
+
+        // if no socket addresses are cached, do a new DNS resolve
+        let addr = match cached.find_additional(&address, active_pool_peers).await {
+            Some(addr) => addr,
+            None => todo!("the pool is now smaller than max_peers; what to do?"),
```
folkertdev, line 183: I think we're kind of stuck at this point. We want to add another peer, but there just aren't any left in the pool. Any thoughts?

ntp-daemon/src/peer_manager.rs
```diff
@@ -78,41 +153,45 @@ impl<C: NtpClock> Peers<C> {
...
+                if &address == peer_address {
+                    Some(*socket_address)
+                } else {
+                    None
                 }
             }
+        });
+
+        // if no socket addresses are cached, do a new DNS resolve
+        let addr = match cached.find_additional(&address, active_pool_peers).await {
```
folkertdev, line 181: I think there is a race condition here where `cached` is not locked, but the `addr` is not yet added to `peers` either. That means that technically another thread could pick the same `addr`, and we duplicate a peer. However, we cannot keep `cached` locked because it is part of the `PeerAddress::Pool` below. Not sure how to deal with that yet.

davidv1992: I think this is actually an indication that the state is still too fragmented. Perhaps it is better to reorganize everything such that the list of active pool peers and the cache of unused addresses are part of the same mutex. Access/modification of them is related already so from a performance standpoint that is not that problematic, and it makes correctness like here easier.

ntp-daemon/src/peer_manager.rs
davidv1992: I see two potential options. My preference is to just print an error here letting the user know they borked up their configuration and letting them fix it through a restart.

We could instead consider starting a background process that tries to fix this after some time has passed (together ofcourse with a warning or error), but I think that is a bad idea as it is fairly complex and adds a significant risk of actually starting to ddos the pools dns servers should this situation ever occur.

folkertdev: sounds good! I'll add some error logging and then this function needs to return an `Option`

ntp-daemon/src/peer_manager.rs
folkertdev: that is kind of what `attempt-2` did? unless you're implying that pool peers should not be in the list of peers any more?

davidv1992: No they definitely should be, I just think that the identification of which of the peers are part of the pool should be part of the pool state.

davidv1992: However, now that I look into more detail here, the race condition is a bit unclear to me. The list of peers cant change when this function runs outside of what this function does since we have a self reference.

ntp-daemon/src/peer_manager.rs
```diff
@@ -78,41 +153,69 @@ impl<C: NtpClock> Peers<C> {
...
+            PeerAddress::Peer { .. } => None,
+            PeerAddress::Pool {
+                address: peer_address,
+                socket_address,
+                ..
+            } => {
+                if &address == peer_address {
+                    Some(*socket_address)
+                } else {
+                    None
```
davidv1992, line 175: Why are we matching on peer address here, instead of just checking whether they use the same peerconfig?

ntp-daemon/src/peer_manager.rs
folkertdev: I think that's right, in practice. In theory the `Arc` could have been cloned, but all clones are contained in the `Peers` structure so we should be in the clear.

ntp-daemon/src/peer_manager.rs
folkertdev: this suggests we really want a pool index, similar to how we already have a peer index: some unique identifier for a pool. As you say, the normalized address is not guaranteed to be unique (though it's weird if it isn't)

davidv1992: I dont think this needs to be an index per see, if all peers have an arc to the same peerconfig then simple pointer equality on that would suffice. Or perhaps the peerconfig could contain a list of the peerindices associated with it?

folkertdev: idk, I implemented the index and I think that works very nicely. Means we don't need additional `Arc` stuff

ntp-daemon/src/config/peer.rs
```diff
@@ -91,12 +106,36 @@ impl NormalizedAddress {
...
+    #[cfg(test)]
+    pub async fn lookup_host(&self) -> std::io::Result<impl Iterator<Item = SocketAddr> + '_> {
+        let addresses = if !self.hardcoded_dns_resolve.is_empty() {
+            self.hardcoded_dns_resolve.to_vec()
+        } else {
+            tokio::net::lookup_host(&self.address).await?.collect()
+        };
+
+        Ok(addresses.into_iter())
+    }
```
davidv1992, line 138: Although I see how this came about, I feel kind of uncomfortable that this leaves us with a function we could effectively never exercise in (non-integration) tests.

rnijveld (approved):

PR #346, opened 2022-10-20
Packet refactor

davidv1992:


test-binaries/src/bin/rate-limit-server.rs
```diff
@@ -11,6 +11,8 @@ async fn main() -> Result<(), Box<dyn Error>> {
     let clock = ntp_os_clock::UnixNtpClock::new();
     let mut last_message = Instant::now();
 
+    let system = SystemSnapshot::default();
+
     let mut buf = [0; 48];
     #[allow(clippy::field_reassign_with_default)] // allow the explicit stratum
```
rnijveld, line 17: I don't think this is needed anymore?

davidv1992: removed

PR #347, opened 2022-10-20
Made poll interval limits and initial value configurable.

davidv1992:


ntp-proto/src/clock.rs
```diff
@@ -63,7 +63,7 @@ pub enum ClockUpdateResult {
 }
 
 impl<C: NtpClock> ClockController<C> {
-    pub fn new(clock: C, system: &SystemSnapshot) -> Self {
+    pub fn new(clock: C, system: &SystemSnapshot, config: &SystemConfig) -> Self {
```
rnijveld, line 66: I don't really like all that passing around of the config, now that it also is required here and in `do_step`. I do get that we need to update it when the config is changed, but that would only happen very irregularly right? Any opinions?

davidv1992: Fair enough, but that would mean moving back to a channel for the config, which originally we decided to like less since this acted (and still does) more like shared data than something channeled.

rnijveld: Yeah, that might be too much without enough gains, but I feel that in future we might not get around this as I think we'll see this config pop up in more and more places.

PR #386, opened 2022-11-13
Updated sentry dependencies.

davidv1992:
Replacement for both #381 and #383 to deal with the pipeline failures in those.


davidv1992:
Never mind this attempt, it seems there is a more fundamental issue with a dependency on chrono in sentry-types.

PR #392, opened 2022-11-17
datastructures for NTS key exchange

folkertdev:


ntp-proto/src/record.rs
```diff
@@ -0,0 +1,320 @@
...
+                    algorithm_ids,
+                }
+            }
+            5 if !critical => {
+                let cookie_data = read_bytes_exact(reader, record_len)?;
+                Record::NewCookie { cookie_data }
+            }
+            6 => {
+                let str_data = read_bytes_exact(reader, record_len)?;
+                match String::from_utf8(str_data) {
```
davidv1992, line 194: Spec says str_data must only ever be ascii, do we want to enforce that here?

ntp-proto/src/record.rs
```diff
@@ -0,0 +1,320 @@
...
+                let length = cookie_data.len() as u16;
+                writer.write_all(&length.to_be_bytes())?;
+
+                writer.write_all(cookie_data)?;
+            }
+            Record::Server { name, .. } => {
+                let length = name.len() as u16;
+                writer.write_all(&length.to_be_bytes())?;
+
+                writer.write_all(name.as_bytes())?;
```
davidv1992, line 266: Again, spec says name should only consist of ascii, should we enforce that here?

davidv1992 (commented):
In general, looks good. I have left two minor comments around encoding of server names we should document our decision about. Also, one minor nit, the current filename record.rs doesn't clearly indicate this is for nts.

In general though this looks good.

davidv1992 (approved):

PR #393, opened 2022-11-17
async NTS KE binary example

folkertdev:


.github/workflows/build.yaml
```diff
@@ -144,27 +144,29 @@ jobs:
...
+      #          toolchain: stable
+      #          override: true
+      #          default: true
+      #          components: clippy
+      #          target: armv7-unknown-linux-gnueabihf
+      #      - name: Run clippy
+      #        uses: actions-rs/cargo@844f36862e911db73fe0815f00a4a2602c279505
+      #        with:
+      #          command: clippy
+      #          args: --target armv7-unknown-linux-gnueabihf --workspace --all-targets -- -D warnings
```
folkertdev, line 169: @djc are you aware of an easy way to cross-compile rustls (or really ring, a dependency of rustls) to raspberry pi on CI? it looks like it needs a custom gcc binary, but I could not find a convenient way of installing that from a package manager. Does the rust component for `armv7-unknown-linux-gnueabihf` make this binary available somehow?

folkertdev: or actually I'd expect `clang` in this case, which based on `ring`'s documentation should also work 

https://github.com/briansmith/ring/issues/1559
https://gist.github.com/fm4dd/c663217935dc17f0fc73c9c81b0aa845

folkertdev: zig to the rescue! we're now using `cargo zigbuild` to provide a valid C compiler for the raspberry pi target

djc: Makes sense. Did you consider [`cross`](https://github.com/japaric/rust-cross) as well?

.github/workflows/build.yaml
```diff
@@ -160,8 +160,21 @@ jobs:
           default: true
           components: clippy
           target: armv7-unknown-linux-gnueabihf
+      # Use zig as our C compiler for convenient cross-compilation. We run into rustls having a dependency on `ring`.
+      # This crate uses C and assembly code, and because of its build scripts, `cargo clippy` needs to be able to compile
+      # that code for our target.
+      - uses: goto-bus-stop/setup-zig@v2
```
davidv1992, line 166: This should be pinned to a specific commit to be consistent with the other actions.

zigcc-armv7-unknown-linux-gnueabihf.sh
```diff
@@ -0,0 +1,2 @@
+#!/usr/bin/env sh
+/home/runner/.cargo/bin/cargo-zigbuild zig cc -- -target arm-linux-gnueabihf -mcpu=generic+v7a+vfp3-d32+thumb2-neon -g $@
```
davidv1992, line 2: What is this script for? And why does it need to be in the project root?

davidv1992 (commented):
Code wise this looks good to me. I have some comments on the organization of the ci pipeline stuff.

zigcc-armv7-unknown-linux-gnueabihf.sh
folkertdev: that was a failed experiment. removed now

davidv1992 (approved):

PR #409, opened 2022-11-24
Added some docs for the prometheus metrics exporter

### LINKED ISSUE #404: Create documentation for the prometheus exporter

rnijveld:
Fixes #404

### DIFF
--- a/MANAGEMENT_CLIENT.md
+++ b/MANAGEMENT_CLIENT.md
@@ -171,6 +171,6 @@ HTTP server and serve the metrics on the `/metrics` endpoint. By default the
 metrics exporter listens on localhost port 9975, but this can be changed via
 command line parameters, see `ntp-metrics-exporter --help` for details. Note
 that the metrics exporter does not do any authentication or HTTPS, so if the
-metrics are transfered via a public network you should add a reverse proxy that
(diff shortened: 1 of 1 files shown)


MANAGEMENT_CLIENT.md
```diff
@@ -163,3 +163,14 @@ ntp_server_response_send_errors_total{listen_address="127.0.0.1:123"} 0
...
 ```
+
+## Prometheus metrics exporter
+Prometheus prefers to use a pull-based architecture via HTTP. To facilitate this
+a separate executable `ntp-metrics-exporter` can be used. This will start up a
+HTTP server and serve the metrics on the `/metrics` endpoint. By default the
+metrics exporter listens on localhost port 9975, but this can be changed via
+command line parameters, see `ntp-metrics-exporter --help` for details. Note
+that the metrics exporter does not do any authentication or HTTPS, so if the
+metrics are transfered via a public network you should add a reverse proxy that
```
davidv1992, line 174:
```suggestion
metrics are transferred via a public network you should add a reverse proxy that
```

davidv1992 (commented):
One minor spelling error, other than that, looks good to me.

PR #412, opened 2022-11-25
Refactored system::run to be a bit more self-contained.

davidv1992:


ntp-daemon/src/system.rs
```diff
@@ -210,32 +184,93 @@ impl<C: NtpClock> System<C> {
...
+        measurement: ntp_proto::Measurement,
+        packet: ntp_proto::NtpPacket<'static>,
+    ) {
+        self.handle_peer_snapshot(index, snapshot);
+        let result = self.controller.peer_measurement(index, measurement, packet);
+        if let Some((used_peers, timedata)) = result {
+            self.system.update(
+                used_peers
+                    .iter()
+                    .map(|v| self.peers.get(v).and_then(|data| data.snapshot).unwrap()),
```
folkertdev, line 259: is `unwrap`  the right call here (`filter_map` would be another option)

folkertdev (approved):

ntp-daemon/src/system.rs
davidv1992: It should be a critical error if a peer is used for synchronization that system doesn't know anything about, hence the unwrap.

You are right it is a bit awkward though, I'll make it an expect that also conveys a bit more about what went wrong.

folkertdev (approved):

PR #420, opened 2022-12-01
extension fields (de)serialization

folkertdev:
I've run fuzz testing on this a bunch. Probably still needs more though, because very few bit patterns are interesting extension fields


ntp-proto/src/packet.rs
```diff
@@ -113,25 +113,134 @@ pub struct NtpPacket<'a> {
...
 
-#[derive(Debug, Clone, PartialEq, Eq)]
+#[derive(Clone, PartialEq, Eq)]
 pub enum ExtensionField<'a> {
-    Unknown { typeid: u16, data: Cow<'a, [u8]> },
+    UniqueIdentifier(Cow<'a, [u8]>),
+    NtsCookie(Cow<'a, [u8]>),
+    NtsCookiePlaceholder {
+        body_length: u16,
+    },
```
folkertdev, line 122: maybe we should just store the input here? we'll have to see when we start using this field

ntp-proto/src/packet.rs
```diff
@@ -113,25 +113,134 @@ pub struct NtpPacket<'a> {
...
+                .finish(),
+            Self::NtsEncryptedField { nonce, ciphertext } => f
+                .debug_struct("NtsEncryptedField")
+                .field("nonce", nonce)
+                .field("ciphertext", ciphertext)
+                .finish(),
+            Self::Unknown { typeid, data } => f
+                .debug_struct("Unknown")
+                .field("typeid", typeid)
+                .field("length", &data.len())
```
folkertdev, line 151: this is the custom field here, it's useful for debugging

ntp-proto/src/packet.rs
```diff
@@ -140,41 +249,115 @@ impl<'a> ExtensionField<'a> {
...
         let typeid = u16::from_be_bytes(data[0..2].try_into().unwrap());
         let ef_len = u16::from_be_bytes(data[2..4].try_into().unwrap()) as usize;
-        if ef_len < Self::MINIMUM_SIZE || ef_len > data.len() {
+        if ef_len < Self::MINIMUM_SIZE {
             return Err(PacketParsingError::IncorrectLength);
         }
-        Ok((
-            ExtensionField::Unknown {
+
+        let value = data.get(4..ef_len).ok_or(IncorrectLength)?;
```
folkertdev, line 276: the removed length check above is now performed by the `get`

ntp-proto/src/packet.rs
```diff
@@ -140,41 +249,115 @@ impl<'a> ExtensionField<'a> {
...
     }
 }
 
 #[derive(Debug, Clone, PartialEq, Eq)]
 enum ExtensionFieldData<'a> {
-    Raw(Cow<'a, [u8]>),
     #[allow(dead_code)]
+    Raw(Cow<'a, [u8]>),
     List(Vec<ExtensionField<'a>>),
 }
```
folkertdev, line 355: is it useful to keep the raw option around here?

davidv1992 (changes requested):
The aproach taken here places the cyphertext itself in the extensionfield. Given how tightly the interpretation of the cyphertext is coupled to the serialized representation of the packet, I think we shouldn't do this, but instead generate/decrypt it on serialization/deserialization.

This will likely require quite a bit of refactoring of the entire packet parsing logic, but will likely make it much easier for us to do the crypto right. Also, enforcing the crypto on serialization/deserialization also has the added benefit of providing a harder to misuse layer to the rest of the application.

ntp-proto/src/packet.rs
```diff
@@ -140,41 +269,135 @@ impl<'a> ExtensionField<'a> {
...
+                }
+
+                let ciphertext = value
+                    .get(ciphertext_start..ciphertext_start + ciphertext_length)
+                    .ok_or(IncorrectLength)?;
+
+                let plaintext = match cipher.decrypt(Nonce::from_slice(nonce), ciphertext) {
+                    Ok(plain) => plain,
+                    Err(e) => panic!("{e:?}"),
+                };
```
folkertdev, line 365: this allocates. Doing the decryption in-place does not work with the rest of the code (because the data behind the mutable reference is then not available for further decoding. 

We could pass in a "scratchpad" buffer in the future (using some sort of pool allocator) to minimize the allocation cost. I'm not sure that's worth it, but it's an option.

ntp-proto/src/packet.rs
```diff
@@ -113,25 +114,162 @@ pub struct NtpPacket<'a> {
...
+                let mut remaining = next_multiple_of(*body_length as usize, 4);
+                while remaining > 0 {
+                    let n = usize::min(zeros.len(), remaining);
+                    w.write_all(&zeros[..n])?;
+                    remaining -= n;
+                }
+            }
+            ExtensionField::NtsEncryptedField {
+                nonce,
+                plaintext: _,
```
folkertdev, line 242: this is kind of weird. The plaintext is implicit when encoding, but I suspect you'll want to be able to access it when decoding (otherwise what is even the point).

fuzz/Cargo.toml
```diff
@@ -20,6 +20,9 @@ features = ["fuzz"]
 path = "../ntp-daemon"
 features = ["fuzz"]
 
+[dependencies.pretty_assertions]
+version = "1.3.0"
+
```
davidv1992, line 25: Do we really need this additional crate? I know we have strayed quite a bit from the original target of keeping the number of dependencies down, but this feels somewhat superfluous to me.

ntp-proto/src/packet.rs
```diff
@@ -113,25 +114,162 @@ pub struct NtpPacket<'a> {
...
+#[derive(Clone, PartialEq, Eq)]
 pub enum ExtensionField<'a> {
-    Unknown { typeid: u16, data: Cow<'a, [u8]> },
+    UniqueIdentifier(Cow<'a, [u8]>),
+    NtsCookie(Cow<'a, [u8]>),
+    NtsCookiePlaceholder {
+        body_length: u16,
+    },
+    NtsEncryptedField {
+        nonce: Cow<'a, [u8]>,
```
davidv1992, line 125: The nonce is essentially part of the details of encryption/decryption, and shouldn't be needed anywhere else, so I would suggest we just not keep it around here.

ntp-proto/src/packet.rs
```diff
@@ -113,25 +114,162 @@ pub struct NtpPacket<'a> {
...
+        }
+    }
+
+    pub fn serialize_without_encryption(&self, w: &mut Cursor<&mut [u8]>) -> std::io::Result<()> {
+        use aes_siv::{aead::KeyInit, Key};
+
+        let cipher = Aes256SivAead::new(Key::<Aes256SivAead>::from_slice([0; 64].as_slice()));
+
+        self.serialize(w, &cipher)
+    }
```
davidv1992, line 202: This feels like a very insecure function to have available as part of the public non-testing interface. Why is it needed?

ntp-proto/src/packet.rs
```diff
@@ -113,25 +114,162 @@ pub struct NtpPacket<'a> {
...
-    Unknown { typeid: u16, data: Cow<'a, [u8]> },
+    UniqueIdentifier(Cow<'a, [u8]>),
+    NtsCookie(Cow<'a, [u8]>),
+    NtsCookiePlaceholder {
+        body_length: u16,
+    },
+    NtsEncryptedField {
+        nonce: Cow<'a, [u8]>,
+        plaintext: Cow<'a, [u8]>,
+    },
```
davidv1992, line 127: Do we want to keep this in our datastructures as an explicit extension field? An NTS packet logically should never have more than one of these, and its existence essentially creates just a split in the extension fields (into open but signed, open and unsiged (if occuring after the nts encryption field) and encrypted (the plaintext in this design). It may be a better option to just handle that split on the main packet datastructure.

ntp-proto/src/packet.rs
```diff
@@ -140,41 +278,135 @@ impl<'a> ExtensionField<'a> {
...
-            return Err(PacketParsingError::IncorrectLength);
+    pub fn deserialize_without_encryption(
+        data: &'a [u8],
+    ) -> Result<(ExtensionField<'a>, usize), PacketParsingError> {
+        use aes_siv::{aead::KeyInit, Key};
+
+        let cipher = Aes256SivAead::new(Key::<Aes256SivAead>::from_slice([0; 64].as_slice()));
+
+        Self::deserialize(data, &cipher)
+    }
```
davidv1992, line 300: Again, this feels like a footgun to me to keep around.

davidv1992 (changes requested):
Haven't placed comments on every occurence of it, but given my comments so far, I think this still needs significantly more work to get to a better (and safer!) api.

fuzz/Cargo.toml
folkertdev: it's quite difficult when you see a list of 100+ numbers to figure out where the error is. Is it needed? maybe not, but this is in our fuzzing, where I don't think it does any harm and is occasionally convenient.

ntp-proto/src/packet.rs
folkertdev: well, it has the shape of an extension field, so we at least need to parse it as one. But we could separate that from the data that we actually present to the daemon layer.

Just to clarify, the only use that we foresee for this field is its current use or similar: the "plaintext" is the message so far? there is never some custom data that we put in there?

davidv1992: No, the plaintext shouldn't be the message so far, or at least that is not what I assumed it would be. The plaintext in the crypto are the "encrypted" extension fields, (i.e. the ones that are not readable by third parties). As I read the spec, there are essentially 3 types of extension field in an nts packet. The validated ones (send in cleartext before the encryption ef), the encrypted ones (in the cyphertext of the encryption ef) and the unvalidated ones (send after the encryption ef). My suggestion here is to make that distinction explicit at the packet level.

PR #425, opened 2022-12-02
Extended clock type to also allow ignoring of kernel discipline.

davidv1992:


ntp-proto/src/clock.rs
```diff
@@ -18,4 +24,12 @@ pub trait NtpClock: Clone + Send + 'static {
         poll_interval: PollInterval,
         leap_status: NtpLeapIndicator,
     ) -> Result<(), Self::Error>;
+    // Update with NTP discipline disabled
```
rnijveld, line 27: I personally don't really like this. It feels like we are trying to abstract the clock away here even though in reality it is an essential part of the whole time sync controller you choose (i.e. in the standard algorithm it makes no sense to use the bare update, whereas in other algorithms it might make no sense to call the update_clock method)

davidv1992:
Upon further reflection, this is indeed not the direction to go, I will make a second go at this.

PR #435, opened 2022-12-08
Refactored structure of clock interface.

davidv1992:
Replacement for #425 attempting to provide a better api.

### DIFF
--- a/ntp-os-clock/src/lib.rs
+++ b/ntp-os-clock/src/lib.rs
@@ -64,6 +64,18 @@ const EMPTY_TIMEX: libc::timex = libc::timex {
     __unused11: 0,
 };
 
+fn adjtime(timex: &mut libc::timex) -> Result<(), Error> {
+    // We don't care about the time status, so the non-error
+    // information in the return value of ntp_adjtime can be ignored.
+    // The ntp_adjtime call is safe because the reference always
+    // points to a valid libc::timex.
+    if unsafe { libc::ntp_adjtime(timex as *mut _) } == -1 {
+        Err(convert_errno())
+    } else {
+        Ok(())
+    }
+}
+
 /// NTP Clock that uses the unix NTP KAPI clock functions to get/modify the
 /// current time.
 // Implementation note: this is intentionally a bare struct, the NTP Clock defined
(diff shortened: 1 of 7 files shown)


ntp-os-clock/src/lib.rs
```diff
@@ -98,109 +110,103 @@ fn duration_in_nanos(duration: NtpDuration) -> libc::c_long {
...
-        if unsafe { libc::clock_settime(libc::CLOCK_REALTIME, &tp as *const _) } == -1 {
-            return Err(convert_errno());
-        }
-        // End time critical section
+    fn enable_ntp_algorithm(&self) -> Result<(), Self::Error> {
+        let mut timex = EMPTY_TIMEX;
+        adjtime(&mut timex)?;
+        timex.modes = libc::MOD_STATUS;
+        timex.status |= libc::STA_PLL;
+        timex.status &= !libc::STA_FLL & !libc::STA_PPSTIME & !libc::STA_PPSFREQ;
```
folkertdev, line 162: could `!a & !b = !(a | b)` be used to clean this up? maybe some comments are nice here too, something like "this turns off/on the flags for ... "

ntp-proto/src/algorithm/standard/clock_controller.rs
```diff
@@ -368,6 +373,7 @@ mod tests {
         last_max_error: RefCell<Option<NtpDuration>>,
         last_poll_interval: RefCell<Option<PollInterval>>,
         last_leap_status: RefCell<Option<NtpLeapIndicator>>,
+        last_algorithm_status: RefCell<Option<bool>>,
```
folkertdev, line 376: what does that bool mean? could it be a custom type?

ntp-proto/src/clock.rs
```diff
@@ -8,14 +8,21 @@ pub trait NtpClock: Clone + Send + 'static {
...
+
+    fn error_estimate_update(
+        &self,
         est_error: NtpDuration,
         max_error: NtpDuration,
-        poll_interval: PollInterval,
-        leap_status: NtpLeapIndicator,
     ) -> Result<(), Self::Error>;
+    fn status_update(&self, leap_status: NtpLeapIndicator) -> Result<(), Self::Error>;
 }
```
folkertdev, line 28: these methods need at least some documentation

ntp-proto/src/algorithm/standard/clock_controller.rs
davidv1992: It represents whether the user of the test clock has enabled or disabled the ntp discipline. You could make this a custom type, but given that it is part of testing code, I dont see the point (it would just create more code to understand when a test fails).

ntp-os-clock/src/lib.rs
davidv1992: Done

ntp-proto/src/clock.rs
davidv1992: Done

ntp-proto/src/algorithm/standard/clock_controller.rs
folkertdev: fine, maybe rename the field to `last_ntp_discipline_enabled` then? "status = true" is rather meaningless if you're just reading the code.

davidv1992: done.

folkertdev (approved):

PR #437, opened 2022-12-09
New clock algorithm

davidv1992:


ntp-proto/src/algorithm/kalman/peer.rs
```diff
@@ -0,0 +1,1067 @@
...
+/// 1 (t2-t1)
+/// 0    0
+/// and a noise matrix given by
+/// v*(t2-t1)^3/3 v*(t2-t1)^2/2
+/// v*(t2-t1)^2/2   v*(t2-t1)
+/// where v is a constant describing how much the frequency drifts per
+/// unit of time.
+///
+/// This modules input consists of measurements containing:
+///  - the time of the measurement tm
```
folkertdev, line 29: `t_m` maybe?

ntp-proto/src/algorithm/kalman/peer.rs
```diff
@@ -0,0 +1,1067 @@
...
+///  - the measured offset d
+///  - the measured transmission delay r
+/// On these, we assume that
+///  - there is no impact from frequency differences on r
+///  - individual measurements are independent
+///
+/// This information bare is not enough to feed the kalman filter. For
```
folkertdev, line 36: "bare"? what is that?

ntp-proto/src/algorithm/kalman/peer.rs
```diff
@@ -0,0 +1,1067 @@
...
+/// this, two further pieces of information are needed: a measurement
```
folkertdev, line 37: you go on to give only one of the pieces needed

ntp-proto/src/algorithm/kalman/peer.rs
```diff
@@ -0,0 +1,1067 @@
...
+/// related to the frequency difference. Although mathematically not
+/// entirely sound, we construct the frequency measurement also using
+/// the previous measurement (which we will denote with tp and dp).
```
folkertdev, line 40: then you should motivate why this is not an entirely unreasonable choice

folkertdev (commented):
first look, mostly at the non-kalman stuff in this PR

ntp-daemon/src/system.rs
```diff
@@ -204,15 +217,35 @@ impl<C: NtpClock> System<C> {
         self.config = config;
     }
 
-    async fn handle_peer_update(&mut self, msg: MsgForSystem) -> std::io::Result<()> {
+    fn handle_timer(&mut self, wait: &mut Pin<&mut T>, wait_enabled: &mut bool) {
```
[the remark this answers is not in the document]
davidv1992: Fixed with SingleshotSleep future

ntp-proto/src/algorithm/mod.rs
```diff
@@ -18,6 +18,29 @@ pub struct ObservablePeerTimedata {
     pub last_update: NtpTimestamp,
 }
 
+#[derive(Debug, Clone)]
+pub struct StateUpdate<PeerID: Eq + Copy + Debug> {
```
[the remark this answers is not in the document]
davidv1992: By whom? I am not familiar with this convention and personally prefer to have them on the datatype as well to make it conceptually clear that this data type doesn't make sense without these bounds.

ntp-proto/src/algorithm/kalman/config.rs
```diff
@@ -0,0 +1,219 @@
+use serde::Deserialize;
+
+#[derive(Debug, Copy, Clone, Deserialize)]
+pub struct AlgorithmConfig {
```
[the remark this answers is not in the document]
davidv1992: Plan to do that as part of the end user documentation.

ntp-proto/src/algorithm/kalman/peer.rs
```diff
@@ -0,0 +1,1067 @@
...
+};
+
+use super::{
+    config::AlgorithmConfig,
+    matrix::{Matrix, Vector},
+    sqr, PeerSnapshot,
+};
+
+#[derive(Debug, Default, Copy, Clone)]
+struct RttBuf {
```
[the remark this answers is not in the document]
davidv1992: RoundTripTime, but yeah this probably needs a better name although I don't have any ideas right now.

ntp-proto/src/algorithm/kalman/peer.rs
```diff
@@ -0,0 +1,1067 @@
...
+    data: [f64; 8],
+    next_idx: usize,
+}
+
+impl RttBuf {
+    fn mean(&self) -> f64 {
+        self.data.iter().sum::<f64>() / (self.data.len() as f64)
+    }
+
+    fn var(&self) -> f64 {
```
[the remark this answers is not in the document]
davidv1992: fixed

ntp-proto/src/algorithm/kalman/peer.rs
```diff
@@ -0,0 +1,1067 @@
...
+/// delays r is used. Some straight-forward math shows that given s, the
+/// measurement noise matrix is (assuming independence between dp and d)
+/// s/4 s/4
+/// s/4 s/2
+///
+/// This setup leaves two major issues:
+///  - How often do we want measurements (what is the desired polling interval)
+///  - What is v
+///
+/// For the polling interval, this is changed dynamically such that
```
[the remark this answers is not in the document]
davidv1992: fixed

ntp-proto/src/algorithm/kalman/peer.rs
```diff
@@ -0,0 +1,1067 @@
...
+/// entirely sound, we construct the frequency measurement also using
+/// the previous measurement (which we will denote with tp and dp).
+///
+/// The observation is then the vector (tm, tm-tp), and the observation
+/// matrix is given by
+/// 1   0
+/// 0 tm-tp
+///
+/// To estimate the measurement noise, the variance s of the tranmission
+/// delays r is used. Some straight-forward math shows that given s, the
```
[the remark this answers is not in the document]
davidv1992: fixed

ntp-proto/src/algorithm/kalman/peer.rs
davidv1992: Don't really have a motivation, but added that this works well in practice.

ntp-proto/src/algorithm/kalman/peer.rs
davidv1992: fixed

ntp-proto/src/algorithm/kalman/peer.rs
davidv1992: fixed

ntp-proto/src/algorithm/kalman/peer.rs
davidv1992: fixed

ntp-proto/src/algorithm/kalman/peer.rs
```diff
@@ -0,0 +1,1067 @@
...
+/// The filter tracks the time difference between the local and remote
+/// timescales. For ease of implementation, it actually is programmed
+/// mostly as if the local timescale is absolute truth, and the remote
+/// timescale is the one that is estimated. The filter state is kept at
+/// a local timestamp t, and progressed in time as needed for processing
+/// measurements and producing estimation outputs.
+///
+/// The state is a vector (D, w) where
+///  - d is the offset between the remote and local timescales
+///  - w is (in seconds per second) the frequency difference.
```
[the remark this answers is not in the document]
davidv1992: fixed

ntp-proto/src/algorithm/kalman/peer.rs
```diff
@@ -0,0 +1,1067 @@
+/// This module implements a kalman filter to filter the measurements
+/// provided by the peers.
+///
+/// The filter tracks the time difference between the local and remote
+/// timescales. For ease of implementation, it actually is programmed
+/// mostly as if the local timescale is absolute truth, and the remote
+/// timescale is the one that is estimated. The filter state is kept at
+/// a local timestamp t, and progressed in time as needed for processing
+/// measurements and producing estimation outputs.
```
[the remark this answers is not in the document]
davidv1992: fixed

ntp-proto/src/algorithm/kalman/config.rs
```diff
@@ -0,0 +1,219 @@
...
+    /// measurements with a significantly larger network delay
+    /// are rejected. (standard deviations, 0+)
+    #[serde(default = "default_delay_outlier_threshold")]
+    pub delay_outlier_threshold: f64,
+
+    /// Initial estimate of the clock wander of the combination
+    /// of our local clock and that of the peer. (s/s^2)
+    #[serde(default = "default_initial_wander")]
+    pub initial_wander: f64,
+    /// Initail uncertainty of the frequency difference between
```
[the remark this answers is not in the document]
davidv1992: fixed

ntp-daemon/src/system.rs
```diff
@@ -276,19 +309,40 @@ impl<C: NtpClock> System<C> {
...
+        wait_enabled: &mut bool,
+    ) {
+        if let Some(ref used_peers) = update.used_peers {
+            self.system.update_used_peers(used_peers.iter().map(|v| {
+                self.peers.get(v).and_then(|data| data.snapshot).expect(
+                    "Critical error: Peer used for synchronization that is not known to system",
+                )
+            }));
+        }
+        if let Some(timesnapshot) = update.timesnapshot {
```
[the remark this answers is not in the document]
davidv1992: fixed

ntp-proto/src/time_types.rs
```diff
@@ -98,6 +98,10 @@ impl NtpTimestamp {
         NtpTimestamp::from_bits(timestamp.to_be_bytes())
     }
 
+    pub fn is_before(self, other: NtpTimestamp) -> bool {
+        self - other < NtpDuration::ZERO
```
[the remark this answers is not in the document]
davidv1992: Added also as comment, but in short, this is needed to deal with era wrapping. Around an era change, self can be near the maximum value for NtpTimestamp and other near the minimum, and that must be interpreted as self being before other (which it is due to wrapping in substraction of NtpTimestamp)

ntp-proto/src/algorithm/kalman/matrix.rs
```diff
@@ -0,0 +1,288 @@
...
+    fn add(self, rhs: Self) -> Self::Output {
+        Matrix {
+            data: self
+                .data
+                .iter()
+                .zip(rhs.data.iter())
+                .map(|(a, b)| a + b)
+                .collect::<Vec<_>>()
+                .try_into()
+                .unwrap(),
```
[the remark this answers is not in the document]
davidv1992: Fixed, just wrote it out (is also more clear in what it does)

ntp-proto/src/algorithm/kalman/peer.rs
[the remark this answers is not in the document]
folkertdev: if in doubt just write it out in full. That name does not come up often, so 1) it's not much typing effort and 2) it's very helpful in context

ntp-proto/src/algorithm/mod.rs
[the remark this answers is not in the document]
davidv1992: Sorry, but I am still missing the rationale for this. The standard library has a number of conventions specifically to do with backwards compatibility, and as far as I know this might just be one of those. Backwards compatibility isn't applicable that way in our context so I am a bit loathe to just take this convention because "the standard library does it that way."

ntp-proto/src/algorithm/kalman/peer.rs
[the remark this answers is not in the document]
davidv1992: Fixed, had inspiration and renamed the datastructure to AveragingBuffer, and wrote out the names of the fields.

ntp-proto/src/algorithm/kalman/matrix.rs
```diff
@@ -0,0 +1,288 @@
...
+            "Matrix((({:?},{:?}),({:?},{:?})))",
+            self.data[0], self.data[1], self.data[2], self.data[3]
+        ))
+    }
+}
+
+impl std::fmt::Display for Matrix {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        f.write_fmt(format_args!(
+            "{} {}\n{} {}",
```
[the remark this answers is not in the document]
davidv1992: Fixed

ntp-proto/src/system.rs
```diff
@@ -122,19 +109,33 @@ mod tests {
...
+            },
+        );
+
+        assert_eq!(system.time_snapshot, snapshot);
+
         assert_eq!(
-            system.time_snapshot.root_delay,
-            NtpDuration::from_seconds(1.0)
+            system.accumulated_steps_threshold,
+            Some(NtpDuration::from_seconds(1.0)),
```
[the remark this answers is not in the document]
davidv1992: Fixed

ntp-proto/src/system.rs
```diff
@@ -78,31 +76,20 @@ mod tests {
...
 
-        system.update(
-            std::iter::empty(),
-            TimeSnapshot {
-                root_delay: NtpDuration::from_seconds(1.0),
-                ..Default::default()
-            },
-            &SystemConfig::default(),
-        );
+        system.update_used_peers(std::iter::empty());
```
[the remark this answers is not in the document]
davidv1992: fixed with comment.

ntp-proto/src/algorithm/kalman/mod.rs
```diff
@@ -0,0 +1,358 @@
...
+        freq_update
+    }
+
+    fn update_desired_poll(&mut self) {
+        self.timedata.poll_interval = self
+            .peers
+            .values()
+            .map(|(state, _)| state.get_desired_poll(&self.config.poll_limits))
+            .min()
+            .unwrap_or(self.config.poll_limits.max);
```
folkertdev, line 275: why does max make sense here? We try to get the minimal poll from all the peers, so then `min` would intuitively make more sense to me.

ntp-proto/src/algorithm/kalman/mod.rs
```diff
@@ -0,0 +1,358 @@
...
+
+        KalmanClockController {
+            peers: HashMap::new(),
+            clock,
+            config,
+            algo_config,
+            freq_offset: 0.0,
+            desired_freq: 0.0,
+            timedata: TimeSnapshot::default(),
+            in_startup: false,
```
folkertdev, line 304: should this not be `true`? based on my searching, `in_startup` is never true currently.

ntp-proto/src/algorithm/kalman/select.rs
```diff
@@ -0,0 +1,230 @@
...
+            .iter()
+            .filter(|snapshot| {
+                let radius = snapshot.offset_uncertainty() * algo_config.range_statistical_weight
+                    + snapshot.delay * algo_config.range_delay_weight;
+                radius <= algo_config.max_peer_uncertainty
+                    && snapshot.offset() - radius <= maxt
+                    && snapshot.offset() + radius >= maxt
+                    && snapshot.leap_indicator.is_synchronized()
+            })
+            .cloned()
```
folkertdev, line 56: `copied()` more clearly communicates what actually happens here

ntp-proto/src/algorithm/kalman/select.rs
```diff
@@ -0,0 +1,230 @@
+use crate::SystemConfig;
+
+use super::{config::AlgorithmConfig, PeerSnapshot};
+
+enum BoundType {
+    Start,
+    End,
+}
+
+pub(super) fn select<Index: Copy>(
```
folkertdev, line 10: maybe here is not the place, but I remember that we discussed how the approach here is slightly different from the normal NTP algorithm. 

If I remember correctly, the approach here is to find the largest slice where most intervals overlap. The standard algorithm rather looks at midpoints of the intervals. 

- I think the approach here should be explained (at least a little)
- it would be useful to write down also how this is different from standard NTP (and what the tradeoffs are), but I'm not sure here is the right place for that.

ntp-proto/src/algorithm/kalman/select.rs
```diff
@@ -0,0 +1,230 @@
...
+            peer_uncertainty: NtpDuration::from_seconds(0.01),
+            peer_delay: NtpDuration::from_seconds(0.01),
+            leap_indicator: crate::NtpLeapIndicator::NoWarning,
+            last_update: NtpTimestamp::from_fixed_int(0),
+        }
+    }
+
+    #[test]
+    fn test_weighing() {
+        let candidates = vec![
```
folkertdev, line 90: these tests quickly become big blocks of code and magic constants. For quicker reading, it is helpful to quickly describe what each test is for conceptually. Without it, you have to visually scan for what numbers changed, and then really reason about what the expected outcome is.

ntp-proto/src/algorithm/standard/mod.rs
```diff
@@ -192,24 +196,29 @@ impl<C: NtpClock, PeerID: Hash + Eq + Copy + Debug> TimeSyncController<C, PeerID
...
+            return StateUpdate::default();
         }
 
         self.recalculate_clock(now)
     }
 
+    fn time_update(&mut self) -> StateUpdate<PeerID> {
+        // Not needed for standard algorithm
+        StateUpdate::default()
+    }
```
folkertdev, line 220: why not make it `unreachable!()` then?

ntp-proto/src/algorithm/kalman/mod.rs
davidv1992: max makes sense here exactly because we try to get the minimal poll from all the peers. Using max then guarantees that this will only ever get lower if there are more peers, not higher. (for the min operation, maximum is essentialy the identity element)

ntp-proto/src/algorithm/kalman/combiner.rs
```diff
@@ -0,0 +1,288 @@
...
+    } else {
+        None
+    }
+}
+
+pub(super) fn combine<Index: Copy>(
+    selection: &[PeerSnapshot<Index>],
+    algo_config: &AlgorithmConfig,
+) -> Option<Combine<Index>> {
+    if let Some(first) = selection.first() {
```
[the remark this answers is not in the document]
davidv1992: done

ntp-proto/src/algorithm/kalman/combiner.rs
```diff
@@ -0,0 +1,288 @@
...
+
+            used_peers.push((snapshot.index, peer_uncertainty.determinant()));
+
+            // Merge measurements
+            let mixer = (uncertainty + peer_uncertainty).inverse();
+            estimate = estimate + uncertainty * mixer * (peer_estimate - estimate);
+            uncertainty = uncertainty * mixer * peer_uncertainty;
+        }
+
+        used_peers.sort_by(|a, b| a.1.partial_cmp(&b.1).unwrap());
```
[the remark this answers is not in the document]
davidv1992: the total ordering is fine here, so decided to go with that.

ntp-proto/src/algorithm/kalman/peer.rs
```diff
@@ -0,0 +1,1077 @@
...
+    }
+}
+
+#[derive(Debug, Clone)]
+struct PeerFilter {
+    state: Vector,
+    uncertainty: Matrix,
+    clock_wander: f64,
+
+    rtt_stats: AveragingBuffer,
```
folkertdev, line 140: maybe rename to `roundtrip_times`, or else document what `rtt` stands for

ntp-proto/src/algorithm/kalman/peer.rs
```diff
@@ -0,0 +1,1077 @@
...
+        self.state = update * self.state;
+        self.uncertainty = update * self.uncertainty * update.transpose() + process_noise;
+        self.filter_time = time;
+
+        trace!(?time, "Filter progressed");
+    }
+
+    /// Absorb knowledge from a measurement
+    fn absorb_measurement(&mut self, measurement: Measurement) -> (f64, f64, f64) {
+        // Measurement paramaters
```
folkertdev, line 184: while we're here: paramaters > parameters

ntp-proto/src/algorithm/kalman/peer.rs
```diff
@@ -0,0 +1,1077 @@
...
+
+        // Statistics
+        let chi = difference.inner(difference_covariance.inverse() * difference);
+        let weight = measurement_noise.determinant() / difference_covariance.determinant();
+
+        self.last_measurement = measurement;
+
+        trace!(chi, weight = 1. - weight, "Measurement absorbed");
+
+        (chi, 1. - weight, m_delta_t)
```
folkertdev, line 219: the `weight = 1.0 - weight`  would benefit from a comment

ntp-proto/src/algorithm/kalman/mod.rs
```diff
@@ -0,0 +1,358 @@
...
+        }
+    }
+
+    fn steer_offset(&mut self, change: f64, freq_delta: f64) -> Option<NtpTimestamp> {
+        self.check_offset_steer(change);
+        if change.abs() > self.algo_config.jump_threshold {
+            // jump
+            self.clock
+                .step_clock(NtpDuration::from_seconds(change))
+                .unwrap();
```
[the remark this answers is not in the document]
davidv1992: Made it an expect

ntp-proto/src/algorithm/kalman/mod.rs
davidv1992: Good find, I'll fix it and add tests for this flag and its functioning.

ntp-proto/src/algorithm/kalman/peer.rs
```diff
@@ -0,0 +1,1077 @@
...
+        }))
+    }
+
+    pub fn update(
+        &mut self,
+        config: &SystemConfig,
+        algo_config: &AlgorithmConfig,
+        measurement: Measurement,
+        packet: NtpPacket<'static>,
+    ) -> bool {
```
folkertdev, line 387: unclear from name and type: does this perform the update, or signal that an update is needed?

ntp-proto/src/algorithm/kalman/mod.rs
```diff
@@ -0,0 +1,358 @@
...
+        &mut self,
+        id: PeerID,
+        measurement: Measurement,
+        packet: NtpPacket<'static>,
+    ) -> bool {
+        self.peers.get_mut(&id).map(|state| {
+            state
+                .0
+                .update(&self.config, &self.algo_config, measurement, packet)
+                & state.1
```
folkertdev, line 86: this is `&` on bool, so I think `&&` is clearer

ntp-proto/src/algorithm/kalman/mod.rs
davidv1992: Fixed

ntp-proto/src/algorithm/standard/mod.rs
davidv1992: There is no harm in it accidentally getting called, so crashing the entire program for that seems harsh.

ntp-proto/src/algorithm/kalman/peer.rs
davidv1992: Fixed

ntp-proto/src/algorithm/kalman/select.rs
davidv1992: No, the snapshots are not copy so this is actually a clone operation.

ntp-proto/src/algorithm/kalman/mod.rs
davidv1992: Fixed

ntp-proto/src/algorithm/kalman/peer.rs
davidv1992: Fixed

ntp-proto/src/algorithm/kalman/select.rs
davidv1992: Added explanations

ntp-proto/src/algorithm/kalman/select.rs
davidv1992: Added a short explanation that should cover most questions, and will make this a section in the white paper.

ntp-proto/src/algorithm/kalman/select.rs
folkertdev: missed the other definition of PeerSnapshot. By the looks of it the peersnapshot here could implement `Copy`, but it is pretty big so probably not worth it

ntp-proto/src/algorithm/standard/mod.rs
folkertdev: fair enough

folkertdev (approved):
I'm approving because we have reached a state where the code is in good shape, and we want to build on top of it.

PR #465, opened 2023-01-05
implement NTS key exchange state machine

folkertdev:
adds and tests the NTS KE (de)serialization and our state machine

### DIFF
--- a/ntp-proto/src/nts_record.rs
+++ b/ntp-proto/src/nts_record.rs
@@ -1,4 +1,7 @@
-use std::io::{Read, Write};
+use std::{
+    io::{Read, Write},
+    ops::ControlFlow,
+};
 
 #[derive(Debug)]
 pub enum WriteError {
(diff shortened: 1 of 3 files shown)


ntp-proto/src/nts_record.rs
```diff
@@ -366,3 +379,418 @@ impl NtsRecordDecoder {
...
+    pub cookies: Vec<Vec<u8>>,
+}
+
+pub type KeyExchangeStep = ControlFlow<Result<KeyExchange, KeyExchangeError>, KeyExchange>;
+
+pub enum NextStep {
+    NeedMoreInput(KeyExchange),
+    Done(KeyExchange),
+    ReadError(std::io::Error),
+    KeyExchangeError(KeyExchangeError),
```
davidv1992, line 407: I'd suggest merging these two error types. Externally, we mostly care about that an error happens, not which type precisely.

ntp-proto/src/nts_record.rs
```diff
@@ -366,3 +379,418 @@ impl NtsRecordDecoder {
...
+}
+
+#[derive(Debug, PartialEq, Eq)]
+pub struct KeyExchange {
+    pub remote: String,
+    pub port: u16,
+    pub cookies: Vec<Vec<u8>>,
+}
+
+pub type KeyExchangeStep = ControlFlow<Result<KeyExchange, KeyExchangeError>, KeyExchange>;
```
davidv1992, line 401: Nit: type doesn't need to be pub

ntp-proto/src/nts_record.rs
```diff
@@ -366,3 +379,418 @@ impl NtsRecordDecoder {
...
+
+pub enum NextStep {
+    NeedMoreInput(KeyExchange),
+    Done(KeyExchange),
+    ReadError(std::io::Error),
+    KeyExchangeError(KeyExchangeError),
+}
```
davidv1992, line 408: Looking at this whole enum, I think it would be better to make this a result type. Also, for the parser, we would provide a nicer interface I think if the final result (after Done) is the only place where data can be extracted from, as that would prevent accidentally using a partial exchange result.

davidv1992 (changes requested):
I think this is a good direction to move towards, but i think it can use some improvements, see below. Also, to me this feels like a sort of half interface, as it only provides the cookies, but not the key. I think the final public interface from ntp-proto should also handle the TLS and key extraction, including sending the request. I am open for doing that in a separate PR though if that is preferred.

The provided test set for decoding looks really good to me. I don't see any obviously missing cases.

davidv1992 (approved):

PR #473, opened 2023-01-10
Alternative setup for spawning

rnijveld:


rnijveld:
@davidv1992 You can take this up if you want, but otherwise I’ll finish it when I get back!

ntp-daemon/src/spawn/mod.rs
```diff
@@ -0,0 +1,240 @@
...
+
+/// Unique identifier for a spawner
+/// This is used to identify which spawner was used to create a peer
+#[derive(Copy, Clone, PartialEq, Eq, Debug)]
+pub struct SpawnerId(u64);
+
+impl SpawnerId {
+    pub fn new() -> SpawnerId {
+        static COUNTER: AtomicU64 = AtomicU64::new(1);
+        SpawnerId(COUNTER.fetch_add(1, std::sync::atomic::Ordering::Relaxed))
```
folkertdev, line 22: this is fancy, but does it not cause problems when testing?

ntp-daemon/src/spawn/mod.rs
```diff
@@ -0,0 +1,240 @@
...
+pub enum PeerRemovalReason {
+    Demobilized,
+    NetworkIssue,
+}
+
+impl SpawnEvent {
+    pub fn new(id: SpawnerId, action: SpawnAction) -> SpawnEvent {
+        SpawnEvent { id, action }
+    }
+}
```
folkertdev, line 100: move this up to the struct definition?

ntp-daemon/src/spawn/mod.rs
```diff
@@ -0,0 +1,240 @@
...
+        nts: Option<PeerNtsData>,
+    ) -> SpawnAction {
+        SpawnAction::Create(PeerCreateParameters {
+            id,
+            addr,
+            normalized_addr,
+            nts,
+        })
+    }
+}
```
folkertdev, line 124: this function seems kind of pointless?

ntp-daemon/src/spawn/mod.rs
```diff
@@ -0,0 +1,240 @@
...
+        &mut self,
+        _event: PeerCreateParameters,
+        _action_tx: &mpsc::Sender<SpawnEvent>,
+    ) -> Result<(), Self::Error> {
+        Ok(())
+    }
+
+    fn get_id(&self) -> SpawnerId;
+    fn get_addr_description(&self) -> String;
+}
```
folkertdev, line 201: trait functions should have some documentation on the expected contract and invariants

ntp.toml
```diff
@@ -5,7 +5,7 @@ log-filter = "info"
 log-format = "full"
 
 # Peers can be configured as a simple list (pool servers from ntppool.org)
-peers = ["0.pool.ntp.org", "1.pool.ntp.org", "2.pool.ntp.org", "3.pool.ntp.org"]
+peers = ["0.pool.ntp.org"]
```
folkertdev, line 8: is this change on purpose?

ntp-daemon/src/system.rs
```diff
@@ -212,6 +198,16 @@ impl<C: NtpClock, T: Wait> System<C, T> {
         )
     }
 
+    fn add_spawner(&mut self, spawner: impl Spawner + Send + Sync + 'static) -> SpawnerId {
+        let (notify_tx, notify_rx) = mpsc::channel(Self::MESSAGE_BUFFER_SIZE);
+        let id = spawner.get_id();
+        let spawner_data = SystemSpawnerData { id, notify_tx };
+        info!(id=?spawner_data.id, addr=spawner.get_addr_description(), "Running spawner");
```
folkertdev, line 205: would it be clear here what kind of spawner this is? You could probably get a type name with the functions from `std::any` if that is helpful

ntp-daemon/src/system.rs
```diff
@@ -377,137 +365,72 @@ impl<C: NtpClock, T: Wait> System<C, T> {
...
+        if let Some(spawner) = opt_spawner {
+            spawner
+                .notify_tx
+                .send(SystemEvent::peer_removed(
+                    peer_id,
+                    PeerRemovalReason::Demobilized,
+                ))
+                .await
+                .expect("Could not notify spawner");
+        }
```
folkertdev, line 385: has some duplication with the `handle_peer_network_issue`. is there an easy way to combine their logic?

folkertdev (commented):
Looks good on a technical level. Left a bunch of docs/clarification comments

ntp-daemon/src/spawn/standard.rs
```diff
@@ -0,0 +1,104 @@
...
+    ) -> Result<(), StandardSpawnError> {
+        self.spawn(action_tx).await
+    }
+
+    async fn handle_peer_removed(
+        &mut self,
+        _removed_peer: PeerRemovedEvent,
+        action_tx: &mpsc::Sender<SpawnEvent>,
+    ) -> Result<(), StandardSpawnError> {
+        self.spawn(action_tx).await
```
davidv1992, line 94: Problem: Respawn should only happen when not demobilized

ntp-daemon/src/spawn/mod.rs
rnijveld: Should not be a problem! This should generate unique ids in any random combination of threads, ordering does not matter as there is no reliance on the exact number being generated here, the only requirement is for it to be a unique number.

ntp-daemon/src/spawn/mod.rs
rnijveld: I noticed that having the enum and struct construction in the spawning algorithm made that little bit less readable than when this function was in place, especially given that creating the enum variant is more administration than anything algorithmically relevant. It would be my preference to keep it if you're ok with it.

ntp.toml
rnijveld: Accident, but while I'm at it, I've changed this to be a pool now instead of the individual peers

ntp-daemon/src/system.rs
rnijveld: I've added a get_description() function to the trait to help with this and will display that as well to make sure that we know what kind of spawner we are talking about in the logs

ntp-daemon/src/system.rs
```diff
@@ -377,137 +365,72 @@ impl<C: NtpClock, T: Wait> System<C, T> {
...
-            PeerState {
-                snapshot: None,
-                peer_address: PeerAddress::Peer { address: addr },
-            },
-        );
-        self.controller.peer_add(index);
+        // Try and find a related spawner and notify that spawner
+        if let Some(s) = self.spawners.iter().find(|s| s.id == spawner_id) {
+            let _ = s.notify_tx.send(SystemEvent::PeerRegistered(params)).await;
+        }
```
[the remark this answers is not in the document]
rnijveld: This only notifies back to the spawner that the peer is now registered with the system, mostly useful for testing purposes right now. But it can also be used as a feedback mechanism if we ever add a blocking mechanism in system (i.e. we only allow NTS peers in system and block peers that aren't NTS)

PR #480, opened 2023-01-13
Nts client spawning

folkertdev:


ntp-daemon/src/config/peer.rs
```diff
@@ -189,11 +233,31 @@ impl<'de> Deserialize<'de> for PeerConfig {
                             let raw: String = map.next_value()?;
 
                             let parsed_addr =
-                                NormalizedAddress::from_string(raw.as_str().to_string())
+                                NormalizedAddress::from_string_ntp(raw.as_str().to_string())
                                     .map_err(de::Error::custom)?;
 
                             addr = Some(parsed_addr);
                         }
+                        "ke_addr" => {
```
davidv1992, line 241: Do we want this to be a different name? For the end user there isn't really a need to differentiate here and it might be a bit confusing.

ntp-daemon/src/keyexchange.rs
```diff
@@ -1,13 +1,42 @@
...
+    path::Path,
     pin::Pin,
     task::{Context, Poll},
 };
 
 use ntp_proto::{KeyExchangeClient, KeyExchangeError, KeyExchangeResult};
+use rustls::Certificate;
 use tokio::io::{AsyncRead, AsyncWrite, ReadBuf};
 
+#[allow(unused)]
```
davidv1992, line 13: Shouldn't be needed right?

ntp-daemon/src/system.rs
```diff
@@ -272,7 +299,17 @@ impl<C: NtpClock> System<C> {
         self.peers.remove(&index);
     }
 
-    fn handle_spawn(&mut self, peer_address: PeerAddress, addr: SocketAddr) {
+    #[cfg(test)]
+    fn handle_spawn_no_nts(&mut self, peer_address: PeerAddress, addr: SocketAddr) {
+        self.handle_spawn(peer_address, addr, None)
+    }
```
davidv1992, line 305: Nit: Do we really need this?

ntp-daemon/src/system.rs
```diff
@@ -552,6 +624,48 @@ impl Spawner {
...
+                },
+                Err(e) => {
+                    warn!(error = ?e, "error while resolving peer address, retrying");
+                    tokio::time::sleep(NETWORK_WAIT_PERIOD).await
+                }
+            }
+        };
+
+        // we assume that when KE succeeds, it gives back a structurally valid remote and port.
+        let address = NormalizedAddress::new_unchecked(&ke.remote, ke.port);
```
davidv1992, line 660: Given that the remote can be provided by the server, blindly assuming here is maybe not the best option. What are the consequences if the remote part is not valid?

davidv1992 (commented):
Looks very strong, a few minor nitpicks and one issue that I want your opinion on before we merge, but other than that looks good to me.

ntp-daemon/src/system.rs
folkertdev: fully moved it into the `cfg(test)` module now

davidv1992 (approved):

PR #504, opened 2023-02-03
Introduced CipherProvider to allow dynamic choice of cipher based on package extension fields.

davidv1992:


ntp-proto/src/packet/mod.rs
```diff
@@ -270,9 +270,9 @@ impl<'a> NtpPacket<'a> {
         }
     }
 
-    pub fn deserialize(
+    pub fn deserialize<CP: CipherProvider>(
         data: &'a [u8],
-        cipher: Option<&Cipher>,
+        cipher: &CP,
```
folkertdev, line 275: a matter of style, but you could use `cipher: &impl CipherProvider` here

ntp-proto/src/packet/crypto.rs
```diff
@@ -0,0 +1,112 @@
...
+
+pub trait CipherProvider {
+    fn get(&self, context: &[ExtensionField<'_>]) -> Option<&dyn Cipher>;
+}
+
+impl CipherProvider for () {
+    fn get<'a>(&self, _context: &[ExtensionField<'_>]) -> Option<&dyn Cipher> {
+        None
+    }
+}
```
folkertdev, line 35: this looks weird at the use site. Can't we have something with a better name? e.g. `struct NoCipher;` I think would look much better at the use site

```diff
- let rec_packet = NtpPacket::deserialize(&buf, &()).unwrap();
+ let rec_packet = NtpPacket::deserialize(&buf, &NoCipher).unwrap();
```

ntp-proto/src/peer.rs
```diff
@@ -21,8 +20,8 @@ pub enum NtsError {
 
 pub struct PeerNtsData {
     pub(crate) cookies: CookieStash,
-    pub(crate) c2s: Aes128SivAead,
-    pub(crate) s2c: Aes128SivAead,
+    pub(crate) c2s: Box<dyn Cipher>,
+    pub(crate) s2c: Box<dyn Cipher>,
```
folkertdev, line 24: so, we use `Box<dyn Cipher>` here to support multiple ciphers, that possibly use different sizes for their keys. Additionally the keys are pretty big so storing them on the heap to keep `PeerNtsData` small is pretty reasonable.

maybe add a comment here (to me Box<dyn T> is a bit of an exotic feature)

ntp-proto/src/packet/crypto.rs
```diff
@@ -0,0 +1,112 @@
...
+        &self,
+        nonce: &[u8],
+        ciphertext: &[u8],
+        associated_data: &[u8],
+    ) -> Result<Vec<u8>, PacketParsingError>;
+}
+
+pub trait CipherProvider {
+    fn get(&self, context: &[ExtensionField<'_>]) -> Option<&dyn Cipher>;
+}
```
folkertdev, line 29: none of the current impls use the `context` argument. Your assumption is that certain ciphers will in the future?

davidv1992: This is for the server's CipherProvider, so that it can get access to the cookie it needs to get the key.

ntp-proto/src/packet/mod.rs
davidv1992: Done

ntp-proto/src/packet/crypto.rs
davidv1992: Done

ntp-proto/src/peer.rs
davidv1992: Done

folkertdev (approved):

PR #505, opened 2023-02-04
implement a generic unix NTP clock (for hardware timestamping, freebsd, macos)

folkertdev:
our existing implementation uses linux-specific functionality to reduce the number of syscalls. This is very adventageous when supported, but from what I can see, linux is the only OS that does. In particular, FreeBSD and Macos don't.

The solution implemented here is to fall back to a slightly less efficient approach using `ntp_gettime`. This is suggested by the NTP spec, and implemented in chrony as well.

The generic unix clock is not tested (yet) on freebsd or macos, but it does (cross)compile for those targets. I've tried the generic unix clock on linux, but cannot actually judge whether it is correct (some documentation on how to verify clocks would be nice!)


folkertdev:
@davidv1992 this PR has now also become relevant for hardware timestamping, because setting the time on the network card uses the same mechanism as e.g. macos for updating the clock

ntp-os-clock/src/unix.rs
```diff
@@ -0,0 +1,402 @@
...
+    }
+
+    unsafe fn from_path(path: &CStr) -> Result<Self, Error> {
+        let fd = match unsafe { libc::open(path.as_ptr(), libc::O_RDWR) } {
+            -1 => return Err(convert_errno()),
+            valid => valid,
+        };
+
+        Ok(unsafe { Self::from_file_descriptor(fd as RawFd) })
+    }
```
davidv1992, line 108: Why is this function unsafe to use, and what preconditions must hold to use it?

ntp-os-clock/src/unix.rs
```diff
@@ -0,0 +1,402 @@
...
+
+    unsafe fn from_file_descriptor(fd: RawFd) -> Self {
+        let id = ((!(fd as libc::clockid_t)) << 3) | 0b11;
+
+        Self::custom(id)
+    }
```
davidv1992, line 114: Why is this function unsafe to use, and what preconditions must hold to use it?

ntp-os-clock/src/unix.rs
```diff
@@ -0,0 +1,402 @@
...
+        Self { clock: id }
+    }
+
+    #[cfg_attr(target_os = "linux", allow(unused))]
+    pub fn ptp0() -> Result<Self, Error> {
+        let path = CStr::from_bytes_with_nul(b"/dev/ptp0\0").unwrap();
+        // SAFETY
+        //
+        // we know this is a file descriptor of a clock
+        unsafe { Self::from_path(path) }
```
davidv1992, line 98: How do we know this? a user could have just created a regular file under /dev/ptp0

ntp-os-clock/src/unix.rs
```diff
@@ -0,0 +1,402 @@
...
+        unsafe { cerr(libc::clock_settime(self.clock, &timespec))? };
+
+        Ok(())
+    }
+
+    fn clock_adjtime(&self, timex: &mut libc::timex) -> Result<(), Error> {
+        // We don't care about the time status, so the non-error
+        // information in the return value of ntp_adjtime can be ignored.
+        // The ntp_adjtime call is safe because the reference always
+        // points to a valid libc::timex.
```
davidv1992, line 142: This comment is out of date, plus are there requirements here on self.clock?

ntp-os-clock/src/unix.rs
```diff
@@ -0,0 +1,402 @@
...
+    }
+
+    #[cfg_attr(target_os = "linux", allow(unused))]
+    pub fn ptp0() -> Result<Self, Error> {
+        let path = CStr::from_bytes_with_nul(b"/dev/ptp0\0").unwrap();
+        // SAFETY
+        //
+        // we know this is a file descriptor of a clock
+        unsafe { Self::from_path(path) }
+    }
```
davidv1992, line 99: What is the purpose of this function? ptp0 is not guaranteed to always exist, nor to always be the clock we need when hardware timestamping (a user may have multiple network cards with ptp clocks). I think we should just let the user always specify the path to the relevant clock

davidv1992 (changes requested):
I have a few problems with this pr:
 - Why the two clock types, UnixNtpClock seems to implement everything the linux one does, or am I missing something?
 - The new unsafe code around file clocks needs documentation on its safety assumptions, these seem haphazard

ntp-os-clock/src/unix.rs
folkertdev: removed from this PR; will be part of the hardware timestamping PR

ntp-os-clock/src/unix.rs
folkertdev: updated. There are no requirements on `self.clock`; an invalid clock is handled gracefully with an EINVAL IO error

davidv1992:
Looks good, but have some worries about keeping track which clock is in use in combination with the cross platform functionality.

ntp-os-clock/src/unix.rs
```diff
@@ -0,0 +1,426 @@
...
+
+        Ok(current_time_timeval(_timex.time, precision))
+    }
+
+    #[cfg(not(target_os = "linux"))]
+    {
+        // clock_gettime always gives nanoseconds
+        let timespec = clock_gettime()?;
+        Ok(current_time_timespec(timespec, Precision::Nano))
+    }
```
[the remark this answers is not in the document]
folkertdev: this broke and I did not backport it from the macos/freebsd branch. Fixed now by taking `&self`

ntp-os-clock/src/unix.rs
```diff
@@ -0,0 +1,439 @@
...
+        // using an invalid clock id is safe. `clock_adjtime` will return an EINVAL error
+        // https://man.archlinux.org/man/clock_adjtime.2.en#EINVAL~4
+        #[cfg(target_os = "linux")]
+        use libc::clock_adjtime as adjtime;
+
+        #[cfg(any(target_os = "freebsd", target_os = "macos"))]
+        let adjtime = {
+            extern "C" {
+                fn clock_adjtime(clk_id: libc::clockid_t, buf: *mut libc::timex) -> libc::c_int;
+            }
```
folkertdev, line 144: based on https://www.gnu.org/software/gnulib/manual/html_node/clock_005fadjtime.html this function should exist on modern macos/freebsd. I'll see if we can get that upstreamed into libc

folkertdev: but this is not so relevant here (or for hardware timestamping)

davidv1992: This seems not to exist looking at the freebsd man pages, so we should be careful with this once we start actually testing that.

PR #508, opened 2023-02-06
Support more timestamp methods

folkertdev:
adds support for `SO_TIMESTAMP` and `SO_TIMESTAMPNS`. The former will be required to support macos and freebsd. The latter is included for completeness (and may be useful for older versions of linux, though we make no effort to support those at the moment).


ntp-udp/src/raw_socket.rs
```diff
@@ -27,62 +27,101 @@ pub(crate) fn cerr(t: libc::c_int) -> std::io::Result<libc::c_int> {
...
-        let mut options = 0;
-
-        if timestamping.rx_software || timestamping.tx_software {
-            // enable software timestamping
-            options |= libc::SOF_TIMESTAMPING_SOFTWARE
-        }
+        let options = match method {
+            TimestampMethod::SoTimestamp => timestamping.rx_software as u32,
+            #[cfg(target_os = "linux")]
+            TimestampMethod::SoTimestampns => timestamping.rx_software as u32,
```
davidv1992, line 87: These cases are somewhat unclear in what they do, maybe add a comment?

ntp-udp/src/raw_socket.rs
```diff
@@ -230,23 +294,32 @@ mod recv_message {
...
+                    ControlMessage::Timestamping(read_ntp_timestamp_timespec(timespec))
+                }
+
+                (libc::SOL_SOCKET, libc::SO_TIMESTAMP) => {
+                    // Safety:
+                    // current_msg was constructed from a pointer that pointed to a valid control message.
+                    // SO_TIMESTAMP always has a timeval in the data
+                    let cmsg_data = unsafe { libc::CMSG_DATA(current_msg) } as *const libc::timeval;
+                    let timeval = unsafe { std::ptr::read_unaligned(cmsg_data) };
+                    ControlMessage::Timestamping(read_ntp_timestamp_timeval(timeval))
```
davidv1992, line 322: Bit of a design nit, but this now makes this function do 2 things instead of 1, namely:
 - Provide a safe interface to libc
 - Handle conversion into ntp types
I would suggest we stick with just the first here to keep the unsafe code simpler and handle the conversion a layer up. This will likely also make it easier once we want to start sharing code with ptp

ntp-udp/src/raw_socket.rs
```diff
@@ -27,62 +27,101 @@ pub(crate) fn cerr(t: libc::c_int) -> std::io::Result<libc::c_int> {
...
 
+#[derive(Debug, Clone, Copy)]
+#[repr(i32)]
+#[allow(clippy::enum_variant_names)]
+pub(crate) enum TimestampMethod {
+    #[allow(dead_code)]
+    SoTimestamp = libc::SO_TIMESTAMP,
+    #[cfg(target_os = "linux")]
+    #[allow(dead_code)]
+    SoTimestampns = libc::SO_TIMESTAMPNS,
```
davidv1992, line 38: Should we even keep this around? There seems to be no case that uses this which risks this becoming unusable fairly quickly.

davidv1992 (commented):
Looks like a good solution direction to me. A few minor points included below.

ntp-udp/src/raw_socket.rs
folkertdev: ntpsec and chrony still have this so :man_shrugging: We do run a test with it, and it doesn't really cost anything to support it. My guess is that there are linux versions where `SO_TIMESTAMPNS` is available but `SO_TIMESTAMPING` is not. Or maybe `SO_TIMESTAMPNS` is better when you don't use send timestamps?

ntp-udp/src/raw_socket.rs
folkertdev: yeah. We have to deal with both `timespec` and `timeval` here, but I found a way to split it out of this function at least.

ntp-udp/src/raw_socket.rs
davidv1992: Disagree with this being resolved, the type conversions still seem magical to me.

ntp-udp/src/raw_socket.rs
```diff
@@ -211,8 +262,43 @@ mod recv_message {
...
+                    // truncates the higher bits of the i64
+                    let seconds = (timeval.tv_sec as u32).wrapping_add(EPOCH_OFFSET);
+
+                    let micros = timeval.tv_usec as u32;
+                    let nanos = micros * 1000;
+
+                    NtpTimestamp::from_seconds_nanos_since_ntp_era(seconds, nanos)
+                }
+            }
+        }
```
davidv1992, line 297: Design nit: I still feel like this shouldn't be part of the low-level potentially unsafe module, and would prefer to move it to socket.rs, just to make it very clear that nothing that happens here is relevant to the safety discussion.

ntp-udp/src/raw_socket.rs
davidv1992: Ok, let's keep it in

ntp-udp/src/raw_socket.rs
folkertdev: moved to the `lib.rs`

PR #509, opened 2023-02-06
add second encryption algorithm

folkertdev:
adds a second encryption algorithm, and the required infrastructure. I picked `AesSivCmac512` which seemed the most useful and straightforward to me.

I've verified using the `nts-ke` example that `AesSivCmac512` works (with `nts.time.nl`, cloudflare does not seem to support it)

### DIFF
--- a/ntp-proto/src/nts_record.rs
+++ b/ntp-proto/src/nts_record.rs
@@ -4,9 +4,13 @@ use std::{
     sync::Arc,
 };
 
-use aes_siv::Aes128SivAead;
+use aead::KeySizeUser;
+use aes_siv::{Aes128SivAead, Aes256SivAead};
 
-use crate::{cookiestash::CookieStash, packet::AesSivCmac256, peer::PeerNtsData};
+use crate::{
+    cookiestash::CookieStash, packet::AesSivCmac256, packet::AesSivCmac512, peer::PeerNtsData,
+    Cipher,
+};
 
 #[derive(Debug)]
 pub enum WriteError {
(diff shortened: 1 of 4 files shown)


ntp-proto/src/nts_record.rs
```diff
@@ -416,10 +421,37 @@ pub enum KeyExchangeError {
...
+        [0, 0, 0, self as u8, 0]
+    }
+
+    // per https://www.rfc-editor.org/rfc/rfc8915.html#section-5.1
+    pub const fn s2c_context(self) -> [u8; 5] {
+        // The final octet SHALL be 0x01 for the S2C key
+        [0, 0, 0, self as u8, 1]
+    }
+
+    const IN_ORDER_OF_PRECEDENCE: &'static [Self] =
```
davidv1992, line 446: Nit: I think you mean preference, not precedence.

ntp-proto/src/nts_record.rs
```diff
@@ -576,30 +640,53 @@ impl KeyExchangeClient {
...
+                                            Err(e) => return ControlFlow::Break(Err(e)),
+                                        };
+
+                                    PeerNtsData {
+                                        cookies: result.cookies,
+                                        c2s: Box::new(AesSivCmac512::new(c2s)),
+                                        s2c: Box::new(AesSivCmac512::new(s2c)),
+                                    }
+                                }
+                            };
```
davidv1992, line 684: This is getting rather bulky, I think we should at least refactor this into it's own function, and perhaps try to find a way to refactor more of the repetition out of this.

Maybe pushing this onto AeadAlgorithm is a good idea?

davidv1992 (commented):
Overal good initiative, few minor points below.

ntp-proto/src/nts_record.rs
folkertdev: yup

ntp-proto/src/nts_record.rs
folkertdev: yes, refactored this and it looks nicer now

PR #516, opened 2023-02-10
Return a parsed packet when decryption failed.

davidv1992:
This is needed for a server to be able to send nts-nak packets, since it then does need information contained unencrypted in the clients packet.

### DIFF
--- a/ntp-proto/src/packet/error.rs
+++ b/ntp-proto/src/packet/error.rs
@@ -12,27 +12,27 @@ pub enum ParsingError<T> {
 }
 
 impl<T> ParsingError<T> {
-    pub(super) fn coerce<U>(self) -> Option<ParsingError<U>> {
+    pub(super) fn get_decrypt_error<U>(self) -> Result<T, ParsingError<U>> {
         match self {
-            ParsingError::InvalidVersion(v) => Some(ParsingError::InvalidVersion(v)),
-            ParsingError::IncorrectLength => Some(ParsingError::IncorrectLength),
+            ParsingError::InvalidVersion(v) => Err(ParsingError::InvalidVersion(v)),
+            ParsingError::IncorrectLength => Err(ParsingError::IncorrectLength),
             ParsingError::MalformedNtsExtensionFields => {
-                Some(ParsingError::MalformedNtsExtensionFields)
+                Err(ParsingError::MalformedNtsExtensionFields)
             }
-            ParsingError::MalformedNonce => Some(ParsingError::MalformedNonce),
-            ParsingError::DecryptError(_) => None,
+            ParsingError::MalformedNonce => Err(ParsingError::MalformedNonce),
+            ParsingError::DecryptError(decrypt_error) => Ok(decrypt_error),
         }
     }
 }
 
 impl ParsingError<std::convert::Infallible> {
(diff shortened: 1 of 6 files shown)


ntp-proto/src/packet/error.rs
```diff
@@ -1,24 +1,54 @@
...
         match self {
             Self::InvalidVersion(version) => f.write_fmt(format_args!("Invalid version {version}")),
             Self::IncorrectLength => f.write_str("Incorrect packet length"),
             Self::MalformedNtsExtensionFields => f.write_str("Malformed nts extension fields"),
             Self::MalformedNonce => f.write_str("Malformed nonce (likely invalid length)"),
-            Self::DecryptError => f.write_str("Failed to decrypt NTS extension fields"),
+            Self::DecryptError(_) => f.write_str("Failed to decrypt NTS extension fields"),
         }
     }
 }
```
folkertdev, line 52: I want to suggest

```rust
impl<T> ParsingError<T> {
    pub(super) fn get_decrypt_error<U>(self) -> Result<T, ParsingError<U>> {
        match self {
            ParsingError::InvalidVersion(v) => Err(ParsingError::InvalidVersion(v)),
            ParsingError::IncorrectLength => Err(ParsingError::IncorrectLength),
            ParsingError::MalformedNtsExtensionFields => {
                Err(ParsingError::MalformedNtsExtensionFields)
            }
            ParsingError::MalformedNonce => Err(ParsingError::MalformedNonce),
            ParsingError::DecryptError(decrypt_error) => Ok(decrypt_error),
        }
    }
}

impl ParsingError<std::convert::Infallible> {
    pub(super) fn generalize<U>(self) -> ParsingError<U> {
        match self {
            ParsingError::InvalidVersion(v) => ParsingError::InvalidVersion(v),
            ParsingError::IncorrectLength => ParsingError::IncorrectLength,
            ParsingError::MalformedNtsExtensionFields => ParsingError::MalformedNtsExtensionFields,
            ParsingError::MalformedNonce => ParsingError::MalformedNonce,
            ParsingError::DecryptError(decrypt_error) => match decrypt_error {},
        }
    }
}
```

- this removes an `unwrap` below and would generally improve control flow in a few places
- uses exhaustive matching on `Infallibe` instead of an `unreachable!` (probably would compile away? not sure how rust deals with empty sum types)
- a function `forall U. Infallible -> U`  "generalizes the type". I think that name fits better than "force"

davidv1992: Implemented.

ntp-proto/src/packet/extensionfields.rs
```diff
@@ -391,41 +404,60 @@ impl<'a> ExtensionFieldData<'a> {
...
+                                .push(ExtensionField::InvalidNtsEncryptedField);
+                            has_invalid_nts = true;
+                            continue;
+                        }
+                    };
+
+                    let encrypted_fields =
+                        match encrypted.decrypt(cipher, &data[..header_size + offset]) {
+                            Ok(encrypted_fields) => encrypted_fields,
+                            Err(e) => match e.coerce() {
```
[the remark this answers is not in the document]
davidv1992: done

folkertdev (approved):

PR #520, opened 2023-02-10
remove sentry

folkertdev:


davidv1992:
Just figured out where we still use this, please hold off on merging.

folkertdev:
turns out we do want Sentry actually

PR #522, opened 2023-02-13
switch to workspace dependencies

folkertdev:


folkertdev:
codecov does not like this PR for some reason...

davidv1992 (approved):
Looks good to me, but given the status of the sentry changes we should consider doing this directly on master.

PR #536, opened 2023-02-23
NTS KE server

folkertdev:
functionally complete. Remaining issues are configuration and error handling.


ntp-proto/src/nts_record.rs
```diff
@@ -1027,6 +1315,31 @@ mod test {
...
+        let decode_output = || {
+            for chunk in bytes.chunks(24) {
+                decoder = match decoder.step_with_slice(chunk) {
+                    ControlFlow::Continue(d) => d,
+                    ControlFlow::Break(done) => return done,
+                };
+            }
+
+            Err(KeyExchangeError::IncompleteResponse)
+        };
```
folkertdev, line 1336: this can be made a bit nicer with labelled blocks. Needs a newer rust again though

ntp-daemon/src/keyexchange.rs
```diff
@@ -36,6 +44,48 @@ pub(crate) async fn key_exchange(
...
+        let fut = async move {
+            BoundKeyExchangeServer::run(stream, config, keyset)
+                .await
+                .map_err(|ke_error| std::io::Error::new(std::io::ErrorKind::Other, ke_error))
+        };
+
+        tokio::spawn(async move {
+            match fut.await {
+                Err(err) => tracing::error!(?err, ?peer_addr, "NTS KE failed"),
+                Ok(()) => tracing::info!(?peer_addr, "NTS KE completed"),
```
davidv1992, line 83: Given how often both of these will occur on a typical server, and the fact that errors are quite often not errors on our side (it could just be the client closing the connection earlier than we expect or something like that), I would suggest significantly lowering the importance of these messages (debug for Ok, and debug or info for the error case)

ntp-daemon/src/keyexchange.rs
```diff
@@ -36,6 +44,48 @@ pub(crate) async fn key_exchange(
...
+        let keyset = provider.get();
+
+        let fut = async move {
+            BoundKeyExchangeServer::run(stream, config, keyset)
+                .await
+                .map_err(|ke_error| std::io::Error::new(std::io::ErrorKind::Other, ke_error))
+        };
+
+        tokio::spawn(async move {
+            match fut.await {
```
davidv1992, line 81: Missing timeout here, which makes this server very vulnerable to DoS attacks. (just open a lot of connections and never close them)

ntp-proto/src/nts_record.rs
```diff
@@ -182,6 +182,44 @@ impl NtsRecord {
         ]
     }
 
+    pub fn server_key_exchange_records(
```
davidv1992, line 185: nit: does this need to be pub?

davidv1992 (commented):
Short preliminary review to speed up the process once this is fully done. Looks very promising already, just a few minor issues.

ntp-proto/src/nts_record.rs
folkertdev: nope, fixed

ntp-daemon/src/keyexchange.rs
folkertdev: good call

davidv1992:
Looks good to me, once we have the plumbing this should be fine to merge

Issue #518, opened 2023-02-10
Make sure that the configuration file has correct permissions and warn/error if it does not 


rnijveld:
We don't want configuration files that have 0o777 permissions

0xcrust:
Hey. I would like to take this one on. Is that fine with you?

davidv1992:
Feel free to propose a pull request. We do however have some license issues still to discuss internally so it might take a bit to merge it even if it is good (think a few months)

0xcrust:
Thanks for the heads up. I'd like to make the PR anyways. Some feedback would be appreciated

PR #539, opened 2023-02-24
Warn on unrestricted write permissions for config file

0xcrust:
Addresses #518

### DIFF
--- a/ntp-daemon/src/config/mod.rs
+++ b/ntp-daemon/src/config/mod.rs
@@ -13,6 +13,7 @@ use ntp_proto::{DefaultTimeSyncController, SystemConfig, TimeSyncController};
 use serde::{de, Deserialize, Deserializer};
 use std::{
     io::ErrorKind,
+    os::unix::fs::PermissionsExt,
     path::{Path, PathBuf},
     sync::Arc,
 };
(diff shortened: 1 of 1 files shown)


ntp-daemon/src/config/mod.rs
```diff
@@ -201,6 +202,13 @@ pub enum ConfigError {
 
 impl Config {
     async fn from_file(file: impl AsRef<Path>) -> Result<Config, ConfigError> {
+        let meta = std::fs::metadata(&file).unwrap();
+        let perm = meta.permissions();
+        
+        if perm.mode() & (libc::S_IRWXO | libc::S_IRWXG | libc::S_IRWXU) == 0o777 {
```
davidv1992, line 208: Sorry, the original issue wasn't very worked out on this, but thinking about it a bit more, I think we should primarily warn when the config is world writeable, and this can happen even if the permissions arent 0o777 (a somewhat stupid, but valid example of this on linux is a file with permissions 0117).

davidv1992 (commented):
Thank you for your contribution, the general approach looks good to me, I have one minor comment on the exact check done. I will also take this up with the rest of the team on getting us to a point where we can accept this licensing wise.

davidv1992:
closes #539

ntp-daemon/src/config/mod.rs
davidv1992: It is best to include this in this PR

0xcrust: Done

ntp-daemon/src/config/mod.rs
```diff
@@ -205,6 +206,13 @@ pub enum ConfigError {
 
 impl Config {
     async fn from_file(file: impl AsRef<Path>) -> Result<Config, ConfigError> {
+        let meta = std::fs::metadata(&file).unwrap();
+        let perm = meta.permissions();
+
+        if perm.mode() & libc::S_IWOTH == 0o002 {
```
davidv1992, line 212:
```suggestion
        if perm.mode() & libc::S_IWOTH != 0 {
```
Removes magic number

davidv1992 (approved):

PR #548, opened 2023-03-02
Added proper responses to nts messages.

davidv1992:


ntp-proto/src/packet/mod.rs
```diff
@@ -456,7 +456,87 @@ impl<'a> NtpPacket<'a> {
...
+                        .iter()
+                        .chain(input.efdata.encrypted.iter())
+                        .filter_map(|f| {
+                            if let ExtensionField::NtsCookiePlaceholder { cookie_length } = f {
+                                let new_cookie = keyset.encode_cookie(cookie);
+                                if new_cookie.len() > *cookie_length as usize {
+                                    None
+                                } else {
+                                    Some(ExtensionField::NtsCookie(Cow::Owned(new_cookie)))
+                                }
```
folkertdev, line 524: this logic can be factored out (and hence deduplicated)

ntp-proto/src/packet/mod.rs
```diff
@@ -456,7 +456,87 @@ impl<'a> NtpPacket<'a> {
...
+                    authenticated: input
+                        .efdata
+                        .authenticated
+                        .iter()
+                        .filter_map(|f| {
+                            if matches!(f, ExtensionField::UniqueIdentifier(_)) {
+                                Some(f.clone())
+                            } else {
+                                None
+                            }
```
folkertdev, line 509: can use `into_iter` here if the `authenticated` field is moved down

ntp-proto/src/packet/mod.rs
```diff
@@ -456,7 +456,87 @@ impl<'a> NtpPacket<'a> {
...
+                        .untrusted
+                        .iter()
+                        .chain(input.efdata.authenticated.iter())
+                        .filter_map(|f| {
+                            if matches!(f, ExtensionField::UniqueIdentifier(_)) {
+                                Some(f.clone())
+                            } else {
+                                None
+                            }
+                        })
```
folkertdev, line 474: this operation is done a bunch of times. Maybe factor this out into a function?

ntp-proto/src/packet/mod.rs
folkertdev: maybe even move this whole closure into its own function

ntp-proto/src/packet/mod.rs
davidv1992: done

ntp-proto/src/packet/mod.rs
davidv1992: done

ntp-proto/src/packet/mod.rs
davidv1992: done

folkertdev (approved):

PR #550, opened 2023-03-02
Allow the server to properly handle larger ntp messages.

davidv1992:


ntp-daemon/src/server.rs
```diff
@@ -69,10 +72,10 @@ pub struct ServerTask<C: 'static + NtpClock + Send> {
...
 enum AcceptResult<'a> {
-    Accept(NtpPacket<'a>, SocketAddr, NtpTimestamp),
+    Accept(NtpPacket<'a>, usize, SocketAddr, NtpTimestamp),
     Ignore,
-    Deny(NtpPacket<'a>, SocketAddr),
-    RateLimit(NtpPacket<'a>, SocketAddr),
+    Deny(NtpPacket<'a>, usize, SocketAddr),
+    RateLimit(NtpPacket<'a>, usize, SocketAddr),
     NetworkGone,
 }
```
folkertdev, line 80: I think we should move to struct variants here, so we can name the fields

folkertdev (approved):

PR #577, opened 2023-03-03
Bug fix classic algorithm

davidv1992:
Fixes bug caused by system receiving updates in a different order than they are timestamped (due to concurrency)


folkertdev (approved):
nice!

PR #580, opened 2023-03-03
add config files for a local nts client and server

### LINKED ISSUE #570: add example configurations (for NTS in particular)

folkertdev:
fix #570


davidv1992:
While I approve of the idea, the current naming scheme makes it insufficiently clear that this is an example/testing configuration only, that leaves the user vulnerable (because of known keys) if they use it anywhere.

PR #623, opened 2023-03-20
use Zeroize to zero out secrets on Drop

folkertdev:


ntp-proto/src/packet/crypto.rs
```diff
@@ -89,6 +90,18 @@ impl AesSivCmac256 {
     }
 }
 
+impl Zeroize for AesSivCmac256 {
+    fn zeroize(&mut self) {
+        self.key.zeroize()
+    }
+}
```
davidv1992, line 97: This allows you to (unintendedly) end up with a cipher object with a predictable (read, zero) key, which could pose its own issues. Do we want this?

folkertdev: unintentionally? you'd have to call `zeroize` which won't happen by accident I'd think?

in any case, I don't have a problem with it

PR #626, opened 2023-03-21
remove webpki as a dependency

folkertdev:
it is still used by rustls at the moment, but they will switch over to their fork (that is maintained and has some security/bug fixes). By upgrading rustls when it is released we should get this updated version of webpki automatically.


sanmai-NL:
Ah, I see, you mean the same thing but I interpreted your wording differently.

PR #635, opened 2023-03-27
Configure timestamps, clock and interface

folkertdev:


davidv1992 (changes requested):
Needs a bit more documentation, particularly also around status as mostly experimental, but other than that looks good.

folkertdev:
not sure what you mean by experimental? This stuff all works fine.

The docs are now already ready for hardware timestamping, they don't make much sense without it.

davidv1992 (approved):

PR #647, opened 2023-03-31
Some ratelimiting fixes and clarifications

davidv1992:


ntp-daemon/src/server.rs
```diff
@@ -483,13 +487,14 @@ impl<T: std::hash::Hash + Eq> TimestampedCache<T> {
...
             elements: std::iter::repeat_with(|| None).take(length).collect(),
+            randomstate: RandomState::new(),
         }
     }
 
     fn index(&self, item: &T) -> usize {
         use std::hash::Hasher;
 
-        let mut hasher = std::collections::hash_map::DefaultHasher::default();
+        let mut hasher = self.randomstate.build_hasher();
```
folkertdev, line 497: so this works because

> A particular instance RandomState will create the same instances of [Hasher](https://doc.rust-lang.org/std/hash/trait.Hasher.html), but the hashers created by two different RandomState instances are unlikely to produce the same result for the same values.

so on different runs we use a different random state, but within a run the same value hashes to the same index.

folkertdev (approved):

PR #649, opened 2023-03-31
Deal better with unreachable peers.

davidv1992:


ntp-daemon/src/spawn/standard.rs
```diff
@@ -93,6 +93,10 @@ impl BasicSpawner for StandardSpawner {
         removed_peer: PeerRemovedEvent,
         action_tx: &mpsc::Sender<SpawnEvent>,
     ) -> Result<(), StandardSpawnError> {
+        if removed_peer.reason == PeerRemovalReason::Unreachable {
+            // force new resolution
+            self.resolved = None;
+        }
         if removed_peer.reason != PeerRemovalReason::Demobilized {
             self.spawn(action_tx).await
         } else {
```
folkertdev, line 102: can this be an exhaustive pattern match?

PR #652, opened 2023-03-31
hardware timestamping

folkertdev:


ntp-daemon/src/config/mod.rs
```diff
@@ -110,23 +110,41 @@ where
...
     let data: Option<PathBuf> = Deserialize::deserialize(deserializer)?;
 
-    if let Some(_path_buf) = data {
-        Err(serde::de::Error::custom("not yet supported"))
+    if let Some(path) = data {
+        tracing::info!("using custom clock {path:?}");
+        DefaultNtpClock::from_path(&path).map_err(|e| serde::de::Error::custom(e.to_string()))
     } else {
-        Ok(DefaultNtpClock::default())
+        tracing::info!("using REALTIME clock");
```
davidv1992, line 117: Suggest making this a debug level message to not needlessly confuse regular users.

davidv1992 (commented):
Looks good, only one minor issue with a log message.

ntp-daemon/src/config/mod.rs
folkertdev: fixed

PR #658, opened 2023-04-04
Cleaned up example config to avoid confusing people

davidv1992:
Removed clock section as defaults should be fine and the actual content could cause significant numbers of warnings on some systems (notably my laptop). Also removed the nts config with the test ca because it is actively harmful.


folkertdev (approved):
makes sense, I'll add some custom config for these in the `config/` folder

PR #696, opened 2023-04-19
Provide musl libc compatibility

sanmai-NL:
Musl libc is a popular alternative to glibc. It was designed for [simplicity, correctness and efficiency](https://musl.libc.org/about.html). It has some slight [incompatibilities with glibc](https://wiki.musl-libc.org/compatibility.html).

Musl libc is [used by](https://wiki.musl-libc.org/projects-using-musl.html) e.g., Alpine Linux and Gentoo as overlay. Alpine Linux is a popular OS for container images.

This change also allows for compilation to statically linked binaries.

### DIFF
--- a/ntp-os-clock/src/unix.rs
+++ b/ntp-os-clock/src/unix.rs
@@ -15,7 +15,7 @@ use ntp_proto::{NtpClock, NtpDuration, NtpLeapIndicator, NtpTimestamp, PollInter
 
 // Libc has no good other way of obtaining this, so let's at least make our functions
 // more readable.
-#[cfg(target_os = "linux")]
+#[cfg(all(target_os = "linux", target_env = "gnu"))]
 pub(crate) const EMPTY_TIMEX: libc::timex = libc::timex {
     modes: 0,
     offset: 0,
@@ -51,6 +51,34 @@ pub(crate) const EMPTY_TIMEX: libc::timex = libc::timex {
     __unused9: 0,
     __unused10: 0,
     __unused11: 0,
+};
+
+#[cfg(all(target_os = "linux", target_env = "musl"))]
+pub(crate) const EMPTY_TIMEX: libc::timex = libc::timex {
+    modes: 0,
+    offset: 0,
+    freq: 0,
+    maxerror: 0,
+    esterror: 0,
+    status: 0,
+    constant: 0,
+    precision: 0,
+    tolerance: 0,
+    time: libc::timeval {
+        tv_sec: 0,
+        tv_usec: 0,
+    },
+    tick: 0,
+    ppsfreq: 0,
+    jitter: 0,
+    shift: 0,
+    stabil: 0,
+    jitcnt: 0,
+    calcnt: 0,
+    errcnt: 0,
+    stbcnt: 0,
+    tai: 0,
+    __padding: [0; 11]
 };
 
 #[cfg(any(target_os = "freebsd", target_os = "macos"))]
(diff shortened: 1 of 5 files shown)


folkertdev (approved):

ntp-udp/src/raw_socket.rs
```diff
@@ -190,15 +204,15 @@ mod recv_message {
...
-        };
+        let mut mhdr = empty_msghdr();
+
+        mhdr.msg_control = control_buf.as_mut_ptr().cast::<libc::c_void>();
+        mhdr.msg_controllen = control_buf.len() as _;
+        mhdr.msg_iov = (&mut buf_slice as *mut IoSliceMut).cast::<libc::iovec>();
+        mhdr.msg_iovlen = 1;
+        mhdr.msg_flags = 0;
+        mhdr.msg_name = (&mut addr as *mut libc::sockaddr_storage).cast::<libc::c_void>();
+        mhdr.msg_namelen = std::mem::size_of::<libc::sockaddr_storage>() as u32;
```
sanmai-NL, line 215: I wrote it in the more verbose style because the current style would enable undefined behavior. This creates unaligned references. See https://doc.rust-lang.org/stable/std/ptr/macro.addr_of_mut.html

See also https://lucumr.pocoo.org/2022/1/30/unsafe-rust/

sanmai-NL (commented):
@folkertdev Maybe you want to revert the merge pending a review on the safety aspect.

ntp-udp/src/raw_socket.rs
folkertdev: I think this is fine, because all of the fields are `repr(C)` and `0` is a valid value. Therefore we can safely use `MaybeUninit::zeroed().assume_init()`. Now we have a valid value of type `msghdr` on the stack, and setting fields on should be fine as well. I have verified this with miri. 

I believe that the example given does not work because `String` cannot be initialized with all zeros. The internal implementation uses this definition:

```rust
pub(crate) struct RawVec<T, A: Allocator = Global> {
    ptr: Unique<T>,
    cap: usize,
    alloc: A,
}
```

and a null pointer is not a valid value for `Unique<T>`, because

```
pub struct Unique<T: ?Sized> {
    pointer: NonNull<T>,
    // NOTE: this marker has no consequences for variance, but is necessary
    // for dropck to understand that we logically own a `T`.
    //
    // For details, see:
    // https://github.com/rust-lang/rfcs/blob/master/text/0769-sound-generic-drop.md#phantom-data
    _marker: PhantomData<T>,
}
```

ntp-os-clock/src/unix.rs
```diff
@@ -238,7 +243,15 @@ impl UnixNtpClock {
 
         let mut timespec = self.clock_gettime()?;
 
-        timespec.tv_sec += offset_secs as libc::time_t;
+        timespec.tv_sec += {
+            // this looks a little strange. it is to work around the `libc::time_t` type being
+            // deprectated for musl in rust's libc.
```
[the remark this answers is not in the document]
folkertdev: yeah I read the issue, and ... it's really unclear what they actually want to happen there. it must eventually be changed because that is just what musl expects now, right? I'll amend this to ignore the warning, and link to the issue

ntp-udp/src/raw_socket.rs
folkertdev: I copied the relevant code to a new project

PR #712, opened 2023-04-30
Avoid trying to compile Linux-only features on macOS

andrewaylett:
This is a blunt instrument, we merely conditionally compile out things that macOS doesn't support.  But it's enough to be able to build and run on macOS.

I've assumed a lax support policy, please feel free to commit to a stricter policy if you so desire :).

I also fixed up tests that use DNS to not hit a real resolver - macOS's resolver doesn't seem to randomise results on the timescale needed for tests, and it's cleaner to not rely on real nameservers if we don't have to.

### DIFF
--- a/CONFIGURATION.md
+++ b/CONFIGURATION.md
@@ -136,6 +136,7 @@ The ntp-daemon's primary configuration method is through a TOML configuration fi
 #### Peer configuration
 
 Peers are configured in the peers section, which should consist of a list of peers. Per peer, the following options are available:
+
 | Option | Default | Description |
 | --- | --- | --- |
 | mode | server | Type of peer connection to create. Can be any of `server`, `nts-server` or `pool` (for meaning of these, see below). |
@@ -188,6 +189,7 @@ max_peers = 4
 #### Server
 
 Interfaces on which to act as a server are configured in the `server` section. Per interface configured, the following options are available:
+
 | Option | Default | Description |
 | --- | --- | --- |
 | addr | | Address of the interface to bind to. |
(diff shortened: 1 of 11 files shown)


folkertdev:
Neat!

I've had ntpd-rs working on macOS a little while ago. We want to pick it up as part of FreeBSD support (macOS is very similar). You're right that we likely cannot really support macOS in the near future (e.g. CI for mac is tricky), but we'd like to at least make it compile, and given its similarity to FreeBSD, as long as we keep that working macOS should work as well.

We expect to continue real development in a couple of weeks, but I'll try to get to this sooner.

andrewaylett:
Nice -- no rush from my side, if I deploy to production it'll be on Linux, but I was sitting at a Mac when I found the project and figured it would be fun to try to get it working :).

valpackett:
I've gotten ntpd-rs working on FreeBSD some time ago, the one thing I got mildly stuck on was trying to implement `step_clock` using `clock_settime`, but that's not absolutely critical for the first run. I guess I'll rebase and submit what I have once this lands :)

folkertdev:
neat! I've combined my earlier work and bits of this PR into https://github.com/pendulum-project/ntpd-rs/pull/718. we can hopefully land that this week (some folks were on vacation this past week). On that PR, `cargo clippy` succeeds for the FreeBSD target, but I have no idea whether it actually works there. My attempts to run a VM have not worked out so far.

so (and maybe you did this already) can you check that the time is actually synchronized if you manually move the clock forward/backward by (say) a minute? be sure to only start `ntpd-rs` after you've changed the clock. changing the clock while ntpd-rs is running has some weird behavior (in part this is inherent, but I also think we can do better in this case). 

Some technical points: I wonder if this actually compiles? docs for it have been a bit spotty.

```
        #[cfg(target_os = "freebsd")]
        let adjtime = {
            extern "C" {
                fn clock_adjtime(clk_id: libc::clockid_t, buf: *mut libc::timex) -> libc::c_int;
            }
            clock_adjtime
        };
```

also FreeBSD can give us nanosecond precision in the timestamps (which macos does not support) so we'll want to set that up too.

valpackett:
> I wonder if this actually compiles?

No, there is no `clock_adjtime`, where did you even find that name? :) Ok, I guess moving to #718 then

andrewaylett:
Closing in favour of #718, thanks for the positive feedback :).

PR #716, opened 2023-05-01
use the newly added async_io function on AsyncFD

folkertdev:


ntp-udp/src/socket.rs
```diff
@@ -137,7 +137,26 @@ impl UdpSocket {
...
+        let send_size = match result {
+            Ok(size) => {
+                trace!(sent = size, "sent bytes");
+                size
+            }
+            Err(e) => {
+                debug!(error = debug(&e), "error sending data");
+                return Err(e);
+            }
+        };
```
folkertdev, line 156: general question: is this logging useful? it makes the code a lot noisier.

davidv1992 (approved):
Looks good, and is definetly an improvement. However, we should be aware that in combination with some uncertainty around how the epoll stuff should actually work this might break software/hardware send timestamping. Not sure whether that should change whether we merge this though, wat do you think @folkert?

PR #718, opened 2023-05-01
macOS compilation and testing

folkertdev:
**THIS DOES NOT MEAN MACOS IS NOW SUPPORTED!**

taking a bunch of my earlier work and changes from @andrewaylett 's PR #712 

I've tested this a bunch on macOS, and found one issue: the timestamps reported by the OS can have more than one second's worth of nanos. the timestamp looked fine otherwise, so we now correct this (add to seconds, subtract from nanos). Not doing so results in an assert.

@andrewaylett made some great suggestions for testing our spawner. This introduces (limited) non-determinism into the tests, but I'd argue that is actually what we want for a dns resolver.

an open question for the future (when we look at properly supporting FreeBSD) is the config. For instance, macos and probably freebsd do not allow steering of particular clocks with the APIs that we use. Tokio also does not allow setting of interfaces on those platforms. So some of our configuration just does not apply for these platforms.

### DIFF
--- a/.cargo/config.toml
+++ b/.cargo/config.toml
@@ -1,2 +1,13 @@
 [target.'cfg(all())']
 rustflags = ["-C", "force-unwind-tables"]
+
+# runs tests under sudo, useful for certain (hardware) clock manipulation in tests
+# [target.x86_64-unknown-linux-gnu]
+# runner = 'sudo -E'
+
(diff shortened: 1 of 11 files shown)


davidv1992 (approved):
Looks good to me, though for the long term I think we might need to consider separate or almost separate network/clock crates for the different platforms, this is getting quite heavy on the conditional compilation.

PR #724, opened 2023-05-07
FreeBSD fixes

### LINKED ISSUE #274: Add support for FreeBSD

valpackett:
It works, including the initial jump :)

Yeah `SCM_REALTIME` I'll need to submit to libc…

Fixes #274

### DIFF
--- a/ntp-udp/src/raw_socket.rs
+++ b/ntp-udp/src/raw_socket.rs
@@ -39,7 +39,7 @@ pub(crate) enum TimestampMethod {
     SoTimestamping = libc::SO_TIMESTAMPING,
     /// Original timestamping for unix (linux, freebsd, macos)
     ///
-    /// - microsecond precision
+    /// - microsecond precision (configurable on freebsd, we set nanoseconds)
     /// - only receive timestamps
     #[allow(dead_code)]
     SoTimestamp = libc::SO_TIMESTAMP,
@@ -59,10 +59,16 @@ mod set_timestamping_options {
 
     use super::{cerr, TimestampMethod};
 
+    enum SockOpt {
+        Method(TimestampMethod),
+        #[cfg(target_os = "freebsd")]
+        Clock,
+    }
+
     fn configure_timestamping_socket(
         udp_socket: &std::net::UdpSocket,
-        method: TimestampMethod,
-        options: u32,
+        option: SockOpt,
+        value: u32,
     ) -> std::io::Result<libc::c_int> {
         // Documentation on the timestamping calls:
         //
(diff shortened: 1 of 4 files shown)


ntp-udp/src/raw_socket.rs
```diff
@@ -61,7 +61,7 @@ mod set_timestamping_options {
 
     fn configure_timestamping_socket(
         udp_socket: &std::net::UdpSocket,
-        method: TimestampMethod,
```
folkertdev, line 64: why remove the type safety here? this was meant as a safeguard to just passing in integers. An extra method can be added to the enum if that would help.

folkertdev (commented):
this is great!

- I left 1 inline comment/question
- yes it would be nice to have that constant in libc. I've found that process to be straightforward.
- I'll try to setup CI here. this is a bit tricky because we need the appropriate linker (because `ring` has a `build.rs`), and this failed earlier for freebsd. but maybe it works now that the code works.
- we'll likely want to merge the macos stuff first, and then merge this into main. this may take a little while

ntp-udp/src/raw_socket.rs
valpackett: Well, because it's not *a timestamp method* (but rather an extra configuration sockopt for an existing one) and it shouldn't be exposed outside of this module… I guess the better option would be like:

```rust
enum SocketOption {
    Method(TimestampMethod),
    #[cfg(target_os = "freebsd")]
    TimestampClock,
}
```

folkertdev:
eh, I'm not sure what happened here? this should have rebased onto `main`. weird

folkertdev:
@valpackett do you have thoughts on how to have users install ntpd-rs on freebsd? We provide deb and rpm installers which put the binary and config in the right place, and configure some permissions and user groups. 

The installers are built with [ploutos](https://github.com/NLnetLabs/ploutos), which does not support freebsd currently. I'm really not sure how easy that would be to add (though we know the NLNet labs folks well, so we can check with them)

In the meantime we could maybe have some instructions that a user would have to perform manually?

valpackett:
It's uncommon for projects to build their own .pkg independently and for anyone to want to install such packages :) Generally someone would submit a port and the official package will exist. It's no big deal, there is no barrier to entry, and for such a high profile project you can be sure someone will show up to do it soon. (I'll drop a link to here in some places in fact.. :D)

ploutos is really just using [cargo-deb](https://crates.io/crates/cargo-deb) and [cargo-generate-rpm](https://crates.io/crates/cargo-generate-rpm), I'm pretty sure nothing like that exists for FreeBSD pkg for the above reason.

ntp-udp/src/raw_socket.rs
```diff
@@ -157,7 +167,10 @@ mod set_timestamping_options {
             }
         };
 
-        configure_timestamping_socket(udp_socket, method, options)?;
+        configure_timestamping_socket(udp_socket, SockOpt::Method(method), options)?;
+
+        #[cfg(target_os = "freebsd")]
+        configure_timestamping_socket(udp_socket, SockOpt::Clock, libc::SO_TS_REALTIME as u32)?;
```
davidv1992, line 173: This results in two calls to configure_timestamping_socket on freebsd, is that desirable?

davidv1992 (commented):
Looks good, one question on a weird second call.

ntp-udp/src/raw_socket.rs
davidv1992: Never mind, I see what this is for

folkertdev (approved):
ok, great. thanks so much for your help here @valpackett !

PR #736, opened 2023-05-11
proper waiter for timestamp messages in the error queue

folkertdev:


ntp-udp/src/raw_socket.rs
```diff
@@ -486,42 +484,61 @@ pub(crate) mod timestamping_config {
...
+        }
+
+        pub async fn wait(&self) -> std::io::Result<()> {
+            self.epoll_fd
+                .async_io(Interest::READABLE, |fd| {
+                    let mut ev = libc::epoll_event { events: 0, u64: 0 };
+
+                    match unsafe { libc::epoll_wait(*fd, &mut ev as *mut _, 1, 0) } {
+                        0 => Err(std::io::ErrorKind::WouldBlock.into()),
+                        _ => Ok(()),
```
folkertdev, line 538: what does `0` actually mean here? the docs for `async_io` note

> The closure should only return a [WouldBlock](https://doc.rust-lang.org/nightly/std/io/error/enum.ErrorKind.html#variant.WouldBlock) error if it has performed an IO operation on the file descriptor that failed due to the file descriptor not being ready. Returning a [WouldBlock](https://doc.rust-lang.org/nightly/std/io/error/enum.ErrorKind.html#variant.WouldBlock) error in any other situation will incorrectly clear the readiness flag, which can cause the file descriptor to behave incorrectly.

also

>  while
       specifying a timeout equal to zero cause epoll_wait() to return
       immediately, even if no events are available.

this seems to be the case here, so what does this actually do? isn't it just a busy loop at now?

davidv1992: epoll_wait returns the number of events read. If that is zero, then there weren't any available to be read, hence that is a wouldblock condition.

davidv1992: It would block, but for the timeout of 0

folkertdev: ok, that makes sense

PR #745, opened 2023-05-16
use rcgen to generate the certificates for NTS in a test

folkertdev:


davidv1992 (commented):
The test itself looks good and reasonable, but why not just use the certificates in the test folders? That would save the complexity of generating them.

folkertdev:
well, the idea is that this actually shows how to generate the certificates from scratch, instead of using a file for which we may not be able to retrieve how it was produced. 

Dirkjan suggested that I try this to show that it is actually relatively easy to use a certificate authority to generate some certificates. (the code looks small, but it took quite a while to figure out some minor issues). 

this setup also gives some freedom in testing more input formats.

davidv1992:
I disagree with this being a good idea. This is all generating the certificates in code, which is not something end users are ever going to use. This adds significant complexity over using pregenerated certificates, with very little gain. Also, generating new certificates is a quite expensive operation (or at least, it should be if done well) I would rather not do in our tests.

If we ever are going to document generating certificates this way I would much rather do it in an example or something like that, rather than in a test case

rnijveld:
I think that we can solve the 'having an example' part by explaining certificate generation in the docs, and I would prefer tests to be as deterministic as possible, to prevent any flaky tests (although probably unlikely in this case). As for testing multiple input formats, I think we can just have multiple testing certificates for that purpose. I'll close this PR for now, but please re-open if we think we should do something else!

PR #760, opened 2023-05-28
Refactor to keep peer packets in peers.

davidv1992:
The advantage of this is that it reduces the amount of copying of packets that we do in the client, which is unwanted because:
 - it is somewhat expensive (with nts there are multiple layers of allocations that happen when making packet copies)
 - it means some somewhat sensitive values (mainly cookies) end up duplicated in multiple places where they aren't needed.

### DIFF
--- a/ntp-proto/src/peer.rs
+++ b/ntp-proto/src/peer.rs
@@ -4,7 +4,8 @@ use crate::{
     cookiestash::CookieStash,
     packet::{Cipher, NtpAssociationMode, RequestIdentifier},
     time_types::NtpInstant,
-    NtpDuration, NtpPacket, NtpTimestamp, PollInterval, ReferenceId, SystemConfig, SystemSnapshot,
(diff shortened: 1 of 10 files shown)


ntp-proto/src/peer.rs
```diff
@@ -264,7 +281,7 @@ pub enum AcceptSynchronizationError {
 #[allow(clippy::large_enum_variant)]
 pub enum Update {
     BareUpdate(PeerSnapshot),
-    NewMeasurement(PeerSnapshot, Measurement, NtpPacket<'static>),
+    NewMeasurement(PeerSnapshot, Measurement),
```
folkertdev, line 284: maybe this clippy allow can now be removed?

davidv1992: Good catch, also removed the clippy allow on `MsgForSystem`

folkertdev (approved):

PR #763, opened 2023-05-29
Detect outside meddling with the system clock and deal with it.

### LINKED ISSUE #717: ntpd-rs gets stuck when clock is adjusted while it is running
when the clock is moved (forward, did not test backward) while ntpd-rs is running, it will get stuck, printing 

```
2023-05-01T14:55:02.819620Z  INFO ntp_proto::algorithm::kalman: No concensus cluster found
2023-05-01T14:55:17.925342Z  INFO ntp_proto::algorithm::kalman: No concensus cluster found
2023-05-01T14:55:18.342352Z  INFO ntp_proto::algorithm::kalman: No concensus cluster found
2023-05-01T14:55:19.160934Z  INFO ntp_proto::algorithm::kalman: No concensus cluster found
2023-05-01T14:55:19.242889Z  INFO ntp_proto::algorithm::kalman: No concensus cluster found
2023-05-01T14:55:34.270160Z  INFO ntp_proto::algorithm::kalman: No concensus cluster found
2023-05-01T14:55:34.654405Z  INFO ntp_proto::algorithm::kalman: No concensus cluster found
2023-05-01T14:55:35.546725Z  INFO ntp_proto::algorithm::kalman: No concensus cluster found
2023-05-01T14:55:35.972728Z  INFO ntp_proto::algorithm::kalman: No concensus cluster found
2023-05-01T14:55:50.608621Z  INFO ntp_proto::algorithm::kalman: No concensus cluster found
2023-05-01T14:55:51.119023Z  INFO ntp_proto::algorithm::kalman: No concensus cluster found
```

such an adjustment of the clock should not happen of course, but maybe we can do a better job of recognizing it?

davidv1992:
This patch ensures that any external change to the system clock resets the kalman algorithm, ensuring we at least keep operating when that happens even though it is less than ideal. Depends on #760 

Fixes #717

### DIFF
--- a/ntp-proto/src/algorithm/kalman/peer.rs
+++ b/ntp-proto/src/algorithm/kalman/peer.rs
@@ -72,7 +72,7 @@
 /// If they are often too small, v is quartered, and if they are often too
 /// large, v is quadrupled (note, this corresponds with doubling/halving
 /// the more intuitive standard deviation).
-use tracing::{debug, info, trace};
+use tracing::{debug, error, info, trace};
 
 use crate::{
     Measurement, NtpDuration, NtpTimestamp, PollInterval, PollIntervalLimits, SystemConfig,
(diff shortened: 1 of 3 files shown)


ntp-proto/src/algorithm/kalman/peer.rs
```diff
@@ -421,7 +421,29 @@ impl PeerState {
...
+                // We check that the difference between the localtime and monotonic
+                // times of the measurement is in line with what would be expected
+                // from recent steering. This check needs to be done here since we
+                // need to revert back to the initial state.
+                if ((measurement.localtime - filter.last_measurement.localtime)
+                    - measurement
+                        .monotime
+                        .abs_diff(filter.last_measurement.monotime))
+                .abs()
+                    > algo_config.meddling_threshold
```
folkertdev, line 434: can this be broken up? e.g. give some good names to the intermediate results. This looks really weird and is hard to read

ntp-proto/src/algorithm/kalman/peer.rs
```diff
@@ -569,7 +746,7 @@ mod tests {
                 transmit_timestamp: Default::default(),
                 receive_timestamp: Default::default(),
                 localtime: base + NtpDuration::from_seconds(1000.0),
-                monotime: basei,
+                monotime: basei + std::time::Duration::from_secs(1000),
```
folkertdev, line 749: why this change?

folkertdev (changes requested):
I think something should be added to the changelog too for this change, because it adds a new config option.

folkertdev:
also does this fix https://github.com/pendulum-project/ntpd-rs/issues/717 ?

davidv1992:
Yes it does

ntp-proto/src/algorithm/kalman/peer.rs
davidv1992: done.

ntp-proto/src/algorithm/kalman/peer.rs
davidv1992: Because we now check that both clocks progress at the same rate, otherwise the update is rejected.

folkertdev (approved):

PR #765, opened 2023-05-30
Clearer panic logic

folkertdev:
adds comments, and cleans up logic around slicing. 

Locally it is often difficult to verify that a certain slice operation is correct. By passing around slices instead of indices, the correctness, or at least the property that code will not panic, is more obvious.

### DIFF
--- a/ntp-proto/src/packet/extensionfields.rs
+++ b/ntp-proto/src/packet/extensionfields.rs
@@ -485,28 +485,23 @@ impl<'a> RawEncryptedField<'a> {
     ) -> Result<Self, ParsingError<std::convert::Infallible>> {
         use ParsingError::*;
 
-        if message_bytes.len() < 4 {
+        let [ b0, b1, b2, b3, ref rest @ .. ] = message_bytes[..] else {
             return Err(IncorrectLength);
-        }
+        };
 
-        let value = message_bytes;
-
-        if message_bytes.len() < 4 {
-            return Err(IncorrectLength);
-        }
-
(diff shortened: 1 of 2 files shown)


ntp-proto/src/packet/extensionfields.rs
```diff
@@ -627,7 +636,14 @@ impl<'a> Iterator for ExtensionFieldStreamer<'a> {
...
 
-const fn next_multiple_of(lhs: u16, rhs: u16) -> u16 {
+const fn next_multiple_of_u16(lhs: u16, rhs: u16) -> u16 {
+    match lhs % rhs {
+        0 => lhs,
+        r => lhs + (rhs - r),
+    }
+}
+
+const fn next_multiple_of_usize(lhs: usize, rhs: usize) -> usize {
```
davidv1992, line 646: I suggest investing some time to make this generic. If need be I can take that up (I remember having done that before but can't find where anymore).

ntp-proto/src/packet/extensionfields.rs
```diff
@@ -461,16 +462,17 @@ impl<'a> ExtensionFieldData<'a> {
...
-            Ok((this, size + header_size, cookie))
-        }
+
+        let result = DeserializedExtensionField {
+            efdata: this,
+            remaining_bytes: &extension_field_bytes[size..],
+            opt_cookie: is_valid_nts.then_some(cookie),
+        };
+
+        Ok(result)
```
davidv1992, line 475: This reduces an error condition to a success return, which is not ok in my view.

davidv1992 (changes requested):

davidv1992 (approved):

PR #766, opened 2023-05-30
deploy FreeBSD server

folkertdev:
We want to add a freebsd server running ntpd-rs to the ntp pool. I'm in the process of setting up a google compute cloud server for this purpose.

1. install cargo via rustup, and install git
1. clone the ntpd-rs repo
1. `cargo build --release --features "sentry"`. the sentry feature flag is only needed when you plan on using sentry
1. `sudo mv target/release/ntp-daemon /usr/local/bin/ntp-daemon`
1. ensure `/var/run/ntpd-rs` exists (the linux equivalent is `/run/ntpd-rs`, on freebsd this is nested inside `/var`)
1. put an ntp config at `/etc/ntpd-rs/ntp.toml`
1. be sure that the default `ntpd` deamon is stopped: `sudo service ntpd stop` and even `sudo sysrc ntpd_enable=NO` to prevent it from being started at boot 

At this point, `sudo /usr/local/bin/ntp-daemon` should work and (after a little while) synchronize the time. But we want to run this as a daemon. FreeBSD does not use `systemd`, but instead a thing called `rc.d`.

So I put this file at `/etc/rc.d/ntp-daemon`:

```sh
#!/bin/sh 

. /etc/rc.subr 

# REQUIRE: DAEMON FILESYSTEMS devfs
# BEFORE:  LOGIN
# KEYWORD: nojail resume shutdown

name="ntp_daemon" 
rcvar="ntp_daemon_enable"


# start_cmd="${name}_start" 
stop_cmd=":" 
start_precmd="${name}_precmd"

command="/usr/local/bin/ntp-daemon"
pidfile="/var/run/ntpd-rs.pid"
required_files="$command"

: ${ntp_daemon_enable:="NO"} 
# : ${ntp_daemon_user:="ntpd-rs"}
# : ${ntp_daemon_group:="ntpd-rs"}


ntp_daemon_precmd()
{
    export RUST_LOG="info"
    if [ ! -d /var/run/ntpd-rs ]; then
        install -d -o $ntp_daemon_user -g $ntp_daemon_group -m 0755 /var/run/ntpd-rs
    fi
}

load_rc_config $name 
run_rc_command "$1"
```

We want it to start at boot

```
sudo sysrc ntp_daemon_enable=NO
```

Then the daemon is started using 

```
rc_debug=true sudo -E service ntp-daemon start
```

## Users

The script above has some commented lines regarding users. I used the following commands

```
sudo pw groupadd ntpd-rs
sudo pw useradd ntpd-rs -g ntpd-rs -d /nonexistent -s /usr/sbin/nologin
```

to create an `ntpd-rs` group and and user (and the user is in the group). However, this user is not allowed to touch the clock.

For the linux systems, we use `AmbientCapabilities=CAP_SYS_TIME` in the systemd service file to configure the capabilities. There does not seem to be something equivalent for `rc.d`, and other approaches also don't seem to work on the VM that I have access to.

### DIFF
--- a/ntpd-rs.rc.d.example
+++ b/ntpd-rs.rc.d.example
@@ -1,35 +1,23 @@
-#!/bin/sh 
+#!/bin/sh
 
-. /etc/rc.subr 
+. /etc/rc.subr
 
+# PROVIDE: ntp_daemon
 # REQUIRE: DAEMON FILESYSTEMS devfs
 # BEFORE:  LOGIN
 # KEYWORD: nojail resume shutdown
 
-name="ntp_daemon" 
+name="ntp_daemon"
 rcvar="ntp_daemon_enable"
 
+: ${ntp_daemon_enable:="NO"}
 
-# start_cmd="${name}_start" 
-stop_cmd=":" 
 start_precmd="${name}_precmd"
 
-command="/usr/local/bin/ntp-daemon"
-pidfile="/var/run/ntpd-rs.pid"
-required_files="$command"
+command="/usr/sbin/daemon"
+procname="daemon"
+pidfile="/var/run/${name}.pid"
+command_args=" -f -o /var/log/ntp_daemon.log -H -P ${pidfile} /usr/local/bin/ntp-daemon --"
 
-: ${ntp_daemon_enable:="NO"} 
-# : ${ntp_daemon_user:="ntpd-rs"}
-# : ${ntp_daemon_group:="ntpd-rs"}
-
-
-ntp_daemon_precmd()
-{
-    export RUST_LOG="info"
-    if [ ! -d /var/run/ntpd-rs ]; then
-        install -d -o $ntp_daemon_user -g $ntp_daemon_group -m 0755 /var/run/ntpd-rs
-    fi
-}
-
-load_rc_config $name 
+load_rc_config $name
 run_rc_command "$1"
--- a/ntpd-rs.rc.d.example
+++ b/ntpd-rs.rc.d.example
@@ -15,7 +15,7 @@ rcvar="ntp_daemon_enable"
 start_precmd="${name}_precmd"
 
 command="/usr/sbin/daemon"
-procname="daemon"
+procname="ntp-daemon"
 pidfile="/var/run/${name}.pid"
 command_args=" -f -o /var/log/ntp_daemon.log -H -P ${pidfile} /usr/local/bin/ntp-daemon --"
 
--- a/ntpd-rs.rc.d.example
+++ b/ntpd-rs.rc.d.example
@@ -1,0 +1,35 @@
+#!/bin/sh 
+
+. /etc/rc.subr 
+
+# REQUIRE: DAEMON FILESYSTEMS devfs
+# BEFORE:  LOGIN
+# KEYWORD: nojail resume shutdown
+
+name="ntp_daemon" 
+rcvar="ntp_daemon_enable"
+
+
+# start_cmd="${name}_start" 
+stop_cmd=":" 
+start_precmd="${name}_precmd"
+
+command="/usr/local/bin/ntp-daemon"
+pidfile="/var/run/ntpd-rs.pid"
+required_files="$command"
+
+: ${ntp_daemon_enable:="NO"} 
+# : ${ntp_daemon_user:="ntpd-rs"}
+# : ${ntp_daemon_group:="ntpd-rs"}
+
+
+ntp_daemon_precmd()
+{
+    export RUST_LOG="info"
+    if [ ! -d /var/run/ntpd-rs ]; then
+        install -d -o $ntp_daemon_user -g $ntp_daemon_group -m 0755 /var/run/ntpd-rs
+    fi
+}
+
+load_rc_config $name 
+run_rc_command "$1"


folkertdev:
@valpackett do you have ideas here? always running as root is not great from a security perspective.

valpackett:
> There does not seem to be something equivalent for rc.d

There is no equivalent for Linux capabilities *in the FreeBSD kernel*, currently there's no API that allows you to grant "the ability to do just this one thing that root can do" to a program. (Hm, adding this to my future-kernel-proposals-to-prototype list!…)

@folkertdev the traditional way to do this used by the existing NTP daemons is just process-based privsep: you keep one process running as root that *only* does the adjtime/clock_settime/etc. calls based on commands piped in from the other process that does drop privileges and interacts with the network.

folkertdev:
That's good to know (it's hard to find definitive answers to such questions on the internet).

Running two processes does not work very well with our current setup, but we've decided that just running as root is fine (for now), at least for this server we want to add to the NTP pool.

I think that means there are no real blockers, except that we would like to setup deployment (with ansible and some other stuff). I tried to get cross-compilation to work briefly (from linux to freebsd), but my default solution of just using zig as the linker did not work. So I guess we'll use a freebsd docker container to do the compilation there. 

some extra libraries are needed for cross-compilation, but I cannot find a convenient way to get them:

```
  = note: /usr/bin/ld: cannot find -lexecinfo
          /usr/bin/ld: cannot find -lkvm
          /usr/bin/ld: cannot find -lmemstat
          /usr/bin/ld: cannot find -lkvm
          /usr/bin/ld: cannot find -lprocstat
          /usr/bin/ld: cannot find -ldevstat
          /usr/bin/ld: cannot find -lexecinfo
          collect2: error: ld returned 1 exit status
```

valpackett:
hmm. typically you should have these if you have the libc, they're all part of base. I've heard that zig can do cross-linking to libc without actually having the libc.so or something (???) buuut I still haven't actually looked into how that magic works.

You can just download and extract the base from https://download.freebsd.org/releases/amd64/13.2-RELEASE/base.txz and then extract it and use it as the `--sysroot`.

https://mcilloni.ovh/2021/02/09/cxx-cross-clang/ here's a nice article I just found that conveniently uses FreeBSD/aarch64 as the example target and shows a tar extraction command that only extracts the necessary directories :)

folkertdev:
well, I tried some variations of 

```
TARGET_CC="/home/runner/.cargo/bin/cargo-zigbuild zig cc -target x86_64-unknown-freebsd -g --sysroot=/home/folkertdev/Downloads/freebsd-sysroot/ -L/home/folkertdev/Downloads/freebsd-sysroot/lib" cargo build --target x86_64-unknown-freebsd --bin ntp-daemon
```

here `zig cc` is just clang, but with many cross-compilation targets included. the zig executable bundles the headers of libc for those platforms. Sadly, this just still gives the same error as before. Worth a try, but no luck so far. Doing this will docker will be fine, and should just work (no need to download 1gb of stuff, either)

valpackett:
Well if it's literally the same that just means it's just trying to use `/usr/bin/ld` as the linker.. maybe [telling cargo to use lld](https://stackoverflow.com/questions/63739813/error-linker-cc-not-found-when-cross-compiling-a-rust-project-from-windows) could work but idk why `zig cc` doesn't do this itself

igalic:
> > There does not seem to be something equivalent for rc.d
> 
> There is no equivalent for Linux capabilities _in the FreeBSD kernel_, currently there's no API that allows you to grant "the ability to do just this one thing that root can do" to a program. (Hm, adding this to my future-kernel-proposals-to-prototype list!…)

we have [`mac(4)`](https://man.freebsd.org/mac(4)).
but the only out of the box policy that is really easy to use is [`mac_portacl(4)`](https://man.freebsd.org/mac_portacl(4))

note that the ntpd we have in base uses mac: https://github.com/search?q=repo%3Afreebsd%2Ffreebsd-src%20path%3Acontrib%2Fntp%20TRUSTEDBSD_MAC&type=code 

it appears to be just using this policy: [`mac_ntpd(4)`](https://man.freebsd.org/mac_ntpd(4)), which imo deserves a see also entry in [`mac(4)`](https://man.freebsd.org/mac(4)).

valpackett:
Oh! Cool! That's pretty nice actually and usable immediately.

The linked code in the ntpd itself referring to mac is just for the checks like "getuid() is zero or this thing is on" and returning the nicer error message; it's not required to just have it working.

folkertdev:
I'm having some trouble actually applying this.  https://reviews.freebsd.org/D16281 seems to mostly make changes to C code (which would change how ntpd or freebsd is compiled). It is not clear to me how we'd use it at runtime to drop or not even require root privileges. So 

- should that `rc.d` file change, if yes how?
- any other things that need to be set up (like users/groups and such)

valpackett:
It's the [manpage](https://man.freebsd.org/cgi/man.cgi?mac_ntpd(4)) that's the thing to read, not necessarily the code. tl;dr:

- load the policy right now: `kldload mac_ntpd`
- tell the loader to load it on boot: `echo 'mac_ntpd_load="YES"' >> /boot/loader.conf`
- just run ntpd-rs as the (preexisting in the stock install) `ntpd` user (uid 123)

folkertdev:
Nice, thanks for the help here. We now have this config running

```
#!/bin/sh

. /etc/rc.subr

# PROVIDE: ntp_daemon
# REQUIRE: DAEMON FILESYSTEMS devfs
# BEFORE:  LOGIN
# KEYWORD: nojail resume shutdown

name="ntp_daemon"
rcvar="ntp_daemon_enable"

: ${ntp_daemon_enable:="NO"}

start_precmd="${name}_precmd"

command="/usr/sbin/daemon"
procname="daemon"
pidfile="/var/run/${name}.pid"
command_args=" -f -o /var/log/ntp_daemon.log -H -P ${pidfile} /usr/local/bin/ntp-daemon --"

load_rc_config $name
run_rc_command "$1"
```

it is now available in the ntp pool https://www.ntppool.org/scores/35.204.193.221 (but still needs to get a better score befor it fully joins.

folkertdev:
just adding some final stuff so that it's public, may be helpful for someone.

we have this vagrant setup that runs a FreeBSD VM for the compilation. This takes ~6 minutes for me. Acceptable, but not great. And it requires a lot of dependencies.

```ruby
Vagrant.configure("2") do |config|
  config.vm.box = "generic/freebsd13"

  config.ssh.password = "vagrant"

  config.vm.synced_folder "ntpd-rs/", "/ntpd-rs"

  config.vm.provision "shell", inline: <<-SHELL
    pkg install -y curl
    pkg install -y ca_root_nss
    fetch https://sh.rustup.rs -o rustup-init.sh
    sh rustup-init.sh -y
  SHELL

  # config.vm.provision "file", source: "ntpd-rs", destination: "/tmp/ntpd-rs"

  config.vm.provision "shell", inline: <<-SHELL
    cd /ntpd-rs
    cargo build --release -p ntpd --target-dir target-freebsd
  SHELL
end
```

It would be great if FreeBSD just worked with cargo-zigbuild. I suspect some equivalent of the macos sdk is needed here:

https://github.com/rust-cross/cargo-zigbuild/blob/main/Dockerfile

to provide the system libraries.

then a script to actually run the vagrant VM

```sh
#!/bin/bash

# To use this, make sure (hashicorp) vagrant is installed, normally it should
# be enough to run
#
#   sudo apt install vagrant
#
# But this may not install the most recent version. In that case, find the most recent version here
# https://releases.hashicorp.com/vagrant and install it using
#
# ```sh
# wget https://releases.hashicorp.com/vagrant/2.3.7/vagrant_2.3.7-1_amd64.deb
# sudo dpkg -i vagrant_2.3.7-1_amd64.deb
# ```
#
# Then either install virtualbox, or use the vagrant-libvirt plugin, for the
# latter one you can use these commands:
#
# ```sh
# sudo apt-get install -y libvirt-daemon-system libvirt-dev ebtables libguestfs-tools ruby-fog-libvirt
# vagrant plugin install vagrant-libvirt
# ```
# now run
#
#    vagrant up --provision
#
# you may need to comment out this line in the Vagrantfile
#
#     config.ssh.username = "vagrant"
#
# this should make vagrant set up the ssh details

vagrant up
mkdir freebsd-out
vagrant ssh-config > freebsd-out/tmp.conf
scp -F freebsd-out/tmp.conf default:/ntpd-rs/target-freebsd/release/ntp-daemon ./freebsd-out/ntp-daemon
chmod 0755 ./freebsd-out/ntp-daemon
scp -F freebsd-out/tmp.conf default:/ntpd-rs/target-freebsd/release/ntp-ctl ./freebsd-out/ntp-ctl
chmod 0755 ./freebsd-out/ntp-ctl
scp -F freebsd-out/tmp.conf default:/ntpd-rs/target-freebsd/release/ntp-metrics-exporter ./freebsd-out/ntp-metrics-exporter
chmod 0755 ./freebsd-out/ntp-metrics-exporter
rm freebsd-out/tmp.conf
vagrant destroy --force
``` 

and then the playbook

```yml
- hosts: server_freebsd
  become: yes
  tasks:
    - name: "NTPD-rs: Create group"
      group:
        name: "ntpd-rs"
        system: yes

    - name: "NTPD-rs: Create user"
      user:
        name: "ntpd-rs"
        system: yes
        createhome: no
        group: "ntpd-rs"

    - name: "NTPD-rs: Copy binaries"
      copy:
        src: freebsd-out/{{item}}
        dest: /usr/local/bin/{{item}}
        mode: 0755
      loop:
        - ntp-daemon
        - ntp-ctl
        - ntp-metrics-exporter

    - name: "Create run folder"
      file:
        path: /var/run/ntpd-rs
        state: directory

    - name: "Create config folder"
      file:
        path: /etc/ntpd-rs
        state: directory

    - name: "Copy ntpd-rs config"
      copy:
        src: ntp.client.toml
        dest: /etc/ntpd-rs/ntp.toml

    - name: "Copy ntpd-rs rc.d config"
      copy:
        src: freebsd-rc.sh
        dest: /etc/rc.d/ntp_daemon
        mode: 0755

    - name: "Disable legacy ntp"
      command: sysrc ntpd_enable=NO

    - name: "Set boot for ntp_daemon service"
      command: sysrc ntp_daemon_enable=YES

    - name: "Make sure legacy ntpd is stopped"
      command: service ntpd stop

    - name: "Stop ntp_daemon service"
      command: service ntp_daemon stop

    - name: "Start ntp_daemon service"
      command: service ntp_daemon start
```

ntpd-rs.rc.d.example
```diff
@@ -1,35 +1,23 @@
...
 
-# start_cmd="${name}_start" 
-stop_cmd=":" 
 start_precmd="${name}_precmd"
 
-command="/usr/local/bin/ntp-daemon"
-pidfile="/var/run/ntpd-rs.pid"
-required_files="$command"
+command="/usr/sbin/daemon"
+procname="daemon"
```
igalic, line 18: I would set this to ntp-daemon, since you may have multiple rc scripts on a single system that use daemon.

igalic (changes requested):
👀

davidv1992 (commented):
Other than the comment of igalic, this looks good to me.

PR #768, opened 2023-06-01
Added shim for rand in fuzzers to make them deterministic.

davidv1992:
Since cargo fuzz uses coverage based fuzzing, this helps its performance.


fuzz/fuzz_targets/cookie_parsing_sound.rs
```diff
@@ -8,11 +8,5 @@ fuzz_target!(|data: Vec<u8>| {
...
     let keyset = provider.get();
 
-    if let Ok(decoded1) = keyset.decode_cookie_pub(&data) {
-        let encoded1 = keyset.encode_cookie_pub(&decoded1);
-        let decoded2 = keyset.decode_cookie_pub(&data).unwrap();
-        let encoded2 = keyset.encode_cookie_pub(&decoded2);
-
-        assert_eq!(encoded1, encoded2);
-    }
+    let _ = keyset.decode_cookie_pub(&data);
```
folkertdev, line 11: this now no longer fuzzes cookie encoding. is that intentional?

folkertdev (approved):

fuzz/fuzz_targets/cookie_parsing_sound.rs
davidv1992: yes, the encoding wasn't hit anyways (since that would essentially involve the fuzzer breaking AES_SIV_CMAC_512), and as it is nondeterministic on purpose, the  check would have always failed since a cookie really shouldn't encode the same way twice (they should be using different nonces for one)

PR #772, opened 2023-06-02
Kalman filter improvements

davidv1992:
This simplifies the measurement model used by the kalman filter, hopefully somewhat improving the synchronization quality. The previous one was based on imagined limitations of the kalman filter.

### DIFF
--- a/ntp-proto/src/algorithm/kalman/peer.rs
+++ b/ntp-proto/src/algorithm/kalman/peer.rs
@@ -159,7 +159,7 @@ impl PeerFilter {
             "time {time:?} is before filter_time {:?}",
             self.filter_time
         );
-        if time < self.filter_time {
+        if time.is_before(self.filter_time) {
             return;
         }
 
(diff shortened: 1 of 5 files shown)


ntp-proto/src/algorithm/kalman/peer.rs
```diff
@@ -709,7 +703,7 @@ mod tests {
         let mut peer = PeerState(PeerStateInner::Stable(PeerFilter {
             state: Vector::new_vector([20e-3, 0.]),
             uncertainty: Matrix::new([[1e-6, 0.], [0., 1e-8]]),
-            clock_wander: 1e-8,
+            clock_wander: 0e-8,
```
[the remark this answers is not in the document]
davidv1992: sorry, should have marked this as work in progress first.

ntp-proto/src/algorithm/kalman/peer.rs
```diff
@@ -90,6 +90,22 @@ struct AveragingBuffer {
...
+    const P: f64 = 0.3275911;
+    const A1: f64 = 0.254829592;
+    const A2: f64 = -0.284496736;
+    const A3: f64 = 1.421413741;
+    const A4: f64 = -1.453152027;
+    const A5: f64 = 1.061405429;
+
+    let x = (chi / 2.).sqrt();
+    let t = 1. / (1. + P * x);
+    (A1 * t + A2 * t * t + A3 * t * t * t + A4 * t * t * t * t + A5 * t * t * t * t * t)
```
folkertdev, line 105: are there disadvantages to using https://doc.rust-lang.org/std/primitive.f64.html#method.powi here?

davidv1992: I didn't find it much cleaner (because of brackets) but other than that no.

PR #785, opened 2023-06-13
prevent confusing peer rejected messages

folkertdev:
on startup we'd often see

```
2023-06-13T09:08:03.399656Z  WARN ntp_proto::peer: Peer rejected due to invalid stratum stratum=16
2023-06-13T09:08:03.401224Z  WARN ntp_proto::peer: Peer rejected due to invalid stratum stratum=16
2023-06-13T09:08:03.401353Z  WARN ntp_proto::peer: Peer rejected due to invalid stratum stratum=16
2023-06-13T09:08:03.401475Z  WARN ntp_proto::peer: Peer rejected due to invalid stratum stratum=16
```

this was entirely expected based on our code, but is very confusing to someone unfamiliar with the implementation: the message suggests something is wrong.

### DIFF
--- a/ntp-daemon/src/peer.rs
+++ b/ntp-daemon/src/peer.rs
@@ -133,10 +133,17 @@ where
         self.last_poll_sent = Instant::now();
         self.update_poll_wait(poll_wait, system_snapshot);
 
-        // NOTE: fitness check is not performed here, but by System
-        let snapshot = PeerSnapshot::from_peer(&self.peer);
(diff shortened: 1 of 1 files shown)


davidv1992 (commented):
Not sure I agree with this approach, as it also delays any peer state from being visible to observe until it has responded at least once, which may never happen.

davidv1992 (commented):
Ok, after discussion approach seems bit hacky but mostly okay, needs fix for tests though.

davidv1992 (approved):

PR #795, opened 2023-06-15
Implemented quick startup to first time adjustment.

davidv1992:


ntp-proto/src/algorithm/kalman/peer.rs
```diff
@@ -461,7 +466,49 @@ impl PeerState {
...
+                    .copied()
+                    .fold(None, |v1, v2| {
+                        if v2.is_nan() {
+                            v1
+                        } else if let Some(v1) = v1 {
+                            if v1 > v2 {
+                                Some(v1)
+                            } else {
+                                Some(v2)
+                            }
```
folkertdev, line 486: maybe use `Some(Ord::max(v1, v2))`

davidv1992: Can't, f64 is not Ord (only partial_ord because of NaN)

folkertdev (approved):

PR #804, opened 2023-06-23
Reworked crypto interface to be easier to use and compatible with other crypto libraries.

davidv1992:
These changes aim to make the crypto interface generic over both library and actual algorithm again (since tags are highly cryptographic algorithm specific), whilst keeping the non-allocating nature of encryption we currently have.


ntp-proto/src/keyset.rs
```diff
@@ -170,24 +164,27 @@ impl KeySet {
...
-        let mut plaintext = Self::plaintext(cookie);
-        let plaintext_len = plaintext.as_slice().len();
-
-        let (siv_tag, nonce) = self.keys[self.primary as usize]
-            .encrypt_in_place_detached(plaintext.as_mut(), &[])
+        let mut output = Self::plaintext(cookie);
+        let plaintext_length = output.as_slice().len();
+
+        // Add space for header (4 + 2 bytes), tag (16 bytes) and nonce (16 bytes) plus some margin (16 bytes)
+        output.resize(output.len() + 2 + 4 + 16 + 16 + 16, 0);
```
folkertdev, line 171: what is this margin? we should be able to know exactly how much space we need. Also what is the tag? we don't add the tag here (it's just header, nonce, and ciphertext)

ntp-proto/src/packet/crypto.rs
```diff
@@ -11,13 +11,64 @@ use super::extensionfields::ExtensionField;
...
+    }
+
+    fn truncate(&mut self, len: usize) {
+        self.valid = std::cmp::min(self.valid, len);
+    }
+}
+
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub struct EncryptResult {
+    // Nonce length MUST be a multiple of 4 when encrypting
```
folkertdev, line 58: why is this? is this enforced?

ntp-proto/src/keyset.rs
```diff
@@ -170,24 +164,27 @@ impl KeySet {
     }
 
     pub(crate) fn encode_cookie(&self, cookie: &DecodedServerCookie) -> Vec<u8> {
-        let mut plaintext = Self::plaintext(cookie);
-        let plaintext_len = plaintext.as_slice().len();
-
-        let (siv_tag, nonce) = self.keys[self.primary as usize]
-            .encrypt_in_place_detached(plaintext.as_mut(), &[])
+        let mut output = Self::plaintext(cookie);
```
folkertdev, line 167: I think we now can/should make this a method on `DecodedServerCookie`.

ntp-proto/src/packet/crypto.rs
```diff
@@ -205,3 +262,46 @@ impl std::fmt::Debug for AesSivCmac512 {
...
+        let key = AesSivCmac256::new([0u8; 32].into());
+        let EncryptResult {
+            nonce_length,
+            ciphertext_length,
+        } = key.encrypt(&mut testvec, 16, &[]).unwrap();
+        let result = key
+            .decrypt(
+                &testvec[..nonce_length],
+                &testvec[nonce_length..(nonce_length + ciphertext_length)],
+                &[],
```
folkertdev, line 283: should we also test with associated data at this point? or else just remove it from the `Cipher` trait if we don't use it.

folkertdev (commented):
so I'm not sure 1) what sparked these changes or 2) what the design constraints are, but I'm not a fan. There are lots of indexing operations that can panic, and implicit assumptions about lengths. 

I think we can do better and this needs another design iteration

ntp-proto/src/packet/crypto.rs
davidv1992: will remove this restriction

ntp-proto/src/packet/crypto.rs
```diff
@@ -11,13 +11,64 @@ use super::extensionfields::ExtensionField;
 #[error("Could not decrypt ciphertext")]
 pub struct DecryptError;
 
+struct Buffer<'a> {
+    buffer: &'a mut [u8],
+    valid: usize,
+}
```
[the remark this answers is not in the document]
davidv1992: It pretty much has a similar function, but also with the interface aes_siv requires.

ntp-proto/src/keyset.rs
davidv1992: agreed, done

ntp-proto/src/keyset.rs
davidv1992: Fixed and clarified calculation.

As for tag, see above

ntp-proto/src/packet/crypto.rs
```diff
@@ -99,24 +150,27 @@ impl Drop for AesSivCmac256 {
...
-                return Err(std::io::Error::from(std::io::ErrorKind::Other));
-            }
-        };
+        buffer.copy_within(..plaintext_length, nonce.len());
+        buffer[..nonce.len()].copy_from_slice(&nonce);
+
+        let mut buffer_wrap = Buffer::new(&mut buffer[nonce.len()..], plaintext_length);
+
+        siv.encrypt_in_place([associated_data, &nonce], &mut buffer_wrap)
+            .map_err(|_| std::io::ErrorKind::Other)?;
```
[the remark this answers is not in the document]
davidv1992: Added comments and proper check on sizes.

Ciphertext and plaintext length shouldn't be the same, this still adds the tag, but that detail is handled internally in the encrypt_in_place function

folkertdev (approved):
this looks good now, but maybe merge https://github.com/pendulum-project/ntpd-rs/pull/888 in first?

PR #828, opened 2023-06-30
Minor fixes to nts peer documentation.

davidv1992:


folkertdev (approved):

CONFIGURATION.md
```diff
@@ -162,13 +162,12 @@ addr = "1.pool.ntp.org:123"
...
 
-A peer in `server-nts` mode will use NTS (Network Times Security) to communicate with its server. The server must support NTS. The configuration requires the address of the Key Exchange server (the address of the actual NTP server that ends up being used may be different). For example:
+A peer in `nts-server` mode will use NTS (Network Times Security) to communicate with its server. The server must support NTS. The configuration requires the address of the Key Exchange server (the address of the actual NTP server that ends up being used may be different). For example:
 
 ```
 [[peers]]
-mode = "server-nts"
+mode = "nts-server"
 ke-addr = "time.cloudflare.com:4460"
-certificate = "/path/to/certificates.pem"
```
folkertdev, line 171: wait, why is this line removed? the path to the certificate seems kind of important?!

davidv1992: For a basic example with a server with a public certificate, it is not needed, and only adds confusion. The certificate path is only needed for local testing with root certificates not in the system's list, which is an advanced usecase for which the rest of the documentation should be sufficient.

PR #834, opened 2023-07-03
enable frequency holding

folkertdev:
Based on statime code. And based on the linux documentation, it sounds like this is something that we want.

https://man.archlinux.org/man/core/man-pages/clock_adjtime.2.en#STA_FREQHOLD

### DIFF
--- a/ntp-os-clock/src/lib.rs
+++ b/ntp-os-clock/src/lib.rs
@@ -18,7 +18,7 @@ pub enum Error {
     #[error("Invalid operation requested")]
     Invalid,
     #[error("Clock device has gone away")]
-    NoDev,
+    NoDevice,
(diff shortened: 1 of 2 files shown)


davidv1992:
Freqhold should only be enabled when not using the NTP algorithm. I think the idea we had here originally to only change the status bits in the enable/disable ntp discipline function seems like the best method. Hence, I am against the approach taken by this PR.

folkertdev:
I agree. I would like to understand though what `STA_FREQHOLD` does. Is it relevant when `ADJ_OFFSET` is not used. Could we just use `ADJ_SETOFFSET` in statime instead and not bother with the frequency holding at all there?

> Freqhold should only be enabled when not using the NTP algorithm

what is "the" NTP algorithm here. the algorithm described in the spec, or any NTP algorithm?

PR #873, opened 2023-07-17
Add missing semicolon

sylvestre:


folkertdev (approved):
Thanks,

How did you find these cases? Clippy and cargo format don't report these.

Also in general these semicolons are not "missing". I suppose their omission might be confusing to someone with a C background, but they are not required for rust to have the right semantics.

PR #888, opened 2023-07-24
Crypto refactor nonce padding

folkertdev:
if we allow it, we should test it

### DIFF
--- a/ntp-proto/src/packet/extensionfields.rs
+++ b/ntp-proto/src/packet/extensionfields.rs
@@ -258,12 +258,9 @@ impl<'a> ExtensionField<'a> {
 
         let header_start = w.position();
 
-        // Temporary header
-        w.write_all(
-            &ExtensionFieldTypeId::NtsEncryptedField
-                .to_type_id()
-                .to_be_bytes(),
-        )?;
+        // Placeholder header
+        let type_id: u16 = ExtensionFieldTypeId::NtsEncryptedField.to_type_id();
+        w.write_all(&type_id.to_be_bytes())?;
         w.write_all(&0u16.to_be_bytes())?;
         w.write_all(&0u16.to_be_bytes())?;
         w.write_all(&0u16.to_be_bytes())?;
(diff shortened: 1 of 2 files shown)


ntp-proto/src/packet/extensionfields.rs
```diff
@@ -300,27 +297,39 @@ impl<'a> ExtensionField<'a> {
...
+        let nonce_padding = padded_nonce_length - nonce_length;
+        cur_extension_field[header_size + nonce_length..][..nonce_padding]
+            .copy_from_slice(&padding[..nonce_padding]);
+
+        // zero out the ciphertext padding
+        let ciphertext_padding = padded_ciphertext_length - ciphertext_length;
+        debug_assert_eq!(
+            ciphertext_padding, 0,
+            "extension field encoding should add padding"
+        );
```
folkertdev, line 317: I believe this just should always be zero in practice. We don't test with a `ciphertext_padding > 0` at the moment, and I don't think it's possible to create with our logic a plaintext that would require ciphertext padding

ntp-proto/src/packet/extensionfields.rs
```diff
@@ -549,10 +562,6 @@ impl<'a> RawEncryptedField<'a> {
         let nonce_length = u16::from_be_bytes([b0, b1]) as usize;
         let ciphertext_length = u16::from_be_bytes([b2, b3]) as usize;
 
-        if nonce_length != 16 {
-            return Err(IncorrectLength);
-        }
-
```
folkertdev, line 555: this constraint is now no longer required, and an implementation detail of the ciphers that we provide today.

rnijveld:
The code LGTM, but I'm not enough into crypto to have anything else to say about this, maybe @davidv1992 can take a look as well?

ntp-proto/src/packet/extensionfields.rs
davidv1992: Given the multiple-of-4 requirement even on encrypted efs you might be right. I am comfortable with this debug assert for now.

PR #900, opened 2023-07-27
Rename config vars

cikzh:
This PR just renames the easily-renamable-low-hanging-fruit config vars. The other renames aren't arbitrary and will be done in a seperate PR.

- Renamed Peer related config vars
- Now also throwing an error when the `denylist-action` is present, but the `denylist` is missing
- Added some tests to test that logic

@davidv1992 Added you as a reviewer to double check the denylist(-action) logic


davidv1992 (commented):
Bit late, but the behaviour around allow and denylist looks good to me.

PR #904, opened 2023-07-28
Set lower minimum dependencies in our Cargo.toml

rnijveld:
These lower minimum versions were set by lowering the version constraints until either tests started failing or compilation failed. I've also changed dependabot to no longer update our Cargo.toml, although that will also result in dependabot no longer giving us notifications for any semver incompatible changes, but unfortunately the `increase-if-necessary` dependabot strategy is not supported for cargo. But if we are packaging into OSes then we should treat backwards incompatible upgrades with care anyway. Note that our lockfile still contains the latest versions, and we probably want to add a CI step that runs every once in a while to check that our minimum versions still compile and test correctly.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -5,9 +5,10 @@ updates:
   schedule:
     interval: daily
   open-pull-requests-limit: 10
-  
+
 - package-ecosystem: cargo
   directory: "/"
   schedule:
     interval: daily
+  versioning-strategy: lockfile-only
   open-pull-requests-limit: 10
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -43,17 +43,17 @@ panic = "abort"
 strip = "debuginfo"
 
 [workspace.dependencies]
-tracing = "0.1.37"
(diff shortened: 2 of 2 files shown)


davidv1992 (commented):
Agree with the intent of the changes here. However, this needs quite a bit of testing, especially around the toml and prometheus-client crates where we now support 3 versions that according to semver aren't interchangable. I currently don't have the time to do that properly.

As for CI testing this, I would prefer to have a CI job applying https://github.com/rust-lang/cargo/issues/5657 if possible before merging this (and if that works by now for all the crates we depend on), but it is not a showstopper for me if that turns out to be too much work.

rnijveld:
Both toml and prometheus-client have changes that are irrelevant for us. Prometheus-client specifically also has updated their semver version when no such increase was necessary. I do want to add a CI job that tests against these minimal dependencies, but I would suggest we do this on a nightly/weekly schedule instead, I'll add that with a separate PR.

davidv1992 (approved):

PR #919, opened 2023-08-07
update log levels in the ntp-udp crate

folkertdev:


folkertdev:
point of discussion: this bumps down some log messages from warn to info. Strictly speaking these log messages are emitted per-request (and so scale in the number of requests, which we do not want for the `info` log level).

But, those log message indicate a serious bug in ntpd-rs, and we'd like to know if that were the case, so @rnijveld and I decided that it is best to keep them on the info level.

ntp-udp/src/raw_socket.rs
```diff
@@ -253,14 +251,14 @@ mod recv_message {
         };
 
         if mhdr.msg_flags & libc::MSG_TRUNC > 0 {
-            warn!(
+            tracing::info!(
```
davidv1992, line 254: This loglevel seems still too high for me. This can be remotely triggered and doesn't neccessarily indicate problems on our end.

ntp-udp/src/raw_socket.rs
```diff
@@ -253,14 +251,14 @@ mod recv_message {
...
                 max_len = packet_buf.len(),
                 "truncated packet because it was larger than expected",
             );
         }
 
         if mhdr.msg_flags & libc::MSG_CTRUNC > 0 {
-            warn!("truncated control messages");
+            tracing::info!("truncated control messages");
```
davidv1992, line 261: I think this should only happen because of bugs on our end. As such, I think it should be higher than info, at least warn, and we should probably say something about this likely being a bug.

ntp-udp/src/socket.rs
```diff
@@ -179,7 +179,7 @@ impl UdpSocket {
                     .await
                 {
                     Err(_) => {
-                        warn!("Packet without timestamp");
+                        tracing::info!("Packet without timestamp");
```
davidv1992, line 182: I think this at this point does indicate a problematic condition (we asked the OS for timestamps, but ain't getting them), which the end user should probably want to fix. Hence I think warn is more appropriate

ntp-udp/src/socket.rs
```diff
@@ -304,13 +305,14 @@ fn recv(
...
+                tracing::info!("unexpected error message on the MSG_ERRQUEUE");
             }
 
             ControlMessage::Other(msg) => {
-                warn!(
+                tracing::info!(
                     "weird control message {:?} {:?}",
-                    msg.cmsg_level, msg.cmsg_type
+                    msg.cmsg_level,
+                    msg.cmsg_type
```
davidv1992, line 315: This feels to me like we should either
 - treat it like a bug (hence warn)
 - treat it as non-important operating system behaviour (hence debug  or lower)

The current choice feels weird to me.

ntp-udp/src/socket.rs
```diff
@@ -304,13 +305,14 @@ fn recv(
 
             #[cfg(target_os = "linux")]
             ControlMessage::ReceiveError(_error) => {
-                warn!("unexpected error message on the MSG_ERRQUEUE");
+                tracing::info!("unexpected error message on the MSG_ERRQUEUE");
```
davidv1992, line 308: This feels to me like we should either
 - treat it like a bug (hence warn)
 - treat it as non-important operating system behaviour (hence debug  or lower)

The current choice feels weird to me.

ntp-udp/src/socket.rs
```diff
@@ -352,26 +354,29 @@ fn fetch_send_timestamp_help(
...
             }
 
             ControlMessage::Other(msg) => {
-                warn!(
+                tracing::info!(
                     msg.cmsg_level,
-                    msg.cmsg_type, "unexpected message on the MSG_ERRQUEUE",
+                    msg.cmsg_type,
+                    "unexpected message on the MSG_ERRQUEUE",
                 );
```
davidv1992, line 380: This feels to me like we should either
 - treat it like a bug (hence warn)
 - treat it as non-important operating system behaviour (hence debug  or lower)

The current choice feels weird to me.

davidv1992 (commented):
Some comments on a number of the classification choices. Particularly feel uncomfortable with the apparant logging of bugs on info level, which just seems wrong to me.

ntp-udp/src/raw_socket.rs
davidv1992: Choice: debug

ntp-udp/src/raw_socket.rs
davidv1992: Choice: warn

ntp-udp/src/socket.rs
davidv1992: Choice: warn, unless we find that this causes real trouble

ntp-udp/src/socket.rs
davidv1992: Choice: warn

ntp-udp/src/socket.rs
davidv1992: warn

ntp-udp/src/socket.rs
```diff
@@ -352,26 +354,29 @@ fn fetch_send_timestamp_help(
...
 
                 // Check that this message belongs to the send we are interested in
                 if error.ee_data != expected_counter {
-                    warn!(
+                    tracing::info!(
                         error.ee_data,
-                        expected_counter, "Timestamp for unrelated packet"
+                        expected_counter,
+                        "Timestamp for unrelated packet"
                     );
```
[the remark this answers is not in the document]
davidv1992: choice: debug

ntp-udp/src/socket.rs
davidv1992: warn

davidv1992 (approved):

PR #920, opened 2023-08-07
Ntp proto log levels

folkertdev:


ntp-proto/src/peer.rs
```diff
@@ -227,13 +228,13 @@ impl PeerSnapshot {
         // Note, this can only ever be an issue if the peer is not using
         // hardware as its source, so ignore reference_id if stratum is 1.
         if self.stratum != 1 && self.reference_id == self.our_id {
-            debug!("Peer rejected because of detected synchornization loop");
+            warn!("Peer rejected because of detected synchornization loop");
```
davidv1992, line 231: This can happen during normal operation and is not necessarily the result of any misconfiguration, so we should probably log this as just info.

ntp-proto/src/peer.rs
```diff
@@ -227,13 +228,13 @@ impl PeerSnapshot {
...
             return Err(Loop);
         }
 
         // An unreachable error occurs if the server is unreachable.
         if !self.reach.is_reachable() {
-            warn!("Peer unreachable");
+            info!("Peer is unreachable");
```
davidv1992, line 237: This however does probably signal a problem, so I think we should keep this as a warning.

ntp-udp/src/raw_socket.rs
```diff
@@ -253,14 +251,14 @@ mod recv_message {
         };
 
         if mhdr.msg_flags & libc::MSG_TRUNC > 0 {
-            warn!(
+            tracing::info!(
                 max_len = packet_buf.len(),
                 "truncated packet because it was larger than expected",
             );
```
davidv1992, line 257: This is remote triggerable, and not necessarily a bug on our end, so suggest we make this debug.

ntp-udp/src/raw_socket.rs
```diff
@@ -253,14 +251,14 @@ mod recv_message {
...
         }
 
         if mhdr.msg_flags & libc::MSG_CTRUNC > 0 {
-            warn!("truncated control messages");
+            tracing::info!("truncated control messages");
         }
```
davidv1992, line 262: This very likely is a bug on our end, so should be warn in my view (also not remotely triggerable unless I misunderstand control messages)

ntp-udp/src/socket.rs
```diff
@@ -179,7 +179,7 @@ impl UdpSocket {
                     .await
                 {
                     Err(_) => {
-                        warn!("Packet without timestamp");
+                        tracing::info!("Packet without timestamp");
                         Ok((send_size, None))
```
davidv1992, line 183: This feels to me like it should be a warning, as we are not getting from the operating system what we asked for, and the user likely does want to resolve this.

ntp-udp/src/socket.rs
```diff
@@ -304,13 +305,14 @@ fn recv(
 
             #[cfg(target_os = "linux")]
             ControlMessage::ReceiveError(_error) => {
-                warn!("unexpected error message on the MSG_ERRQUEUE");
+                tracing::info!("unexpected error message on the MSG_ERRQUEUE");
             }
```
davidv1992, line 309: This feels like it is either a bug (and should thus be warn) or some os oddity that we just want to log as debug.

ntp-udp/src/socket.rs
```diff
@@ -304,13 +305,14 @@ fn recv(
...
 
             ControlMessage::Other(msg) => {
-                warn!(
+                tracing::info!(
                     "weird control message {:?} {:?}",
-                    msg.cmsg_level, msg.cmsg_type
+                    msg.cmsg_level,
+                    msg.cmsg_type
```
davidv1992, line 315: This feels like it is either a bug (and should thus be warn) or some os oddity that we just want to log as debug.

ntp-udp/src/socket.rs
```diff
@@ -352,26 +354,29 @@ fn fetch_send_timestamp_help(
...
                 }
 
                 // Check that this message belongs to the send we are interested in
                 if error.ee_data != expected_counter {
-                    warn!(
+                    tracing::info!(
                         error.ee_data,
-                        expected_counter, "Timestamp for unrelated packet"
+                        expected_counter,
+                        "Timestamp for unrelated packet"
```
davidv1992, line 369: I think this should just be debug. The actual problem is that we didn't get the timestamp at the point we expected it, and we should log it as a warning there.

ntp-udp/src/socket.rs
```diff
@@ -352,26 +354,29 @@ fn fetch_send_timestamp_help(
...
                 }
             }
 
             ControlMessage::Other(msg) => {
-                warn!(
+                tracing::info!(
                     msg.cmsg_level,
-                    msg.cmsg_type, "unexpected message on the MSG_ERRQUEUE",
+                    msg.cmsg_type,
+                    "unexpected message on the MSG_ERRQUEUE",
```
davidv1992, line 379: This feels like it is either a bug (and should thus be warn) or some os oddity that we just want to log as debug.

davidv1992 (commented):
Number of minor comments on level choices

ntp-proto/src/nts_record.rs
```diff
@@ -642,7 +642,7 @@ impl KeyExchangeResultDecoder {
                 Break(Err(error))
             }
             Warning { warningcode } => {
-                tracing::warn!(warningcode, "Received key exchange warning code");
+                tracing::info!(warningcode, "Received key exchange warning code");
```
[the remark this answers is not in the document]
davidv1992: choice: warn

ntp-proto/src/nts_record.rs
```diff
@@ -878,7 +878,7 @@ impl KeyExchangeServerDecoder {
...
```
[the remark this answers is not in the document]
davidv1992: choice: debug

ntp-proto/src/peer.rs
davidv1992: choice:info

davidv1992 (approved):

PR #921, opened 2023-08-07
adjust log levels for ntp-daemon

folkertdev:


ntp-daemon/src/nts_key_provider.rs
```diff
@@ -19,7 +19,7 @@ pub async fn spawn(config: KeysetConfig) -> watch::Receiver<Arc<KeySet>> {
             .await
             .unwrap_or_else(|e| Err(std::io::Error::new(std::io::ErrorKind::Other, e)))
             .unwrap_or_else(|e| {
-                error!(error = ?e, "Could not load nts server keys, starting with new set");
+                info!(error = ?e, "Could not load nts server keys, starting with new set");
```
davidv1992, line 22: I think info is too low here, as this is a problem unless this is the very first time you start the server.

davidv1992 (commented):
One minor comment on choice of level. Also, we may want to consider adding the "This is a bug!" message in a lot more places.

ntp-daemon/src/nts_key_provider.rs
davidv1992: choice: warn

PR #949, opened 2023-08-18
Moved some docs into the docs folder

rnijveld:
This does a little more clean up of the docs by moving them into the docs folder and restructuring stuff a little bit.

This also fixes formatting for the plaintext files, where they now have a maximum line length of 80. Depending on word wrap of your editor results in messy to read text at best, so I think it's best not to depend on that anymore. Maybe we should look at a linting tool to check these kinds of things given how much documentation we still need to write?


folkertdev:
why would we still be constrained by 80 columns? that number is completely arbitrary and based on tech that is decades out of date. 80 columns is actually quite annoying to read on a screen that is much wider than that

README.md
```diff
@@ -1,103 +1,151 @@
...
 ```sh
 systemctl stop chronyd
 ```
 
-The ntpd-rs daemon requires elevated permissions to change the system clock. It can be tested against servers in the [NTP pool](https://ntppool.org):
+The ntpd-rs daemon requires elevated permissions to change the system clock.
+
 ```sh
-sudo ./target/release/ntp-daemon -p pool.ntp.org -p pool.ntp.org -p pool.ntp.org -p pool.ntp.org
+sudo ./target/release/ntp-daemon
```
folkertdev, line 46: this will now (try to) use the `ntp.toml` at the default location. It is not guaranteed that this file exists, so it is quite likely that someone following these instructions now runs into a panic

README.md
```diff
@@ -1,103 +1,151 @@
...
+full-featured, it supports NTP client and server with NTS.
 
 Our roadmap for 2023:
 
-- Q2 2023: Adoption work, improved packaging, FreeBSD support
-- Q3 2023: Stable release (pending funding)
-- Q4 2023: Development work on experimental features, NTS pool, NTPv5 (pending funding)
+* Q2 2023: Adoption work, improved packaging, FreeBSD support
+* Q3 2023: Stable release (pending funding)
+* Q4 2023: Development work on experimental features, NTS pool, NTPv5 (pending
```
folkertdev, line 118: this funding has now been secured, right?

folkertdev (commented):
so yeah I don't think we should enforce the 80-character limit on a per-commit/per-PR basis. It seems impossible to do consistently across platforms and editors. We can perhaps add it to the list of steps to perform before a release.

rnijveld:
> why would we still be constrained by 80 columns? that number is completely arbitrary and based on tech that is decades out of date. 80 columns is actually quite annoying to read on a screen that is much wider than that

This is more related to still being easy to understand. It is generally accepted (and supported by some research) that long line length for prose results in lower reading speed and lesser understanding of the content. While the limit of 80 is somewhat arbitrary, it does coincidentally join up quite well with prose found in other places, such as on news websites or in books.

Additionally, some formatting goes completely out of the window when we would depend on word wrap in editors, such as when indented lists are being used, which makes of a messy and unapproachable source file for these texts which are supposed to be readable all the time.

I think we should approach text files no different than source code files, for Rust source there is also a soft limit on 100 characters. But the character density in source code is generally lower, so a longer line length generally works better for that, especially on modern screens. I could adjust the formatting for 100 characters, but I would still propose we introduce some (soft) limit and properly format these files instead of relying on editor functionality.

README.md
rnijveld: I'll add the `-c` option with the example config file in the root of the project to this example to make sure people don't run into that specific error.

README.md
rnijveld: I think so, but I'd rather let someone else do the content updates here.

folkertdev:
>  it does coincidentally join up quite well with prose found in other places, such as on news websites or in books.

well, that very line is ~120 characters wide. News websites are optimized for a mobile layout. We edit these raw markdown files in a text editor, on a screen that is much wider than 80 characters. They are markdown files, so for reading they would just be rendered by github at ~120 characters per line.

also, how should contributors make sure that this limit is respected? different tools make different layout choices for how to break up the text?

davidv1992 (approved):

PR #967, opened 2023-08-24
Improved user experience when setting up server.

### LINKED ISSUE #924: Improved server installation process
- Currently setting up ntpd-rs as a server requires modifying the systemd unit, and additionally requires figuring out how our server configuration works
- Setting up prometheus is not helped with in any way

davidv1992:
Fixes #924

### DIFF
--- a/docs/guide/configuration.md
+++ b/docs/guide/configuration.md
@@ -12,7 +12,7 @@ The recommended way of installing ntpd-rs is with the pre-built packages from th
 
 If you want to setup a server using these packages, you will need to uncomment the `server` section of the configuration in `/etc/ntpd-rs/ntp.toml`. To enable a metrics exporter providing prometheus metrics on port `8123`, the systemd unit `ntpd-rs-metrics` needs to be enabled and started.
 
-Alternatively, you can use `cargo install ntpd` or build from source.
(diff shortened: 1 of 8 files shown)


docs/guide/configuration.md
```diff
@@ -10,6 +10,8 @@ Unlike the ntp reference implementation, ntpd-rs does not support either broadca
 
 The recommended way of installing ntpd-rs is with the pre-built packages from the [releases](https://github.com/pendulum-project/ntpd-rs/releases) page. The installers automatically handle setting up users, permissions and configuration.
 
+If you want to setup a server using these packages, you will need to uncomment the `server` section of the configuration in `/etc/ntpd-rs/ntp.toml`. To enable a metrics exporter providing prometheus metrics on port `8123`, the systemd unit `ntpd-rs-metrics` needs to be enabled and started.
+
 Alternatively, you can use `cargo install ntpd` or build from source.
```
folkertdev, line 15: this alternatively is unclear now. `cargo instlal ntpd` will not setup a server, right?

davidv1992: Clarified the text, suggest that if we remove this we do so in a separate pr.

PR #990, opened 2023-08-25
Initial version of chrony migration guide.

davidv1992:


docs/guide/migrating-chrony.md
```diff
@@ -0,0 +1,154 @@
...
+The allow and deny list configuration is optional in ntpd-rs. By default, if a server is configured it will accept traffic from anywhere. When configuring both allow and deny lists, ntpd-rs will first check if a remote is on the deny list. Only if this is not the case will the allow list be considered. This needs to be taken into account when translating interleaved combinations of chrony's `allow` and `deny` commands.
+
+NTS can be enabled for a server by simply configuring an NTS key exchange server:
+```
+[[nts-ke]]
+key-exchange-listen = "<ip or [::]>:<
+certificate-chain-path = <ntsservercert>
+private-key-path = <ntsserverkey>
+```
+Here the names of the corresponding chrony directives are used on the right hand side of the assignment. Note that unlike chrony, ntpd-rs does not have a default ip address on which it listens for nts-ke traffic, and this needs to be explicitly provided. The port is optional however and defaults to the standard value 4460.
```
cikzh, line 120: Here the names of the corresponding chrony directives are used on the right hand side of the assignment. Note that unlike chrony, ntpd-rs does not have a default IP address on which it listens for nts-ke traffic, and thus needs to be explicitly provided. The port is optional however and defaults to 4460.

docs/guide/migrating-chrony.md
```diff
@@ -0,0 +1,154 @@
...
+
+The keys used to encrypt the cookies are by default ephemeral. If these should be kept across reboots of the server, the path for storing these can be configured:
```
cikzh, line 122: The keys used to encrypt the cookies are ephemeral by default. If these should be kept across reboots of the server, the path for storing these can be configured:

docs/guide/migrating-chrony.md
```diff
@@ -0,0 +1,154 @@
...
+
+Rotation of the keys is by default daily, with one full weeks worth of old keys remembered for serving clients with older cookies. This can be configured with the `key-rotation-interval` and `stale-key-count` parameters:
+```
+[keyset]
+stale-key-count = <number of old keys to keep>
+key-rotation-interval = <rotation interval in seconds>
+```
+Note that the defaults for these settings mean that, in contrast to the chrony defaults, cookies for the server are only valid for slightly more than 1 week instead of 3.
+
+Sharing the keys with which the nts cookies are encrypted between multiple ntpd-rs servers is not yet supported.
```
cikzh, line 137: Sharing the keys, with which the NTS cookies are encrypted between multiple ntpd-rs servers, is not yet supported.

docs/guide/migrating-chrony.md
```diff
@@ -0,0 +1,154 @@
...
+
+If a local stratum for the server is configured through `local stratum`, this can be configured in ntpd-rs through the `local-stratum` key:
+```
+[synchronization]
+local-stratum = <stratum>
+```
+
+Broadcast mode is not currently supported in ntpd-rs. If this is used in your current setup, configuring the ntp server via dhcp instead may be an alternative.
```
cikzh, line 145: Should we inform about security implications of broadcast mode here?

cikzh (changes requested):
Mostly just some spelling and grammatical nitpicks. Accept the changes as you see fit

docs/guide/migrating-chrony.md
```diff
@@ -0,0 +1,154 @@
...
+
+When doing this, it is also important to note a major philosophical difference between chrony and ntpd-rs. For chrony, the majority of the algorithm tuning parameters are set on an individual time source. Within ntpd-rs, all control of the filtering is done via global parameters. Although we do not expect this to be the case, should there be specific parameters you would wish to configure on a per-peer basis, please let us know so we can consider this for future releases.
+
+## Server configuration
+
+Server configuration in ntpd-rs works quite a bit differently from chrony. Rather than enabling time server functionality by `allow`ing remote connections to the server, one or more serving instances can be individually configured. Each of these comes with its own allow and deny list:
+```
+[[server]]
+listen="<ip or [::]>:<port>"
```
cikzh, line 99:
```suggestion
listen="<IP or [::]>:<port>"
```

docs/guide/migrating-chrony.md
```diff
@@ -0,0 +1,154 @@
...
+    <subnet4>
+]
+denylist-action = `deny`
+```
+The allow and deny list configuration is optional in ntpd-rs. By default, if a server is configured it will accept traffic from anywhere. When configuring both allow and deny lists, ntpd-rs will first check if a remote is on the deny list. Only if this is not the case will the allow list be considered. This needs to be taken into account when translating interleaved combinations of chrony's `allow` and `deny` commands.
+
+NTS can be enabled for a server by simply configuring an NTS key exchange server:
+```
+[[nts-ke]]
+key-exchange-listen = "<ip or [::]>:<
```
cikzh, line 116:
```suggestion
key-exchange-listen = "<IP or [::]>:<
```

docs/guide/migrating-chrony.md
```diff
@@ -0,0 +1,144 @@
+# Migrating from chrony
+
+Both chrony and ntpd-rs can serve a similar role on Unix systems. This guide is aimed to help those migrating machines currently running chrony to ntpd-rs. We assume here experience with the [chrony configuration format](https://chrony-project.org/doc/4.3/chrony.conf.html). A user with no or little chrony experience may be better of following our [getting started guide](getting-started.md).
+
+Before we start with the specifics of how to convert individual directives, first a few notes. In contrast with the chrony configuration format, which acts like a list of commands to the client, the ntpd-rs configuration is a toml file. In particular, this means configuration is done by giving values for properties. As such, fields cannot be repeated unless part of a list (such as with the `[[peers]]` sections and `[[server]]` sections). Comments can be added by starting them with a `#`. The remainder of the line is then ignored.
```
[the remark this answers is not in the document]
davidv1992: I think we should, as the toml file used by us is conceptually very different from how configuration in ntpd and chrony is done. (command based configuration vs structure based configuration)

docs/guide/migrating-chrony.md
davidv1992: Made some changes addressing this.

cikzh (approved):

PR #994, opened 2023-08-28
add migrating-ntpd.md

folkertdev:


davidv1992 (changes requested):
This needs significant checking and fixing before it can be merged. I have only taken a cursory look through the ntpd document and put it next to the documentation, but found at least the following:
 - refclock directive is not a thing for ntpd, it uses special ip addresses.
 - missing text about broadcast and manycast associations
 - allow and deny doesn't seem to be an ntpd thing, rather it seems to use restrict.

docs/guide/migrating-ntpsec.md
```diff
@@ -0,0 +1,177 @@
...
+mode = "pool"
+address = "pool.ntp.org"
+count = 4
+```
+
+A source in `pool` mode must explicitly define an integer `count`, the maximum number of connections from this pool. The ntpd-rs daemon will actively try to keep the pool "filled": new connections will be spun up if a source from the pool is unreachable.
+
+<!-- "broadcast" is mentioned in the docs but never defined. Maybe this is an error in their docs?
+Like ntpsec, ntpd-rs deliberately does not support symmetric and broadcasting association modes because these modes have security issues.
+-->
```
folkertdev, line 42: we can just not mention these modes at all because ntpsec should not have them? I'm just confused by their one mention of broadcast 

> While there is a rich set of options available, the only required option is one or more pool, server, peer, or broadcast commands.

pool, server and peer are all defined, but broadcast isn't anywhere.

davidv1992 (approved):

PR #996, opened 2023-08-28
Replace prometheus-client dependency with custom serialization logic

rnijveld:


ntpd/src/daemon/mod.rs
```diff
@@ -1,12 +1,3 @@
-//! This crate contains the main ntp-daemon code for ntpd-rs and is not intended as
-//! a public interface at this time. It follows the same version as the main ntpd-rs
-//! crate, but that version is not intended to give any stability guarantee. Use at
-//! your own risk.
-//!
-//! Please visit the [ntpd-rs](https://github.com/pendulum-project/ntpd-rs) project
-//! for more information.
-#![forbid(unsafe_code)]
```
folkertdev, line 8: removing this suggests that something unsafe might now be happening. can we move this forbid into `ntpd/src/lib.rs`?

folkertdev: ah I missed that that is indeed what happened

folkertdev (approved):
This looks good to me

I do think we should somehow test this output, maybe with snapshot tests?

PR #1008, opened 2023-08-31
config that always sends a demobilize

folkertdev:
a replacement for the test binary that always sent a demobilize response


davidv1992 (changes requested):
Config looks good, but this needs to be included in the CI job that checks whether configurations are still valid.

folkertdev:
added, but that CI step does not actually validate any configs right now (just prints the help text, with a successful status code I guess?)

davidv1992:
It will once the ntp-ctl stuff is merged.

PR #1015, opened 2023-09-01
Added guide on using NTS with ntpd-rs.

davidv1992:


docs/guide/nts.md
```diff
@@ -0,0 +1,109 @@
...
+
+Currently, ntpd-rs does not support using self-signed certificates. If non public certificates are needed, the best option is to create a private CA locally and use that to sign the server's certificate. Instructions for that are included below in the instructions for configuring ntpd-rs as an NTS server.
+
+## Configuring ntpd-rs as an NTS server
+
+To configure ntpd-rs as a server, we need to configure three things. First, ntpd-rs needs to be configured to work as a normal NTP server. Second, we need to enable the server components that allow key exchange with clients. Finally, we need to configure storage for internal keys to ensure clients don't need to reconnect on a reboot of the server.
+
+The process of setting up an NTS server is significantly harder than configuring an NTS time source. We assume here some experience with [setting up a non-NTS NTP server](TODO), and will assume you already have that running on the target machine.
+
+Furthermore, NTPD-rs currently does not support running an NTS server on a server without a domain name through which its clients reach it. This can however be a local domain name, such as one configured through `/etc/hosts`.
```
folkertdev, line 31: what does this mean? can we have an example here of what is/is not possible?

docs/guide/nts.md
```diff
@@ -0,0 +1,109 @@
...
+
+### Certificates
+
+Before setting up the key exchange part of the server, we will first need to create the required key material. We will focus here on locally signing all the certificates for our server. For setting up a public server with NTS support you will need to request a publicly signed certificate. We suggest to use [letsencrypt](https://letsencrypt.org/) for that use case.
```
folkertdev, line 35: the concept of a "key exchange" has not been introduced at this point. In effect you assume some knowledge of how NTS works under the hood. I think it makes sense to explicitly sketch out how NTS works and introduce some of the core concepts

docs/guide/nts.md
```diff
@@ -0,0 +1,109 @@
...
+### Configuring key exchange
+
+We are now ready to configure the key exchange part of NTS. This is used by clients when setting up a connection.
+```toml
+[[nts-ke-server]]
+key-exchange-listen = "[::]:4460"
+certificate-chain-path = "ntpd-rs.test.chain"
+private-key-path = "ntpd-rs.test.key"
+```
+Note that this needs to listen on its own port. Furthermore, the key exchange happens over a TCP connection rather than over UDP packets.
```
folkertdev, line 95: why is the tcp/udp distinction important here?

docs/guide/nts.md
```diff
@@ -0,0 +1,109 @@
...
+
+When configured as above, it is important to ensure that the TLS private key can only be read by NTP-daemon and trusted system administrators. An attacker can use this key material to compromise all connections to clients.
```
folkertdev, line 97: - "TLS" has not been introduced at this point
- NTP daemon does not require the `-`

docs/guide/nts.md
davidv1992: Fixed with larger introduction

docs/guide/nts.md
davidv1992: done

rnijveld:
@davidv1992 I've added my latest version of the NTS guide here. I've also included a separate private CA guide under the development docs. Best reviewed in rendered form I think.

davidv1992:
I would prefer to do any background deep-dives after discussing client configuration, as that is really simple and doesn't need the background I think.

PR #1016, opened 2023-09-01
Rename peer to source and simple to server

rnijveld:


davidv1992 (approved):

docs/guide/configuration.md
```diff
@@ -103,72 +103,71 @@ The ntp-daemon binary can be configured through two channels: via command line o
...
+| ----------------------------------- | ----------------------- | ------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| `-c <FILE>`, `--config <FILE>`      | `/etc/ntpd-rs/ntp.toml` | Which configuration file to use. When provided, the fallback locations are not used.                                                                                                                          |
+| `-l <LEVEL>`, `--log-level <LEVEL>` | From configuration file | Override for the configuration file `log-level` parameter, see explanation there.                                                                                                                             |
+| `-s <ADDR>`, `--server <ADDR>`      |                         | Respond as NTP server to packets arriving to the given address, overrides server configuration in the configuration file. Can be given multiple times to attach as NTP server to multiple network interfaces. |
```
folkertdev, line 110: so the `-p` flag disappears here. That means it is no longer possible to configure sources from the CLI. Is that intentional?

folkertdev (commented):
quite a few occurences of `peer` are still missed here. (e.g. in the README). So, we can merge them and fix the rest later, but more fixes are needed

docs/guide/configuration.md
rnijveld: It was already removed previously.

rnijveld:
I've pushed another commit to remove a few more references to peer, peers and peer_id in the output (this also results in a few more code changes, but better to be thorough than missing more stuff). There are still a few references to peer left in the changelog, but I'd rather not change old changelog entries. In migrating-chrony there are some references left as well, but those feel like they are related to chrony terminology mostly.

folkertdev (approved):

docs/guide/management-client.md
```diff
@@ -26,10 +26,10 @@ they need to be able to change configuration settings of the daemon dynamically.
 ## Using the management client
 
 The current client exposes 3 different commands:
-- `ntp-ctl peers` displays information on the currently active peer connections
+- `ntp-ctl sources` displays information on the currently active source connections
```
moderation, line 29: `ntp-ctl sources` isn't hooked up and is not an option on the command line
```
target/release/ntp-ctl --help
ntp-ctl - ntp-daemon monitoring

usage: ntp-ctl validate [-c PATH]
       ntp-ctl status [-f FORMAT] [-c PATH]
       ntp-ctl -h | ntp-ctl -v

Options:
  -f, --format=FORMAT                  which format to use for printing statistics [plain, prometheus]
  -c, --config=CONFIG                  which configuration file to read the socket paths from
  ```

PR #1020, opened 2023-09-04
Installation guide

folkertdev:


davidv1992 (commented):
Mostly looks good, few minor things but I think I will pick that up:
 - Think we should say a little more on how to use the linux packages.
 - When installing from source we should say something about placing the config in the right place.

PR #1021, opened 2023-09-04
always require permissions when creating a unix socket

### LINKED ISSUE #517: revise default permissions of the observe socket

folkertdev:
- removes configure socket config
- the exposed api for a socket now always requires explicit permissions


folkertdev:
```rust
const fn default_observation_permissions() -> u32 {
    0o666
}
```

we decided that this is the default that we want

rnijveld (approved):

PR #1050, opened 2023-09-12
Flatten observability config to allow deny_unknown_fields to work

rnijveld:
Using the serde `flatten` attribute in combination with `deny_unknown_fields` [does not work](https://serde.rs/field-attrs.html#flatten), so we flatten the struct in code instead.


davidv1992 (approved):

davidv1992:
Needs a conflict resolution.

PR #1067, opened 2023-09-14
Added remote ip to observability.

### LINKED ISSUE #1066: Have a way to identify specific servers in ntp-ctl and prometheus output
Currently we only show the configured address and the unique source id, but admins probably also need the actual socket address of the remote that we are trying to connect to, in order to properly debug any misconfigured services.

davidv1992:
Closes #1066


rnijveld (changes requested):
Code changes look good, but I think we also require some documentation updates. At the very least the getting started guide.

PR #1116, opened 2023-10-05
Allow configuration without sources.

### LINKED ISSUE #1113: server-only mode (allow having no sources configured)
Currently I have to configure a `[[source]]` section defining the time source. While this is sensible if the ntpd (NTP client) is intented to keep system time correct.

In my usecase I rather use the ntdp server for my development machines on a different non-public subnet. This is just for serving the time in case RTC was disconnected from backup voltage.

I workaround this by using myself as a peer with this config snippet:
```
[[source]]
mode = "server"
address = "localhost"
```
But this spams the log with info messages: `INFO ntp_proto::peer: Peer rejected due to invalid stratum. The stratum of a peer must be lower than the own stratum peer_stratum=8 own_stratum=8` which is correct by itself, but to be expect in my case.

Is there a chance to implement server-mode only?

davidv1992:
Closes #1113

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,12 @@
 # Changelog
 
+## [unreleased]
+
+### Changed
+- The sources section can be left out of the configuration now.
+- When no sources are configured, the daemon will merely state it won't change 
+  system time, rather than warn
+
 ## [1.0.0] - 2023-10-05
 
 ### Added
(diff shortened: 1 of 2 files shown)


tq-steina:
Nice, thanks. This does the trick and also shows only an informational message.

cikzh (approved):

PR #1125, opened 2023-10-09
Ntpv5 packet parsing

tdittr:


ntp-proto/src/packet/v5.rs
```diff
@@ -87,17 +87,10 @@ impl NtpFlags {
...
-            flags |= 0x01;
-        }
-
-        if self.interleaved_mode {
-            flags |= 0x02;
-        }
-
-        flags.to_be_bytes()
+        [0x00, flags]
     }
```
tdittr, line 94: This does work, but I think the `if` version is more explicit. And I'm pretty sure the compiler will optimize both roughly the same.

ntp-proto/src/packet/extensionfields.rs
```diff
@@ -375,6 +418,28 @@ impl<'a> ExtensionField<'a> {
...
+    #[cfg(feature = "ntpv5")]
+    fn decode_draft_identification(
+        message: &'a [u8],
+    ) -> Result<Self, ParsingError<std::convert::Infallible>> {
+        let msg = message[..].into();
+
+        let new_msg = match msg {
+            Cow::Borrowed(slice) => std::str::from_utf8(slice).ok().map(Cow::Borrowed),
+            Cow::Owned(vec) => String::from_utf8(vec).ok().map(Cow::Owned),
+        };
```
folkertdev, line 430: the into will always choose `Borrowed` here (right?). So handling the full Cow confuses me

ntp-proto/src/packet/extensionfields.rs
```diff
@@ -806,6 +873,32 @@ mod tests {
...
 
+    #[cfg(feature = "ntpv5")]
+    #[test]
+    fn draft_identification() {
+        let test_id = "draft-ietf-ntp-ntpv5-00\0";
+        let len = u16::try_from(4 + test_id.len()).unwrap();
+        let mut data = vec![];
+        data.extend_from_slice(&[0xF5, 0xFF]);
+        data.extend_from_slice(&len.to_be_bytes());
+        data.extend_from_slice(test_id.as_bytes());
```
folkertdev, line 884: total nit: using `.extend` would work just as well here. A bit shorter and more common

ntp-proto/src/packet/mod.rs
```diff
@@ -354,6 +359,46 @@ impl<'a> NtpPacket<'a> {
                     }
                 }
             }
+            #[cfg(feature = "ntpv5")]
+            5 => {
+                let (header, header_size) =
+                    v5::NtpHeaderV5::deserialize(data).map_err(|e| e.generalize())?;
+
+                let contruct_packet = |remaining_bytes: &'a [u8], efdata| {
```
folkertdev, line 367: typo: "construct"

ntp-proto/src/packet/mod.rs
```diff
@@ -1070,12 +1180,16 @@ mod tests {
...
         let packet = b"\x34\x02\x06\xe9\x00\x00\x02\x36\x00\x00\x03\xb7\xc0\x35\x67\x6c\xe5\xf6\x61\xfd\x6f\x16\x5f\x03\xe5\xf6\x63\xa8\x76\x19\xef\x40\xe5\xf6\x63\xa8\x79\x8c\x65\x81\xe5\xf6\x63\xa8\x79\x8e\xae\x2b";
         assert!(NtpPacket::deserialize(packet, &NoCipher).is_err());
         let packet = b"\x3B\x02\x06\xe9\x00\x00\x02\x36\x00\x00\x03\xb7\xc0\x35\x67\x6c\xe5\xf6\x61\xfd\x6f\x16\x5f\x03\xe5\xf6\x63\xa8\x76\x19\xef\x40\xe5\xf6\x63\xa8\x79\x8c\x65\x81\xe5\xf6\x63\xa8\x79\x8e\xae\x2b";
         assert!(NtpPacket::deserialize(packet, &NoCipher).is_err());
+
+        #[cfg(not(feature = "ntpv5"))]
+        {
+            let packet = b"\x2C\x02\x06\xe9\x00\x00\x02\x36\x00\x00\x03\xb7\xc0\x35\x67\x6c\xe5\xf6\x61\xfd\x6f\x16\x5f\x03\xe5\xf6\x63\xa8\x76\x19\xef\x40\xe5\xf6\x63\xa8\x79\x8c\x65\x81\xe5\xf6\x63\xa8\x79\x8e\xae\x2b";
+            assert!(NtpPacket::deserialize(packet, &NoCipher).is_err());
+        }
```
folkertdev, line 1192: why does ntpv5 change this test?

ntp-proto/src/packet/v5.rs
```diff
@@ -0,0 +1,409 @@
...
+
+impl NtpFlags {
+    fn from_bits(bits: [u8; 2]) -> Result<Self, ParsingError<std::convert::Infallible>> {
+        if bits[0] != 0x00 || bits[1] & 0xFC != 0 {
+            return Err(ParsingError::InvalidFlags);
+        }
+
+        Ok(Self {
+            unknown_leap: bits[1] & 0x01 != 0,
+            interleaved_mode: bits[1] & 0x02 != 0,
```
folkertdev, line 85: random suggestion: to me bitwise operations with binary literals make more sense. Hex always needs some mental conversion

folkertdev (commented):
cool stuff!

ntp-proto/src/packet/v5.rs
folkertdev: it doesn't, weirdly 

https://godbolt.org/z/x19779b87

I'm confused by that choice. I still think the version with the if is clearer.

ntp-proto/src/packet/mod.rs
tdittr: The test actually checks that a packet with NTP version 5 does produce an error. I guess we could instead of disabling it just test that it returns `ok` with the `ntpv5` feature.

ntp-proto/src/packet/v5.rs
tdittr: I would agree, but the spec lists these specifically as hex. So I am not sure if we should go more for matching the spec or be easier to see, or maybe changing the spec to binary as well?

ntp-proto/src/packet/v5.rs
tdittr: Yeah that is weird. With the full function it gets a bit better but definitely not the same: https://godbolt.org/z/3dcs8d3T1
I still think we should use the `if` for better readability. I will ask @marlonbaeten later what he thinks about it.

ntp-proto/src/packet/extensionfields.rs
tdittr: Done!

ntp-proto/src/packet/mod.rs
tdittr: Fixed

ntp-proto/src/packet/mod.rs
tdittr: I added a comment to why it is disables. It can not test for `Ok()` with the feature because it is not a valid V5 packet... So I guess this is the easiest way forward, as all other tests only check for `.is_err()` and succesful parsing is tested at another point.

ntp-proto/src/packet/v5.rs
folkertdev: yeah, apparently C has binary literals since C99. I'd appreciate that change (in the spec too)

ntp-proto/src/packet/extension_fields.rs
```diff
@@ -375,6 +443,20 @@ impl<'a> ExtensionField<'a> {
...
 
+    #[cfg(feature = "ntpv5")]
+    fn decode_draft_identification(
+        message: &'a [u8],
+    ) -> Result<Self, ParsingError<std::convert::Infallible>> {
+        let err = Err(super::v5::V5Error::InvalidDraftIdentification.into());
+        let di = match core::str::from_utf8(message) {
+            Err(_) => return err,
+            Ok(di) if !di.is_ascii() => return err,
+            Ok(di) => di,
```
folkertdev, line 454: make one `Ok(di) if id.is_ascii()` branch, and a wildcard

ntp-proto/src/packet/extension_fields.rs
```diff
@@ -798,35 +924,102 @@ mod tests {
...
         assert_eq!(
             w,
             &[0, 42, 0, 20, 0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15]
         );
     }
 
+    #[cfg(feature = "ntpv5")]
+    #[test]
+    fn draft_identification() {
+        let test_id = "draft-ietf-ntp-ntpv5-00\0";
```
folkertdev, line 954: lift out to a comment

ntp-proto/src/packet/extension_fields.rs
```diff
@@ -798,35 +924,102 @@ mod tests {
...
+
+        let raw = RawExtensionField::deserialize(&data, 0, ExtensionHeaderVersion::V5).unwrap();
+        let ef = ExtensionField::decode(raw).unwrap();
+
+        match ef {
+            ExtensionField::DraftIdentification(ref data) => {
+                assert_eq!(data, test_id);
+            }
+            _ => panic!("Unexpected extensionfield {ef:?}"),
+        }
```
folkertdev, line 969: try let-else?

ntp-proto/src/packet/mod.rs
```diff
@@ -666,69 +764,95 @@ impl<'a> NtpPacket<'a> {
...
         }
     }
 
     pub fn is_kiss(&self) -> bool {
         match self.header {
             NtpHeader::V3(header) => header.stratum == 0,
             NtpHeader::V4(header) => header.stratum == 0,
+            #[cfg(feature = "ntpv5")]
+            // TODO NTPv5 does not have kiss codes yet
+            NtpHeader::V5(_header) => false,
```
folkertdev, line 855: make this a `todo!`

ntp-proto/src/packet/mod.rs
```diff
@@ -810,69 +939,98 @@ impl<'a> NtpPacket<'a> {
...
+            NtpHeader::V5(ref mut header) => header.stratum = stratum,
         }
     }
 
     pub fn set_reference_id(&mut self, reference_id: ReferenceId) {
         match &mut self.header {
             NtpHeader::V3(ref mut header) => header.reference_id = reference_id,
             NtpHeader::V4(ref mut header) => header.reference_id = reference_id,
+            #[cfg(feature = "ntpv5")]
+            NtpHeader::V5(_header) => todo!("NTPv5 does not have refernce IDs"),
```
folkertdev, line 1015: typo

tdittr:
Replaced by rebased version in #1141

PR #1151, opened 2023-10-24
Refactor serve_packet into IO and non IO parts

tdittr:
This decouples the IO during packet handling from the actual logic. This allows easier writing of tests,

This PR also flattens a lot of nested matches to make them more readable.

### DIFF
--- a/ntpd/src/daemon/server.rs
+++ b/ntpd/src/daemon/server.rs
@@ -39,6 +39,34 @@ pub struct ServerStats {
     pub nts_nak_packets: Counter,
 }
 
+impl ServerStats {
+    fn update_from(&self, res: &AcceptResult<'_>) {
+        use AcceptResult::{Accept, CryptoNak, Deny, Ignore, RateLimit};
+
+        self.received_packets.inc();
+
+        match res {
+            Accept { .. } => self.accepted_packets.inc(),
+            Ignore => self.ignored_packets.inc(),
+            Deny { .. } => self.denied_packets.inc(),
+            RateLimit { .. } => self.rate_limited_packets.inc(),
+            CryptoNak { .. } => self.nts_nak_packets.inc(),
+        };
+
+        if res.is_nts() {
+            self.nts_received_packets.inc();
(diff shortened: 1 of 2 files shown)


ntpd/src/daemon/server.rs
```diff
@@ -196,321 +283,262 @@ impl<C: 'static + NtpClock + Send> ServerTask<C> {
...
-                };
+        }
+    }
 
-                if let Err(serialize_err) = serialize_result {
-                    self.stats.response_send_errors.inc();
-                    warn!(error=?serialize_err, "Could not serialize response");
-                    return true;
-                }
+    // TODO: this instrument breaks coverage collection for some reason...
```
davidv1992, line 367: Remaining todo

ntpd/src/daemon/server.rs
```diff
@@ -196,321 +283,262 @@ impl<C: 'static + NtpClock + Send> ServerTask<C> {
...
-                    return true;
                 }
+            }
+            Ok((length, peer_addr, opt_timestamp)) => {
+                let Some(request_buf) = buf.get(..length) else {
+                    warn!("length from socket is out of bounds. This is a bug!");
+                    return SocketConnection::Reconnect;
+                };
+                let mut response_buf = [0; MAX_PACKET_SIZE];
+                let response_buf = response_buf.as_mut_slice();
```
davidv1992, line 345: We should perhaps limit the response buffer size here already.

davidv1992 (commented):
Two minor points, but otherwise looking very good.

ntpd/src/daemon/server.rs
tdittr: Fixed!

ntpd/src/daemon/server.rs
tdittr: Great idea, changed it and removed the old check at the end. So now if this ever happens it should just show up as a serialization error.

davidv1992 (approved):

PR #1153, opened 2023-10-25
Add handling of NTPv5 to the server

tdittr:
- Fix packet deserialization
- Fix response serialization
- Append and check draft identification
- Implement upgrade protocol

### DIFF
--- a/ntp-proto/src/packet/mod.rs
+++ b/ntp-proto/src/packet/mod.rs
@@ -505,6 +505,18 @@ impl<'a> NtpPacket<'a> {
         )
     }
 
+    // pub fn poll_message_v5(poll_interval: PollInterval) -> (Self, RequestIdentifier) {
+    //     let (header, id) = NtpHeaderV5::poll_message(poll_interval);
+    //     (
+    //         NtpPacket {
+    //             header: NtpHeader::V5(header),
+    //             efdata: Default::default(),
+    //             mac: None,
+    //         },
+    //         id,
+    //     )
+    // }
+
     pub fn timestamp_response<C: NtpClock>(
         system: &SystemSnapshot,
         input: Self,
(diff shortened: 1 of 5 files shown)


ntp-proto/src/packet/mod.rs
```diff
@@ -235,6 +235,12 @@ impl NtpHeaderV3V4 {
         recv_timestamp: NtpTimestamp,
         clock: &C,
     ) -> Self {
+        let reference_timestamp = match input.reference_timestamp {
+            #[cfg(feature = "ntpv5")]
+            v5::UPGRADE_TIMESTAMP => v5::UPGRADE_TIMESTAMP,
+            _ => Default::default(),
+        };
```
davidv1992, line 242: I think we should only response with the upgrade timestamp if there is also a draft id extension field for the draft we implement. Otherwise we will confuse future implementations saying that we support ntpv5 when we don't support theirs.

tdittr: I can add that, but it is not mentioned in the draft, so clients would usually not send a draft id extension by default I guess...

davidv1992: I'll write a proposal to add this more explicitly to the next draft. For right now, since we are the first implementation (other than a few demo implementations that don't do time sync) I think we can just require this one-sided.

tdittr: Done!

ntp-proto/src/packet/v5/mod.rs
```diff
@@ -118,9 +121,18 @@ impl NtpServerCookie {
 pub struct NtpClientCookie(pub [u8; 8]);
 
 impl NtpClientCookie {
+    fn new_random() -> Self {
+        // TODO does this match entropy handling of the rest of the system?
+        Self(random())
```
folkertdev, line 126: it is kind of unfortunate for testing (potentially for reproducing fuzzing errors too)

marlonbaeten (approved):

rnijveld (approved):

PR #1156, opened 2023-10-27
Add a 30 minute timeout to FreeBSD tests

tdittr:
This detects boot loops earlier and thus wastes less resources. Also it makes noticing those loops faster (30 minutes instead of 6 hours)


rnijveld:
I'm working on a refactor of the github workflows that already includes this!

PR #1159, opened 2023-10-30
Fix minimum MAC size this fixes a deserialization error

tdittr:
Before the last extension was always ignored if it was not at least 32 bytes without padding.


marlonbaeten (approved):

rnijveld (approved):

rnijveld:
@tdittr You can merge this after a rebase :)

PR #1169, opened 2023-10-31
V5 padding extension

marlonbaeten:


ntp-proto/src/packet/extension_fields.rs
```diff
@@ -113,9 +125,27 @@ impl<'a> ExtensionField<'a> {
             InvalidNtsEncryptedField => InvalidNtsEncryptedField,
             #[cfg(feature = "ntpv5")]
             DraftIdentification(data) => DraftIdentification(Cow::Owned(data.into_owned())),
+            #[cfg(feature = "ntpv5")]
+            Padding(len) => Padding(len),
         }
     }
 
+    pub fn byte_length(&self) -> usize {
```
tdittr, line 133:
```suggestion
    pub fn wire_length(&self) -> usize {
```

ntp-proto/src/packet/mod.rs
```diff
@@ -440,6 +440,16 @@ impl<'a> NtpPacket<'a> {
         Ok(buffer)
     }
 
+    pub fn byte_length(&self) -> usize {
```
tdittr, line 443:
```suggestion
    pub fn wire_length(&self) -> usize {
```

ntp-proto/src/packet/mod.rs
```diff
@@ -440,6 +440,16 @@ impl<'a> NtpPacket<'a> {
...
+        NtpHeaderV3V4::WIRE_LENGTH
+            + self
+                .efdata
+                .untrusted
+                .iter()
+                .map(|ev| ev.byte_length())
```
tdittr, line 449:
```suggestion
                .map(|ev| ev.wire_length())
```

tdittr (commented):
Not sure if this is the way to go, as callinf `.append_padding()` before serializing could break stuff. I think we need to further discuss this and see what is a good solution.

davidv1992:
Honestly, I think a better approach is to not add this as an actual extension field before serialization, but rather to generate this as needed during serialization.

ntp-proto/src/packet/mod.rs
```diff
@@ -467,6 +470,22 @@ impl<'a> NtpPacket<'a> {
             mac.serialize(w)?;
         }
 
+        #[cfg(feature = "ntpv5")]
+        if let NtpHeader::V5(v5::NtpHeaderV5 {
+            desired_size: Some(desired_size),
```
tdittr, line 475: Having this in the header feels weird, why not pass the length to the call, or specify that the buffer we write to should be the same size as the request packet?

tdittr (changes requested):

tdittr (approved):
Looks good! Much cleaner

davidv1992 (approved):

PR #1185, opened 2023-11-09
Rename the COPYRIGHT file to COPYING

rnijveld:
The COPYRIGHT file is not a standard file location whereas the COPYING filename is according to the openssf best practices standard


cikzh (commented):
According to: https://github.com/ossf/scorecard/blob/main/docs/checks.md#license COPYRIGHT as a file name seems fine

rnijveld:
Indeed, but the OpenSSF Best Practices do not suggest it as a name, which would prevent us from getting a passing grade there.

cikzh (approved):

PR #1195, opened 2023-11-13
en/decoding for nts pool records

folkertdev:


ntp-proto/src/nts_record.rs
```diff
@@ -43,6 +43,14 @@ impl NtsRecord {
             NtsRecord::NewCookie { .. } => 5,
             NtsRecord::Server { .. } => 6,
             NtsRecord::Port { .. } => 7,
+            #[cfg(feature = "nts-pool")]
+            NtsRecord::KeepAlive { .. } => 0x4000,
+            #[cfg(feature = "nts-pool")]
+            NtsRecord::SupportedProtocolList { .. } => 0x4001,
```
davidv1992, line 49: My Mistake, this name is not appropriate, should be SupportedAlgorithmList, since this says something about which aead algorithms are allowed, not what next protocol is allowed.

I think it is important to correct on this now because I do think that once NTPv5 becomes a thing, requesting a servers supported next protocols will also become a thing.

davidv1992 (commented):
Looks good, one minor naming change because I confused some naming in my drafts of the spec.

ntp-proto/src/nts_record.rs
folkertdev: right, this was already kind of confusing because we also have a concept of `ProtocolId` (the ntp version). I'll change it

davidv1992 (approved):

PR #1208, opened 2023-11-16
Run nts key provider on dedicated thread

### LINKED ISSUE #1198: Remove blocking code in async function
In https://github.com/pendulum-project/ntpd-rs/blob/main/ntpd/src/daemon/nts_key_provider.rs#L18 we blockingly load file metadata within an async context, which could easily be replaced by non-blocking code that prevents any blocking in async code issues

mickvangelderen:
Loading the initial keyset is done using spawn_blocking since it involves file I/O. The watch loop is ran on a dedicated thread.

Fixes #1198

### DIFF
--- a/ntpd/src/daemon/mod.rs
+++ b/ntpd/src/daemon/mod.rs
@@ -91,7 +91,9 @@ async fn run(options: NtpDaemonOptions) -> Result<(), Box<dyn Error>> {
     config.check();
 
     // we always generate the keyset (even if NTS is not used)
-    let keyset = nts_key_provider::spawn(config.keyset).await;
+    let keyset = tokio::task::spawn_blocking(move || nts_key_provider::spawn(config.keyset))
+        .await
+        .expect("nts_key_provider::spawn should not panic");
 
     #[cfg(feature = "hardware-timestamping")]
     let clock_config = config.clock;
--- a/ntpd/src/daemon/nts_key_provider.rs
+++ b/ntpd/src/daemon/nts_key_provider.rs
@@ -1,7 +1,9 @@
 use std::{
     fs::{File, OpenOptions},
     os::unix::prelude::{OpenOptionsExt, PermissionsExt},
+    path::Path,
     sync::Arc,
+    time::Duration,
 };
 
 use ntp_proto::{KeySet, KeySetProvider};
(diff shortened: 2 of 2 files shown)


rnijveld:
I think we can simplify this by just using the tokio alternative (i.e. tokio::fs::metadata instead of std::fs::metadata) for all the file manipulation stuff, using a separate blocking thread feels a little heavy for me, any thoughts?

mickvangelderen:
> I think we can simplify this by just using the tokio alternative (i.e. tokio::fs::metadata instead of std::fs::metadata) for all the file manipulation stuff

I think doing so makes sense if:

- the `tokio::fs` implementation [does not just spawn a thread for us](https://stackoverflow.com/questions/70599317/is-there-any-point-in-async-file-io)
- we need streaming `AsyncRead` and the code becomes simpler

> using a separate blocking thread feels a little heavy for me

This is what `spawn_blocking` does and what the `tokio::fs` implementation probably also does. I think you need cool new tech like [tokio-uring](`https://github.com/tokio-rs/tokio-uring`) in order not to have to spawn threads to hide the blocking. 




What is concerning me more is that this code is not covered by any tests. Does that need to change?

rnijveld:
Thanks for your response! My comment was more aimed at the large change in code structure, given that the entire spawn function is now no longer async, although my wording probably wasn't great in that regard.

I think our initial version where spawn is an async function is better as it helps with legibility of the code if we don't have to context switch between blocking and async contexts too much, which is solved easily by replacing `std::fs::metadata` and `std::fs::OpenOptions` with their tokio alternatives. Those also keep implementation details hidden and allow us to focus our own program logic as much as possible, and that way we retain the capability of using other async operations within the spawn function in the future.

I agree that this code could use a little testing though! If you want to help with that it is very much appreciated!

mickvangelderen:
> Thanks for your response! My comment was more aimed at the large change in code structure

I've deduplicated the construction of a new keyset and simplified some things and added a more accurate warning when the keyset file path is configured but the file does not exist. 

> , given that the entire spawn function is now no longer async. 

This is a good change in my opinion since `KeySetProvider::store` and `KeySetProvider::load` are sync (implemented using `std::io::Read` instead of `tokio::io::AsyncRead` and nothing in the implementation required async. 

> I think our initial version where spawn is an async function is better as it helps with legibility of the code if we don't have to context switch between blocking and async contexts too much, which is solved easily by replacing `std::fs::metadata` and `std::fs::OpenOptions` with their tokio alternatives. 

Personally, and this is subjective, I like it when we can prevent async from leaking everywhere. This is not my project though. Are there any relevant guidelines for this project about when to use async vs sync?

> Those also keep implementation details hidden and allow us to focus our own program logic as much as possible, and that way we retain the capability of using other async operations within the spawn function in the future.

Personally, I like to design for what you need, not what you might need. 

Anyway I decided to try to use `async` everywhere, but that also mean introducing tokio as a dependency for  

> 
> I agree that this code could use a little testing though! If you want to help with that it is very much appreciated!

davidv1992:
Having taken a look, my opinion currently is that this is overkill for fixing the blocking code in an async context. Most of that is already addressed in the current code through a number of spawn_blockings in specific places, the only additional blocking call this would cover is the metadata one, which has a far simpler fix and on top of that doesn't seem to cause any real issues since it only happens on startup.

That said, the diagnosis that this code could benefit significantly from a refactoring pass is I think correct. However, I am not sure I find the particular approach taken here an improvement. In particular the passing around of closures creates a bit of disconnect between what run actually does and its code. I think if we want to make progress in making this code nicer a better approach would be to take a more holistic view, also removing some of the risk of blocking calls in ntp-proto as Ruben suggest. However, figuring out what the best interface there is is going to be a bit tricky and I don't immediately have a concrete direction there.

davidv1992:
After some internal discussion, we have decided to close this for now. Feel free to reopen it or open a new PR if you want to do the small fix.

If you want to take on the larger refactor work we would prefer it if you would discuss the direction of it with us beforehand, so we increase the chances of success. However, note that given our current priorities we may be quite slow to respond in such a discussion as fixing the style problems here is quite low on our list.

mickvangelderen:
@davidv1992 that's entirely reasonable. Are there any issues I could pick up that *are* a priority?

PR #1209, opened 2023-11-16
Implemented status messages for ntpv5.

davidv1992:
Note: this is based on the current proposal from us, not yet part of the draft.

### DIFF
--- a/ntp-proto/src/packet/mod.rs
+++ b/ntp-proto/src/packet/mod.rs
@@ -858,7 +858,25 @@ impl<'a> NtpPacket<'a> {
                 mac: None,
             },
             #[cfg(feature = "ntpv5")]
-            NtpHeader::V5(_header) => todo!("NTPv5 does not have KISS codes yet"),
+            NtpHeader::V5(header) => NtpPacket {
+                header: NtpHeader::V5(v5::NtpHeaderV5::rate_limit_response(header)),
+                efdata: ExtensionFieldData {
+                    authenticated: vec![],
+                    encrypted: vec![],
+                    // Ignore encrypted so as not to accidentaly leak anything
+                    untrusted: packet_from_client
+                        .efdata
+                        .untrusted
+                        .into_iter()
+                        .chain(packet_from_client.efdata.authenticated)
+                        .filter(|ef| matches!(ef, ExtensionField::UniqueIdentifier(_)))
+                        .chain(std::iter::once(ExtensionField::DraftIdentification(
+                            Cow::Borrowed(v5::DRAFT_VERSION),
+                        )))
+                        .collect(),
+                },
+                mac: None,
+            },
         }
     }
 
@@ -880,7 +898,23 @@ impl<'a> NtpPacket<'a> {
                 mac: None,
             },
             #[cfg(feature = "ntpv5")]
-            NtpHeader::V5(_header) => todo!("No NTS support yet"),
+            NtpHeader::V5(header) => NtpPacket {
+                header: NtpHeader::V5(v5::NtpHeaderV5::rate_limit_response(header)),
+                efdata: ExtensionFieldData {
+                    authenticated: packet_from_client
+                        .efdata
+                        .authenticated
+                        .into_iter()
+                        .filter(|ef| matches!(ef, ExtensionField::UniqueIdentifier(_)))
+                        .chain(std::iter::once(ExtensionField::DraftIdentification(
+                            Cow::Borrowed(v5::DRAFT_VERSION),
+                        )))
+                        .collect(),
+                    encrypted: vec![],
+                    untrusted: vec![],
+                },
+                mac: None,
+            },
         }
     }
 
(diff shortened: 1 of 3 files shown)


ntp-proto/src/packet/mod.rs
```diff
@@ -1074,30 +1159,42 @@ impl<'a> NtpPacket<'a> {
...
+            }
+        }
+    }
+
     pub fn is_kiss(&self) -> bool {
         match self.header {
             NtpHeader::V3(header) => header.stratum == 0,
             NtpHeader::V4(header) => header.stratum == 0,
             #[cfg(feature = "ntpv5")]
             // TODO NTPv5 does not have Kiss codes so we pretend everything is always fine
```
rnijveld, line 1179: This todo no longer seems to apply?

ntp-proto/src/packet/mod.rs
```diff
@@ -1074,30 +1159,42 @@ impl<'a> NtpPacket<'a> {
         }
     }
 
+    pub fn kiss_code(&self) -> ReferenceId {
+        match self.header {
+            NtpHeader::V3(header) => header.reference_id,
+            NtpHeader::V4(header) => header.reference_id,
+            #[cfg(feature = "ntpv5")]
+            // Kiss code in ntpv5 is the first four bytes of the server vookie
```
rnijveld, line 1167: Typo: `vookie` -> `cookie`

rnijveld (changes requested):
Two small things, but other than that LGTM!

ntp-proto/src/packet/mod.rs
```diff
@@ -858,7 +858,25 @@ impl<'a> NtpPacket<'a> {
...
+                    untrusted: packet_from_client
+                        .efdata
+                        .untrusted
+                        .into_iter()
+                        .chain(packet_from_client.efdata.authenticated)
+                        .filter(|ef| matches!(ef, ExtensionField::UniqueIdentifier(_)))
+                        .chain(std::iter::once(ExtensionField::DraftIdentification(
+                            Cow::Borrowed(v5::DRAFT_VERSION),
+                        )))
+                        .collect(),
```
tdittr, line 876: This section now gets repeated a lot. Maybe move it into a function?

ntp-proto/src/packet/mod.rs
```diff
@@ -1074,30 +1159,42 @@ impl<'a> NtpPacket<'a> {
         }
     }
 
+    pub fn kiss_code(&self) -> ReferenceId {
```
tdittr, line 1162: Shouldnt this return an `Option<ReferenceId>` or something like `Option<KissCode>`?

ntp-proto/src/packet/mod.rs
```diff
@@ -1074,30 +1159,42 @@ impl<'a> NtpPacket<'a> {
...
+            NtpHeader::V4(header) => header.reference_id,
+            #[cfg(feature = "ntpv5")]
+            // Kiss code in ntpv5 is the first four bytes of the server vookie
+            NtpHeader::V5(header) => {
+                ReferenceId::from_bytes(header.server_cookie.0[..4].try_into().unwrap())
+            }
+        }
+    }
+
     pub fn is_kiss(&self) -> bool {
```
tdittr, line 1174: If `Self::kiss_code` returns an option this could be removed or made private

ntp-proto/src/packet/v5/mod.rs
```diff
@@ -210,6 +218,49 @@ impl NtpHeaderV5 {
...
+            flags: NtpFlags {
+                unknown_leap: false,
+                interleaved_mode: false,
+                status_message: true,
+            },
+            server_cookie: NtpServerCookie([b'R', b'A', b'T', b'E', 0, 0, 0, 0]),
+            client_cookie: packet_from_client.client_cookie,
+            poll: packet_from_client.poll.force_inc(),
+            ..Self::new()
+        }
```
tdittr, line 233: This is repeated a lot, maybe move the common parts into a function like `Self::kiss_response` and call it instead of `Self::new`

tdittr (changes requested):
Looks good, is just not DRY in some places.

ntp-proto/src/packet/mod.rs
davidv1992: This may be a good idea, but I suggest we do that in a separate PR for both the NTPv4 and v5 code in one go.

ntp-proto/src/packet/mod.rs
davidv1992: Maybe, though I chose to just make this a private function for now.

ntp-proto/src/packet/mod.rs
davidv1992: see above

ntp-proto/src/packet/mod.rs
davidv1992: removed the todo

ntp-proto/src/packet/mod.rs
davidv1992: fixed

ntp-proto/src/packet/v5/mod.rs
davidv1992: done

tdittr (approved):
Looks good now!

rnijveld (approved):

PR #1223, opened 2023-11-22
Fix wrong interpretation of root delay and dispersion to match spec

### LINKED ISSUE #1222: NTPv5 still uses old interpretation of Root Delay and Root Dispersion
NTPv5 changed how to interpret the bits used in Root Delay and Root Dispersion fields. We currently use the definition from NTPv4 here.

NTPv4 uses 16 integral and 16 fractional bits.

NTPv5 uses 4 integral and 28 fractional bits.

tdittr:
Fixes #1222

### DIFF
--- a/ntp-proto/src/time_types.rs
+++ b/ntp-proto/src/time_types.rs
@@ -235,6 +235,27 @@ impl NtpDuration {
         .to_be_bytes()
     }
 
+    #[cfg(feature = "ntpv5")]
+    pub(crate) const fn from_bits_time32(bits: [u8; 4]) -> Self {
+        NtpDuration {
+            duration: (u32::from_be_bytes(bits) as i64) << 4,
+        }
+    }
+
+    #[cfg(feature = "ntpv5")]
+    pub(crate) fn to_bits_time32(self) -> [u8; 4] {
+        // serializing negative durations should never happen
+        // and indicates a programming error elsewhere.
+        // as for duration that are too large, saturating is
+        // the safe option.
+        assert!(self.duration >= 0);
+
+        // On overflow we just saturate to the maximum 16s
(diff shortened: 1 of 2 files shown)


ntp-proto/src/time_types.rs
```diff
@@ -235,6 +235,33 @@ impl NtpDuration {
...
+        assert!(self.duration >= 0);
+
+        // Other then for short implementation this might very well overflow sometimes, we just saturate in that case
+        const MAX_TIME32: NtpDuration = NtpDuration {
+            duration: 0x00_00_00_0F_FF_FF_FF_F0,
+        };
+
+        match self.duration > MAX_TIME32.duration {
+            true => u32::MAX,
+            false => ((self.duration & MAX_TIME32.duration) >> 4) as u32,
```
tdittr, line 260: Arguably this could just be
```suggestion
            false => (self.duration >> 4) as u32,
```
as the `as u32` already strips the high bits.

ntp-proto/src/time_types.rs
```diff
@@ -235,6 +235,33 @@ impl NtpDuration {
...
+        // as for duration that are too large, saturating is
+        // the safe option.
+        assert!(self.duration >= 0);
+
+        // Other then for short implementation this might very well overflow sometimes, we just saturate in that case
+        const MAX_TIME32: NtpDuration = NtpDuration {
+            duration: 0x00_00_00_0F_FF_FF_FF_F0,
+        };
+
+        match self.duration > MAX_TIME32.duration {
```
tdittr, line 258: With the change below this could also be simplified to:
```suggestion
        match self.duration > 0x0F_FF_FF_FF_F0 {
```

davidv1992 (approved):

PR #1226, opened 2023-11-23
Add TLS client certificate acceptance to the NTSKE server.

squell:


ntp-proto/src/nts_record.rs
```diff
@@ -2321,7 +2321,10 @@ mod test {
         let mut serverconfig = rustls::ServerConfig::builder()
             .with_safe_defaults()
             .with_client_cert_verifier(Arc::new(
-                rustls::server::AllowAnyAnonymousOrAuthenticatedClient::new(root_store.clone()),
+                #[cfg(not(feature = "nts-pool"))]
```
squell, line 2324: Note: this is in test-code; it feels better (in the default mode for ntpd-rs) to run tests using a rustls-provided `ClientCertVerifier` instead of the home-brew one.

ntp-proto/src/tls_utils.rs
```diff
@@ -0,0 +1,22 @@
+pub struct AllowAnyAnonymousOrCertificateBearingClient;
```
folkertdev, line 1: why not `AllowAnyClient`?

ntp-proto/Cargo.toml
```diff
@@ -17,7 +17,7 @@ __internal-fuzz = ["arbitrary", "__internal-api"]
 __internal-test = ["__internal-api"]
 __internal-api = []
 ntpv5 = []
-nts-pool = []
+nts-pool = [ "rustls/dangerous_configuration" ]
```
folkertdev, line 20: exciting

ntp-proto/src/tls_utils.rs
folkertdev: in other words, what makes this different from `NoClientAuth`. I'd like some comments for why this is needed/what the goal is.

squell: Check on the "needs more comments part".

- `AllowAnyClient` feels too imprecise; the RusTLS built-in verifier that is closest is called `AllowAnyAnonymousOrAuthenticatedClient`, which also sounds like "allow any client"; except that it will refuse a connection if a certificate is presented that is not validated against a trust root. The intent of this verifier is that it will accept any anonymous connection (i.e. no client certificate) as well as those with a well-formed certificate (i.e. clients bearing a certificate), which is not the same as being "authenticated". (I'm not wedded to the ridiculously long name `AnyAnonymousOrCertificateBearingClient`, but it has the benefit of being very closely related to the name RusTLS chose).

- `NoClientAuth` is the same as `no_client_auth`, i.e. client certificates don't end up in `peer_certificates()` and that means we can't check whether someone is allowed to send `FixedKeyRequests`.

ntp-proto/Cargo.toml
squell: 😎

ntp-proto/src/tls_utils.rs
davidv1992: I approve of this name, consistent with rustls which i think is the most important here.

davidv1992 (dismissed):

PR #1228, opened 2023-11-24
Send error records back to the client

folkertdev:
we neglected to send error records back to the client when the server runs into a problem, either because an unknown record with the critical bit is decoded, or when the client sends insufficient input.

I think this lacks more high-level testing with a faulty client. @tdittr maybe your testing project could focus on this using `tokio_rustls` to just send incomplete byte streams, send all records in one in separate write calls, etc. Without `tokio_rustls` or similar this is very painful to test.


ntp-proto/src/nts_record.rs
```diff
@@ -1288,7 +1288,13 @@ impl KeyExchangeServerDecoder {
                 Continue(state)
             }
 
-            Unknown { .. } => Continue(state),
+            Unknown { critical, .. } => {
+                if critical {
+                    Break(Err(KeyExchangeError::UnrecognizedCriticalRecord))
+                } else {
+                    Continue(state)
+                }
```
folkertdev, line 1296: original cause of this PR. we test now also that an unknown record with a critical bit sends a record to that effect to the client.

ntp-proto/src/nts_record.rs
```diff
@@ -1323,119 +1335,118 @@ impl KeyExchangeServer {
...
+            InternalServerError | Io(_) => INTERNAL_SERVER_ERROR,
+            UnknownErrorCode(_)
+            | NoValidProtocol
+            | NoValidAlgorithm
+            | InvalidFixedKeyLength
+            | NoCookies
+            | Tls(_)
+            | Certificate(_)
+            | DnsName(_)
+            | IncompleteResponse => BAD_REQUEST,
```
folkertdev, line 1372: this seemed sensible to us, but have a look if those choices are reasonable

ntp-proto/src/nts_record.rs
```diff
@@ -1323,119 +1335,118 @@ impl KeyExchangeServer {
...
+
+        if let Err(io) = Self::send_records(tls_connection, &[NtsRecord::Error { errorcode }]) {
+            tracing::debug!(key_exchange_error = ?error, io_error = ?io, "sending error record failed");
         }
     }
 
-    fn progress_help(mut self) -> ControlFlow<Result<Self, KeyExchangeError>, Self> {
+    pub fn progress(
+        mut self,
+    ) -> ControlFlow<Result<rustls::ServerConnection, KeyExchangeError>, Self> {
```
folkertdev, line 1382: this is a side-by-side sort of change I'm afraid. This is pretty nasty, further refactoring suggestions are welcome

ntp-proto/src/nts_record.rs
```diff
@@ -1323,119 +1339,114 @@ impl KeyExchangeServer {
...
+            | NoValidAlgorithm
+            | InvalidFixedKeyLength
+            | NoCookies
+            | Tls(_)
+            | Certificate(_)
+            | DnsName(_)
+            | IncompleteResponse => NtsRecord::BAD_REQUEST,
+        };
+
+        if let Err(io) = Self::send_records(tls_connection, &[NtsRecord::Error { errorcode }]) {
```
davidv1992, line 1375: Missing end of message

ntp-proto/src/nts_record.rs
```diff
@@ -1323,119 +1339,114 @@ impl KeyExchangeServer {
...
-                            // client is sending more bytes, but we don't expect any more
-                            return ControlFlow::Break(Err(KeyExchangeError::InternalServerError));
+                        ControlFlow::Break(Err(error)) => {
+                            Self::send_error_record(&mut self.tls_connection, &error);
+                            ControlFlow::Break(Err(error))
                         }
+                    },
+                    State::Done => {
+                        // client is sending more bytes, but we don't expect any more
+                        let error = KeyExchangeError::InternalServerError;
```
davidv1992, line 1416: This should be a bad request if we error on this.

davidv1992 (commented):
Mostly looks good, two minor issues

davidv1992 (approved):

PR #1229, opened 2023-11-24
Made the artifact downloader a little bit more flexible

rnijveld:


utils/prepare-artifacts.sh
```diff
@@ -0,0 +1,132 @@
...
+
+if [ "$DOWNLOAD_ARTIFACTS" = true ]; then
+    if [ -z "$GITHUB_API_TOKEN" ]; then
+        echo "ERROR: Missing Github API token"
+    fi
+
+    if [ -z "$RUN_ID" ]; then
+        echo "ERROR: Missing github workflow run id"
+    fi
+    print_help
```
tdittr, line 69: I think this will always print the help and exit if `DOWNLOAD_ARTIFACTS` is `true`

tdittr (changes requested):
I think there is a small error in the parameter validation.

tdittr (approved):
Looks great now!

PR #1230, opened 2023-11-26
Client pool state machine

folkertdev:


ntp-proto/Cargo.toml
```diff
@@ -13,6 +13,7 @@ rust-version.workspace = true
 # See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
 
 [features]
+default = ["nts-pool"]
```
folkertdev, line 16: remove before merge

ntpd/src/daemon/spawn/nts.rs
```diff
@@ -94,6 +94,7 @@ impl NtsSpawner {
                 }
                 Err(e) => {
                     warn!(error = ?e, "error while attempting key exchange");
+                    panic!();
```
folkertdev, line 97: remove before merge

nts-pool-ke/src/lib.rs
```diff
@@ -192,6 +209,208 @@ async fn pool_key_exchange_server(
...
+        .any(|(algorithm_id, _)| *algorithm_id == client_data.algorithm as u16)
+    {
+        unreachable!("the default algorithm must be supported by all clients and servers");
+    }
+
+    // this is inefficient of course, but spec-compliant: the TLS connection is closed when the server
+    // receives a EndOfMessage record, so we have to establish a new one. re-using the TCP
+    // connection runs into issues (seems to leave the server in an invalid state).
+    let server_stream = tokio::net::TcpStream::connect((server_name.as_str(), port)).await?;
+    let server_stream = connector.connect(domain, server_stream).await?;
```
folkertdev, line 254: unfortunate, but I don't think we can easily do better. Fine for a PoC I'd say.

nts-pool-ke/src/lib.rs
```diff
@@ -192,6 +209,208 @@ async fn pool_key_exchange_server(
...
+    let supported_algorithms = supported_algorithms_request(&mut server_stream).await?;
+
+    info!("received supported algorithms from the NTS KE server");
+
+    if !supported_algorithms
+        .iter()
+        .any(|(algorithm_id, _)| *algorithm_id == client_data.algorithm as u16)
+    {
+        unreachable!("the default algorithm must be supported by all clients and servers");
+    }
```
folkertdev, line 248: eh, yeah, what logic do we want to implement here?

nts-pool-ke/src/lib.rs
```diff
@@ -192,6 +209,208 @@ async fn pool_key_exchange_server(
     }
 }
 
+async fn handle_client(
+    client_stream: tokio::net::TcpStream,
+    config: Arc<rustls::ServerConfig>,
+    certificate_authority: Arc<[rustls::Certificate]>,
+) -> Result<(), KeyExchangeError> {
```
folkertdev, line 216: in general here too there is the question of how to report errors back to the NTS client. E.g. if the pool <-> server has an error, that must ultimately be reported to the client. Should that forward the error from the server, or just always return an InternalServerError?

squell: I think this can be made into a seperate issue; right now this is "proof of concept" code.

nts-pool-ke/src/lib.rs
squell: I think this can be  part of the work to implement the `KeepAlive` bit; which should be made into a separate issue (and then this can be a part of that).

squell (commented):
Of course the pool-related stuff should be hidden behind feauture flags.

I think (for maintainability) that it is better if the NTS-POOL-KE related decoders are spun off in their own file (`nts_record_extensions.rs` or `nts_pool_record.rs` or some such?). That would also cut down on the amount of `#[cfg(feature = "nts-pool")]` that is needed.

The changes that are not related to the pool-ke look pretty innocuous to me.

nts-pool-ke/src/lib.rs
folkertdev: filed https://github.com/pendulum-project/ntpd-rs/issues/1246

nts-pool-ke/src/lib.rs
folkertdev: filed https://github.com/pendulum-project/ntpd-rs/issues/1245

ntp-proto/src/nts_record.rs
```diff
@@ -1570,8 +1598,9 @@ impl KeyExchangeServer {
     }
 
     #[cfg(feature = "nts-pool")]
-    fn extract_nts_keys(&self, data: ServerKeyExchangeData) -> Result<NtsKeys, KeyExchangeError> {
-        if let Some(keys) = data.fixed_keys {
+    fn extract_nts_keys(&self, data: &ServerKeyExchangeData) -> Result<NtsKeys, KeyExchangeError> {
+        if let Some(keys) = &data.fixed_keys {
+            // TODO remove "true" here when the connection checking works
```
squell, line 1603: You forgot to remove this comment.

ntp-proto/src/nts_record.rs
```diff
@@ -1504,7 +1522,17 @@ impl KeyExchangeServer {
...
+                            // recursively invoke the progress function. This is very unlikely!
+                            //
+                            // Normally, all records are written with a single write call, and
+                            // received as one unit. Using many write calls does not really make
+                            // sense for a client.
+                            //
+                            // So then, the other reason we could end up here is if the buffer is
+                            // full. But 512 bytes is a lot of space for this interaction, and
+                            // should be sufficient in most cases.
+                            self.progress()
```
squell, line 1535: Two questions:
- What's the advantage over a recursive call vs. using the `ControlFlow::Continue(self)`?
- How would this behave if someone (i.e. an attacker) starts making it very likely that this happens with the objective of making the NTS server crash due to a stack overflow?

squell (approved):

PR #1234, opened 2023-11-27
use tokio_rustls to test our server/client TLS

folkertdev:
this identified some issues that will be fixed in later PRs. But in the interest of keeping PRs small here are the helpers and working tests.

### DIFF
--- a/ntpd/src/daemon/keyexchange.rs
+++ b/ntpd/src/daemon/keyexchange.rs
@@ -21,13 +21,9 @@ use tokio::{
 use super::config::NtsKeConfig;
 use super::exitcode;
 
-pub(crate) async fn key_exchange_client(
-    server_name: String,
-    port: u16,
+fn build_client_config(
     extra_certificates: &[Certificate],
-) -> Result<KeyExchangeResult, KeyExchangeError> {
-    let socket = tokio::net::TcpStream::connect((server_name.as_str(), port)).await?;
-
+) -> Result<rustls::ClientConfig, KeyExchangeError> {
     let mut roots = rustls::RootCertStore::empty();
     for cert in rustls_native_certs::load_native_certs()? {
         let cert = rustls::Certificate(cert.0);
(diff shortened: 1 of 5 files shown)


ntpd/src/daemon/keyexchange.rs
```diff
@@ -647,4 +660,243 @@ mod tests {
...
+        assert!(result.is_ok());
+    }
+
+    #[tokio::test]
+    async fn double_end_of_message() {
+        let records = vec![NtsRecord::EndOfMessage, NtsRecord::EndOfMessage];
+        let result = send_records_to_server(records).await;
+
+        assert!(result.is_ok());
+    }
```
folkertdev, line 852: it's unclear if this is really correct according to the spec. Fine in practice assuming a valid client.

ntpd/src/daemon/keyexchange.rs
```diff
@@ -647,4 +660,243 @@ mod tests {
...
+            },
+            NtsRecord::EndOfMessage,
+            NtsRecord::NewCookie {
+                cookie_data: vec![1, 2, 3],
+            },
+        ])
+        .await;
+
+        assert!(result.is_ok());
+    }
```
folkertdev, line 819: I believe this is is technically not spec-compliant 

> The sequence SHALL be terminated by a "End of Message" record

but we can interpret that in different ways (e.g. we do stop reading the sequence when we find the first end of message). Not a problem for valid clients.

ntpd/src/daemon/peer.rs
```diff
@@ -617,7 +617,7 @@ mod tests {
     #[tokio::test]
     async fn test_poll_sends_state_update_and_packet() {
         // Note: Ports must be unique among tests to deal with parallelism
-        let (mut process, socket, _) = test_startup(8004).await;
+        let (mut process, socket, _) = test_startup(8006).await;
```
folkertdev, line 620: this port was used by other tests (causing cargo nextest to fail).

ntpd/src/daemon/keyexchange.rs
```diff
@@ -647,4 +660,243 @@ mod tests {
...
+        buffer.len()
+    }
+
+    async fn send_records_to_client(
+        records: Vec<NtsRecord>,
+    ) -> Result<KeyExchangeResult, KeyExchangeError> {
+        let listener = tokio::net::TcpListener::bind(("localhost", 0))
+            .await
+            .unwrap();
+        let port = listener.local_addr()?.port();
```
folkertdev, line 679: using port 0 makes the OS assign a unique port to this stream. We just need to make sure it is used consistently

folkertdev (commented):
there are some tests with behavior that is maybe not entirely in line with the specification. I don't think those are important issues though, because they don't matter for valid clients.

rnijveld (approved):

PR #1235, opened 2023-11-27
SupportedAlgorithm list response

squell:
This adds a response for the NTS record. There are some design issues in the spec that are unclear that might impact whether or not this is the right way to handle this proposed NTS record.

NOTE: this PR uses #1228 as a base, so that should be merged first.


squell:
Note: this can be reviewed commit-by-commit. The final three commits can be spun off in a seperate PR as desired.

davidv1992 (commented):
Mostly looks good, but would prefer the changes we discussed in private

PR #1236, opened 2023-11-27
pool KE server list

folkertdev:
add config for providing a list of KE servers

### DIFF
--- a/nts-pool-ke/src/lib.rs
+++ b/nts-pool-ke/src/lib.rs
@@ -3,20 +3,26 @@ mod config;
 
 mod tracing;
 
-use std::{io::BufRead, ops::ControlFlow, path::PathBuf, sync::Arc};
+use std::{
+    io::{BufRead, ErrorKind},
+    ops::ControlFlow,
+    path::PathBuf,
+    sync::Arc,
+};
 
-use ::tracing::info;
+use ::tracing::{info, warn};
 use cli::NtsPoolKeOptions;
 use config::{Config, NtsPoolKeConfig};
 use ntp_proto::{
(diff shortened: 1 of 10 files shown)


folkertdev:
@squell based on this CI run https://github.com/pendulum-project/ntpd-rs/actions/runs/7128390621/job/19410268392 I think there is still flakiness in our tests somewhere. I cannot reproduce it on my hardware though.

squell:
I don't see that test failing, but I do see `test_keyexchange_roundtrip_fixed_not_authorized` failing locally.

squell:
This test is passing normally on my machine.

nts-pool-ke/src/lib.rs
```diff
@@ -209,10 +230,73 @@ async fn pool_key_exchange_server(
...
+    denied_servers: &[String],
+) -> Result<(&'a str, u16, ServerName), KeyExchangeError> {
+    for server in servers {
+        if denied_servers.contains(&server.domain) {
+            continue;
+        }
+
+        match pick_nts_ke_server(connector, server, selected_algorithm).await {
+            Ok(x) => return Ok(x),
+            Err(e) => match e {
```
squell, line 281: Shouldn't we also just `continue` if this server produces a `KeyExchangeError::NoValidAlgorithm` ? otherwise one ntske server that doesn't support our algorithm can frustrate the entire connection?

squell (dismissed):

squell (approved):
Can be merged, but the comment should still be addressed (extra commits for that can be added without needing a re-review), or we can handle that in a separate PR.

squell:
Since this also occurs in other cases (e.g. an invalid certificate on a server), let's handle that in issue #1248

PR #1238, opened 2023-11-30
check for exactly one NextProtocol record in the response

folkertdev:
the spec notes

> The NTS Next Protocol Negotiation record has a Record Type number of 1. It MUST occur exactly once in every NTS-KE request and response.

we did not check this requirement

### DIFF
--- a/ntp-proto/src/nts_record.rs
+++ b/ntp-proto/src/nts_record.rs
@@ -905,11 +905,18 @@ impl KeyExchangeResultDecoder {
                     .iter()
                     .find_map(|proto| protocol_ids.contains(&(*proto as u16)).then_some(*proto));
 
-                state.protocol = selected;
-
-                match state.protocol {
+                match selected {
                     None => Break(Err(NoValidProtocol)),
-                    Some(_) => Continue(state),
+                    Some(protocol) => {
+                        // The NTS Next Protocol Negotiation record [..] MUST occur exactly once in every NTS-KE request and response.
+                        match state.protocol {
+                            None => {
+                                state.protocol = Some(protocol);
+                                Continue(state)
+                            }
+                            Some(_) => Break(Err(KeyExchangeError::NoValidProtocol)),
+                        }
+                    }
                 }
             }
             AeadAlgorithm { algorithm_ids, .. } => {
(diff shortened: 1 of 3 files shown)


ntpd/src/daemon/keyexchange.rs
```diff
@@ -840,15 +840,23 @@ mod tests {
         let records = vec![NtsRecord::EndOfMessage];
         let result = send_records_to_server(records).await;
 
-        assert!(result.is_ok());
+        assert!(matches!(result, Err(KeyExchangeError::NoValidProtocol)));
     }
 
     #[tokio::test]
-    async fn double_end_of_message() {
-        let records = vec![NtsRecord::EndOfMessage, NtsRecord::EndOfMessage];
```
folkertdev, line 848: this test is flaky, and fixing that is hard in this PR

ntp-proto/src/nts_record.rs
```diff
@@ -905,27 +919,54 @@ impl KeyExchangeResultDecoder {
...
                     None => Break(Err(NoValidProtocol)),
-                    Some(_) => Continue(state),
+                    Some(protocol) => {
+                        // The NTS Next Protocol Negotiation record [..] MUST occur exactly once in every NTS-KE request and response.
+                        match state.protocol {
+                            None => {
+                                state.protocol = Some(protocol);
+                                Continue(state)
+                            }
+                            Some(_) => Break(Err(KeyExchangeError::BadRequest)),
```
davidv1992, line 931: BadRequest seems a weird description for an error which is essentially "server did not give valid response"

davidv1992 (commented):
Few minor comments regarding choice of errors, but otherwise looks ok to me, though I want @squell's opinion whether this interferes with any of the previous pool work.

ntp-proto/src/nts_record.rs
```diff
@@ -905,27 +919,54 @@ impl KeyExchangeResultDecoder {
...
+                        }
+                    }
                 }
             }
             AeadAlgorithm { algorithm_ids, .. } => {
+                // it MUST include at most one
+                let algorithm_id = match algorithm_ids[..] {
+                    [] => return Break(Err(NoValidAlgorithm)),
+                    [algorithm_id] => algorithm_id,
+                    _ => return Break(Err(BadRequest)),
```
davidv1992, line 941: BadRequest seems a weird description for an error which is essentially "server did not give valid response"

folkertdev:
fair point. None of the existing errors seemed fitting, so I added a `BadResponse` error for this case where the response is syntactically valid but semantically invalid.

squell (approved):
Already had a discussion about this with @folkertdev about these changes last week and they look fine to me.

PR #1251, opened 2023-12-08
prevent 'downstream' errors reported by NTS-KE servers from interfering with NTS-pool operationg

squell:


nts-pool-ke/src/lib.rs
```diff
@@ -265,29 +260,39 @@ async fn pick_nts_ke_server<'a>(
...
             continue;
         }
 
         match pick_nts_ke_server(connector, server, selected_algorithm).await {
             Ok(x) => return Ok(x),
-            Err(e) => match e {
-                KeyExchangeError::Io(e) if e.kind() == ErrorKind::ConnectionRefused => continue,
-                _ => return Err(e),
-            },
+            Err(e) if is_fatal_error(&e) => return Err(e),
```
squell, line 284: **Note for the reviewer**

I am quite unsure if having `is_fatal_error` is a wise idea at all. It was clear that the original code was too strict (already prevented a normal test setup from operating in case one NTSKE server presented the wrong certificate, for example).

But imagine if a NTS-POOL consists of many servers; with this PR, if one of them sends a `BadRequest` it still can bring the whole pool down, which is a DoS risk. But I find it hard to identify legitimate use cases where we would want to stop trying alternate NTSKE servers.

davidv1992 (approved):
Share your doubts on whether the is_fatal_error is a good idea longterm. However, this is sufficient of an improvement that I think we should merge it.

squell:
After discussion with Folkert, I think a better solution (which avoids leaking key material) is to remove a server giving a "fatal error" from the list of NTSKE's to try.

PR #1274, opened 2023-12-21
Add newline at the end of metrics

### LINKED ISSUE #1273: metrics exporter output cannot be parsed by telegraf (github.com/prometheus/common/expfmt go library)
I have configured telegraf to read my ntpd-rs metrics endpoint, configuration as follows:
```
[[inputs.prometheus]]
  urls = ["http://127.0.0.1:9975/"]
```

When configured like this, telegraf complains at every scrape:
```
2023-12-21T14:43:05.339656+10:00 ntp7 telegraf[25529]: 2023-12-21T04:43:05Z E! [inputs.prometheus] Error in plugin: error reading metrics for "http://127.0.0.1:9975/": reading text format failed: text format parsing error in line 126: unexpected end of input stream
```

The ntpd-rs metrics endpoint on my system only sends 125 lines of output with the last line being "# EOF" without a terminating newline.  I suspect the lack of newline is what is causing the issue, but that's only a guess.

The code which produces the error message in telegraf is at https://github.com/influxdata/telegraf/blob/master/plugins/inputs/prometheus/parser.go#L50, and the underlying library it uses is https://github.com/prometheus/common/tree/main/expfmt, which is invoked at https://github.com/influxdata/telegraf/blob/master/plugins/inputs/prometheus/parser.go#L16.

This is probably a bug in the Prometheus golang library not being liberal in what it accepts, but I suspect it will be easier to fix in ntpd-rs than to get the core parsing library in Prometheus updated.  Strangely, my Prometheus server has no problems parsing the metrics, so it must not use its own official parser...

paulgear:
Fixes #1273

### DIFF
--- a/ntpd/src/metrics/mod.rs
+++ b/ntpd/src/metrics/mod.rs
@@ -374,6 +374,6 @@ pub fn format_state(w: &mut impl std::fmt::Write, state: &ObservableState) -> st
         collect_servers!(state, |s| s.stats.nts_nak_packets.get()),
     )?;
 
-    w.write_str("# EOF")?;
+    w.write_str("# EOF\n")?;
     Ok(())
 }


davidv1992 (approved):

davidv1992:
Thank you for noticing this

PR #1312, opened 2024-01-19
Upgraded rustls ecosystem to rustls 0.22

davidv1992:


nts-pool-ke/src/lib.rs
```diff
@@ -572,14 +568,14 @@ async fn supported_algorithms_request(
 
 fn private_key_from_bufread(
     mut reader: impl BufRead,
-) -> std::io::Result<Option<rustls::PrivateKey>> {
+) -> std::io::Result<Option<rustls::pki_types::PrivateKeyDer<'static>>> {
     use rustls_pemfile::Item;
 
     loop {
         match rustls_pemfile::read_one(&mut reader)? {
```
paolobarbolini, line 575: It should be possible to replace this manual `match` implementation with [`rustls_pemfile::private_key`](https://docs.rs/rustls-pemfile/2.0.0/rustls_pemfile/fn.private_key.html)

ntpd/src/daemon/keyexchange.rs
```diff
@@ -528,31 +529,29 @@ impl<'a, 'b, T: AsyncRead + Unpin> Read for ReaderAdapter<'a, 'b, T> {
...
         match rustls_pemfile::read_one(&mut reader)? {
-            Some(Item::RSAKey(key)) => return Ok(Some(rustls::PrivateKey(key))),
-            Some(Item::PKCS8Key(key)) => return Ok(Some(rustls::PrivateKey(key))),
-            Some(Item::ECKey(key)) => return Ok(Some(rustls::PrivateKey(key))),
+            Some(Item::Pkcs1Key(key)) => return Ok(Some(key.into())),
+            Some(Item::Pkcs8Key(key)) => return Ok(Some(key.into())),
+            Some(Item::Sec1Key(key)) => return Ok(Some(key.into())),
             None => break,
             _ => {}
         }
```
paolobarbolini, line 557: Same here

paolobarbolini (commented):
:wave:. As I was checking out the project I stumbled on this PR and found a `rustls-pemfile` v2 abstraction you could use to avoid having to match on the certificate type.

davidv1992:
Thank you for the suggestion, I have incorporated it in the current version.

davidv1992:
@plugwash in what time frame would this upgrade work for debian packaging, given the fact that the new rustls version isn't backwards compatible?

davidv1992:
@sylvestre @plugwash Should we wait with merging this for a while or is this upgrade not going to be a problem within debian?

plugwash:
Don't worry too much about us, we can always revert stuff until we are ready.

rnijveld (approved):

PR #1352, opened 2024-02-16
Create var folder on setup.

### LINKED ISSUE #1115: Packaging should pre-create the /var/lib/ntpd-rs directory
This directory (`/var/lib/ntpd-rs`) is used in examples for keyset storage and is not created by default which makes for a less than optimal user experience. We should create it with the owner and group set to ntpd-rs and with permissions set to `0750` (or possibly even `0700`). This allows the user to easily use the keyset storage with relatively safe defaults.

davidv1992:
Closes #1115


davidv1992:
This needs more thinking, see original issue, since this gives a collision with the name of one of the sockets.

PR #1381, opened 2024-03-15
Spawner refactor

davidv1992:


ntpd/src/daemon/spawn/mod.rs
```diff
@@ -272,24 +259,43 @@ where
...
-        self.handle_init(&action_tx).await?;
-        while let Some(event) = timeout(Duration::from_secs(60), system_notify.recv())
-            .await
-            .unwrap_or(Some(SystemEvent::Idle))
-        {
+        let mut has_ticket = true;
+        let mut last_ticket_time = Instant::now();
+
+        loop {
+            if Instant::now() - last_ticket_time >= NETWORK_WAIT_PERIOD {
```
rnijveld, line 266: I think `last_ticket_time.elapsed() >= NETWORK_WAIT_PERIOD` is easier to understand.

ntpd/src/daemon/spawn/mod.rs
```diff
@@ -272,24 +259,43 @@ where
...
+                has_ticket = true;
+            }
+
+            if has_ticket && !self.is_complete() {
+                self.try_spawn(&action_tx).await?;
```
rnijveld, line 271: Do we still want to add any sort of timeout mechanism here? What are the max timeouts for DNS and for NTS handshake? Is it ok if we wait that long and not receive any system_notify events? I think we should at least add a little more documentation for try_spawn to remind implementers that they should try and keep the spawning logic to as short a time a possible.

ntpd/src/daemon/spawn/mod.rs
```diff
@@ -272,24 +259,43 @@ where
...
+                has_ticket = false;
+                last_ticket_time = Instant::now();
+            }
+
+            let event = if has_ticket {
+                system_notify.recv().await
+            } else {
+                timeout(
+                    last_ticket_time + NETWORK_WAIT_PERIOD - Instant::now(),
```
rnijveld, line 280: `NETWORK_WAIT_PERIOD - last_ticket_time.elapsed()`?

davidv1992: done

ntpd/src/daemon/spawn/mod.rs
davidv1992: done

rnijveld (approved):

PR #1443, opened 2024-04-26
Renamed peer to source, now also in the code.

davidv1992:


docs/guide/migrating-chrony.md
```diff
@@ -62,7 +62,7 @@ initial-poll-interval = <desired initial poll interval>
 
 There is currently no support for bursting in ntpd-rs. When any bursting option (`burst` or `iburst`) is present, these usually can be ignored, although if custom poll limits are in place, these may need to be relaxed.
 
-### Peer directives
```
rnijveld, line 65: I believe this header talks about Chrony peer directives, not ntpd-rs ones.

ntpd/src/daemon/config/ntpsource.rs
rnijveld, line 1: Might be better to rename this to ntp_source to better align with nts_pool and nts_record?

rnijveld (changes requested):

rnijveld (approved):

PR #1456, opened 2024-05-07
Replace certificate loading by a non-blocking variant

rnijveld:
Filling the rootstore with certificates from the native pool can take quite a while (given the large amount of certificates to load), so let's do that on a separate task that is allowed to block to not completely block the tokio executor. I don't think we need to properly handle the JoinError and bubble it up, as that seems very unlikely to happen, and there is not much we can do with it anyway.


davidv1992 (approved):

davidv1992:
Mostly looks good, needs some fixes for the failing checks.

davidv1992 (approved):

PR #1496, opened 2024-06-13
Add length before sending observability messages over socket

### LINKED ISSUE #1495: Handle arbitrarily sized metrics messages

rnijveld:
Messages are now sent as `[u64_be length of message][message]` to allow us to know when the message is complete. This prevents any issues with larger messages from occuring.
 
Fixes #1495


cikzh:
We should check if this same fix should also be implemented for [Statime](https://github.com/pendulum-project/statime)

davidv1992 (approved):

Issue #1497, opened 2024-06-13
Prometheus Alertmanager shows ntpd-rs down intermittently


Bradatletsencrypt:
Hello,

We have been working on deploying `ntpd-rs` and it's been going well. We really appreciate that we are able to get metrics in a way that's very easy and intuitive. Upon deploying to staging and then production, we noticed that sometimes alertmanager had a pending alert for `ntpd-rs` being down. It never fired and would always clear on the next scrape. I decided to see if I could see this same behavior in my dev environment, and I am able to:

![image](https://github.com/pendulum-project/ntpd-rs/assets/132489699/5900f1b9-07d5-4d21-8eb4-5ad2c6907ba9)

When investigating the nodes that report down, systemd doesn't show any restarts of the `ntpd-rs` service or the related metrics service. I enabled debug logging in alertmanager and grepped through logs by the node names that appeared down only to not find any log entries for them. For example, I can see scrape targets that have been reported down with:
```
sudo journalctl -u prometheus-alertmanager --since "12 hours ago" --until "now" | grep ntpd-rs | grep Down | cut -d, -f2 | cut -d\" -f2 | sed -e 's/\\//g' | sort | uniq -c
```
The only output from this is the scrape targets I did _not_ bring up in my dev environment, so that's expected to be in the log. Grepping for the node names alertmanager thought were down for a scrape yields nothing in the logs. I then wanted to see if I could figure out what was actually happening during one of these "failures" -- for this I set up a couple of different loops:

```
# capture the time it takes to cURL each scrape target
for i in {1..500}; do for i in $(cat hosts.txt); do echo $i; time curl -s $i:9975/metrics | wc -l; done >> time_test.log 2>&1; done

# count the number of lines in output of each scrape target
for i in {1..250}; do for i in $(cat hosts.txt); do echo $i; curl -s $i:9975/metrics | wc -l; done; sleep 5; done >> wc_test.log
```
For the former, I can just grep for `real` to see if any large timeouts appear to be happening:
```
$ grep real time_test.log 
real	0m0.010s
real	0m0.013s
real	0m0.008s
real	0m0.008s
real	0m0.009s
real	0m0.007s
real	0m0.008s
real	0m0.008s
real	0m0.007s
real	0m0.008s
real	0m0.009s
real	0m0.012s
real	0m0.009s
real	0m0.012s
real	0m0.010s
real	0m0.008s
real	0m0.008s
real	0m0.012s
real	0m0.011s
[...]

PR #1506, opened 2024-06-20
Run handling of observability on separate tokio tasks

rnijveld:
This should improve any concurrency issues that could happen while scraping metrics from our observability socket. I've tested this with a little siege test (`siege -b -t30S -c10 http://127.0.0.1:9975/metrics`) to check the throughput, before the change we had these numbers:

```json
{	"transactions":			      629647,
	"availability":			      100.00,
	"elapsed_time":			       29.80,
	"data_transferred":		     3063.39,
	"response_time":		        0.00,
	"transaction_rate":		    21129.09,
	"throughput":			      102.80,
	"concurrency":			        9.69,
	"successful_transactions":	      629647,
	"failed_transactions":		           0,
	"longest_transaction":		        0.01,
	"shortest_transaction":		        0.00
}
```

And after the change we get these numbers:
```json
{	"transactions":			     1051395,
	"availability":			      100.00,
	"elapsed_time":			       29.03,
	"data_transferred":		     5120.10,
	"response_time":		        0.00,
	"transaction_rate":		    36217.53,
	"throughput":			      176.37,
	"concurrency":			        9.41,
	"successful_transactions":	     1051395,
	"failed_transactions":		           0,
	"longest_transaction":		        0.02,
	"shortest_transaction":		        0.00
}
```

I'm not entirely sure this solves the issues experienced in #1497, given that the longest transaction time wasn't that long and we had no failures. But it at the very least it doesn't hurt. I expect the root cause of #1497 might also have to do something with the server doing something blocking (which I didn't test in my client-only ntpd-rs setup), but I'll have to explore a little further.

### DIFF
--- a/ntpd/src/daemon/observer.rs
+++ b/ntpd/src/daemon/observer.rs
@@ -2,13 +2,14 @@ use super::server::ServerStats;
 use super::sockets::create_unix_socket_with_permissions;
 use super::spawn::SourceId;
 use super::system::ServerData;
+use libc::{ECONNABORTED, EMFILE, ENFILE, ENOBUFS, ENOMEM};
 use ntp_proto::{ObservableSourceState, SystemSnapshot};
 use std::collections::HashMap;
 use std::os::unix::fs::PermissionsExt;
 use std::sync::Arc;
 use std::{net::SocketAddr, time::Instant};
 use tokio::task::JoinHandle;
-use tracing::warn;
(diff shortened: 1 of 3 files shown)


davidv1992:
This needs limiting of max number of connections, both on the deamon side and on the http exporter binary.

davidv1992 (approved):

PR #1512, opened 2024-06-24
Add force-sync command to ntp-ctl

### LINKED ISSUE #1494: Add a force-sync command to ntp-ctl
Sometimes a system ends up in a state where the clock is not synced and the ntpd-rs daemon no longer wants to sync, the daemon then exists with a decent error message in that it will not jump, but in general users will not know how to change the system time.

We should have some kind of command that does the forced sync for the user. I believe such a command should require user input as to try and prevent its automatic usage (i.e. we want system admins to be aware that this command might sync the time to a wrong time and first offer them the ability to deny the suggested time jump).

rnijveld:
This builds on top of #1508 which should probably be looked at first. I've made a couple of items public in implementing this. I think we could reduce the amount of duplicated code a little more by restructuring a little more, but I wanted to get a quick example of this utility as soon as possible. The source spawning currently happens in serial, but this could be made to be in parallel as well, but I don't think this tool needs to be that performant though. I wonder if we still would want to add a little more interactivity to this as well, just so users understand that this bypasses most safety checks. This fixes #1494.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -39,8 +39,8 @@ serde_json = "1.0"
 serde_test = "1.0.176"
 rand = "0.8.0"
 arbitrary = { version = "1.0" }
-libc = "0.2.145"
-tokio = "1.32"
+libc = "0.2.149"
+tokio = "1.34"
 toml = { version = ">=0.6.0,<0.9.0", default-features = false, features = ["parse"] }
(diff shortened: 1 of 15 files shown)


davidv1992:
@ruben needs rebase

rnijveld:
Superseeded by #1568

PR #1513, opened 2024-06-25
Create FUNDING.yml

rnijveld:
This should add some funding links in the sidebar on our github page

### DIFF
--- a/.github/FUNDING.yml
+++ b/.github/FUNDING.yml
@@ -1,0 +1,14 @@
+# These are supported funding model platforms
+
+github: [pendulum-project]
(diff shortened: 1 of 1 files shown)


erikjee (commented):
Is the format for custom correct?

Example here: https://github.blog/2020-03-24-getting-started-with-github-sponsors/

```
patreon:            patreon_username
custom:             ["custom_url.com", “another_custom_url.com”]
```

rnijveld:
I shouldn't make a difference, given its just yaml, but just to be sure I added some quotation marks around it!

erikjee (approved):

Issue #882, opened 2023-07-22
Add Pulse Per Second (PPS) Support


rlaager:
I could be missing it, but at a quick glance, I'm not seeing any support for PPS input. Ideally, there should be support for two configurations:

A) Using SHM from gpsd. You might call this "SHM support" and consider it a separate feature request.

B) Taking PPS from the kernel PPS API and using it to discipline the clock. The seconds would be numbered from other (e.g. Internet) sources.

I use configuration B in production with NTPsec (and formerly with classic ntpd). I have telecom grade GPS clocks which provide a PPS output (among other things).

While such devices can often act as NTP servers, sometimes this is a separately licensed (expensive) feature. In any case, their NTP server implementations are generally not as good in some respects (e.g. not supporting NTS). Likewise, newer ones tend to support PTP, but again, that can be an expensive licensed feature.

So I just use PPS from them. This is probably more accurate anyway, and allows me to be stratum 1 instead of stratum 2. I'm not trying to operate standalone (i.e. with no Internet access), so I don't need to get the TOD from GPS via serial. (And on my previous clocks, it was wrong due to GPS rollover anyway.)

davidv1992:
We indeed do not support this yet, primarily because we didn't have people who wanted it and so couldn't ask questions, so thank you for asking :).

I'll add this to the discussion internally for prioritization, but already have a few more specific questions:
- If you can disclose this: what hardware are you using for this in production?
- What hardware setup would you recommend for us testing this?

rlaager:
I'm using an Ocilloquartz (now ADVA) product. I was using a Symmetricom product before. I've got another system where I'm using a u-blox GPS evaluation board.

They all output a PPS which is connected into the computer on a serial port control line (DCD, I think).

The steps to configure on Debian (and with NTPsec) are:

1. Install `pps-tools` and `setserial`.
2. `/etc/modules-load.d/pps-ldisc.conf` contains `pps_ldisc`
3. `/etc/systemd/system/pps-ldisc@.service` contains:
```
[Unit]
Description=Line Discipline for PPS Timekeeping for %i
Before=ntpsec.service
BindsTo=dev-%i.device
Wants=systemd-modules-load.service
After=systemd-modules-load.service
After=setserial.service

[Service]
ExecStart=/usr/sbin/ldattach pps /dev/%i
Type=forking

[Install]
WantedBy=ntpsec.service
```
4. `systemctl enable pps-ldisc@ttyS0`
[...]

PR #1517, opened 2024-06-26
Gps and Pps support

krioyo:
This is our implementation for Gps and PPS support for NTPD-rs. It also acts as a handover for the Software Project course we followed at Tu Delft. 

This pr let's the user configure the port name and measurement noises for gps and pps signals. It then integrates these signals into the existing codebase to aid in determining the true time.  It seeks to to close:  #882 

We look forward to getting a review!


JamesB192:
I have yet to join the project; this is not the review you requested.

This fork is IMO underdocumented. You have 180 commits (roughly 5x the main branch's progress minus many dependency bumps), and the entire documentation is a few paragraphs in README.md. You could also label the TOML config excerpt as such; there is no indication of which GPS protocols are supported (of ~20, my guess is only a hypothetical version NMEA0183)

rnijveld:
@JamesB192 Please be aware that this PR was a student project exploring the GPS and PPS functionality. We've worked closely with them, so we have some context that is unavailable on this PR. This code isn't ready for ntpd-rs, but I believe we were planning to look at this in the coming months. Unfortunately we only have limited budget and time to work on it.

PR #1542, opened 2024-07-10
Made ntp-proto algorithm-agnostic.

davidv1992:


ntp-proto/src/config.rs
```diff
@@ -266,9 +263,6 @@ pub struct SynchronizationConfig {
     /// synchronizing the clock
     #[serde(default = "default_local_stratum")]
     pub local_stratum: u8,
-
-    #[serde(default)]
-    pub algorithm: AlgorithmConfig,
```
rnijveld, line 271: I believe this change requires an update to the ntp.toml manpage.

rnijveld (changes requested):

ntp-proto/src/config.rs
davidv1992: This isn't needed because of flattening in the actual read configuration struct.

rnijveld (approved):

Issue #1632, opened 2024-08-30
Implement support for the SOCK protocol


davidv1992:
Support the SOCK protocol for delivering time to ntpd-rs. Relevant implementation details needs to be reverse engineered from chrony.

PR #1673, opened 2024-10-02
Allow measurements without delay

michielp1807:
This PR allows the `delay` property of measurements to be either `NtpDuration` or `()`. If the delay is `()`, we use a constant value as the estimate of the measurement noise instead of an estimate based on the variance of delay values. This is useful for supporting alternative time sources, such as the SOCK protocol (#1632).

### DIFF
--- a/ntp-proto/src/algorithm/kalman/mod.rs
+++ b/ntp-proto/src/algorithm/kalman/mod.rs
@@ -502,7 +502,7 @@ mod tests {
             noise += 1e-9;
 
             let message = source.handle_measurement(Measurement {
-                delay: NtpDuration::from_seconds(0.001 + noise),
+                delay: Some(NtpDuration::from_seconds(0.001 + noise)),
                 offset: NtpDuration::from_seconds(1700.0 + noise),
                 localtime: algo.clock.current_time,
                 monotime: cur_instant,
@@ -712,7 +712,7 @@ mod tests {
             noise += 1e-9;
 
             let message = source.handle_measurement(Measurement {
-                delay: NtpDuration::from_seconds(0.001 + noise),
+                delay: Some(NtpDuration::from_seconds(0.001 + noise)),
                 offset: NtpDuration::from_seconds(1700.0 + noise),
                 localtime: algo.clock.current_time,
                 monotime: cur_instant,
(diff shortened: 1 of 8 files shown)


ntp-proto/src/algorithm/kalman/mod.rs
```diff
@@ -1,5 +1,6 @@
 use std::{collections::HashMap, fmt::Debug, hash::Hash, time::Duration};
 
+pub use source::AveragingBuffer;
```
davidv1992, line 3: I'd strongly prefer if we do not make this type public. This is basically an implementation detail.

ntpd/src/force_sync/algorithm.rs
```diff
@@ -116,8 +150,21 @@ impl<C: NtpClock> TimeSyncController for SingleShotController<C> {
...
-        SingleShotSourceController {
+    fn add_source(&mut self, _id: Self::SourceId) -> Self::NtpSourceController {
+        SingleShotSourceController::<NtpDuration> {
+            delay_type: PhantomData,
+            min_poll_interval: self.min_poll_interval,
+            done: false,
+        }
+    }
+
+    fn add_sock_source(
```
davidv1992, line 161: bit bikeshedding, but I would prefer a more generic name, as this is likely going to be used for more than just sock sources.

ntp-proto/src/system.rs
```diff
@@ -242,52 +242,48 @@ impl<SourceId: Hash + Eq + Copy + Debug, Controller: TimeSyncController<SourceId
         Ok(())
     }
 
-    #[allow(clippy::type_complexity)]
-    pub fn create_ntp_source(
+    pub fn create_source_controller(
```
davidv1992, line 245: I don't think we should leak the controllers out of system. I'd much rather have a thin-ish wrapper for non-ntp sources as well. 

Although the idea of combining both nts and non-nts source creation isn't necessarily a bad thing.

davidv1992 (changes requested):
Some choices around type leaking that I don't agree with. The actual logic changes seem ok though.

ntp-proto/src/system.rs
michielp1807: Should be fixed in cafc220 (as part of the other PR because here we do not have the sock source yet.)

ntp-proto/src/algorithm/kalman/mod.rs
michielp1807: Fixed in 180f7e1

PR #1679, opened 2024-10-12
Pedantic linting

### LINKED ISSUE #1658: Enable pedantic clippy on all crates and deal with the fallout
We really should do as it would have prevented a bug. But this is going to take some effort... Be prepared that this is not the most fun task.

Also, be carefull not to introduce msrv changes with this.

van-sprundel:
- [x] nts-pool-ke
- [x] ntp-proto
- [x] ntpd

Changes for #1658. 

There are some opinionated changes here, so feel free to call out anything that's not wanted.
All changes are first ran with `cargo clippy --fix -p [project] --all-targets --all-features`

I've left out `clippy::module_name_repetitions` as it doesn't add much here.

Some notable lints from the pedantic group:
- Functions that can error/panic now have error docs when relevant.
- Big functions (>100 lines) are split into different functions.
  - Except test modules, which are marked with `#[allow]`
- `ntpd::ctl::NtpCtlOptions` now uses the `action` state instead of bools.
- Some `match` cases became `if () { } else { }` where relevant.
- Some `let foo = if ... else` became `let ... else` where relevant.
- `async` and `Result<_>` are removed on places where it wasn't used and relevant.
- `const` and `enum` mid-function are moved up at the beginning of the function, or moved up a scope.
- `ntpd::daemon:;ntp_source::SourceTask::run` has been split up into multiple functions.
- Conversions that could truncate or lose sign are marked as allowed where relevant.
  - For arm and darwin, all truncation is allowed.

This became a pretty big PR in the end. I'm not sure how we could keep the impact minimal. Splitting it into multiple branches based on crates doesn't help much, because the `ntp-proto` crate has ~1000 lines of changes.


ntp-proto/src/nts_record.rs
```diff
@@ -1143,18 +1196,29 @@ pub struct KeyExchangeClient {
...
+    /// # Errors
+    ///
+    /// Returns error if reading the TLS socket fails.
     pub fn read_socket(&mut self, rd: &mut dyn Read) -> std::io::Result<usize> {
         self.tls_connection.read_tls(rd)
     }
 
+    /// # Errors
+    ///
+    /// Returns error if getting the TLS bool `wants_write` fails.
```
davidv1992, line 1213: Doesn't actually return an error

ntp-proto/src/nts_record.rs
```diff
@@ -1410,74 +1483,22 @@ impl KeyExchangeServerDecoder {
...
-                        .copied()
-                        .find_map(ProtocolId::try_deserialize)
-                };
-
-                #[cfg(not(feature = "ntpv5"))]
-                let selected = protocol_ids
-                    .iter()
-                    .copied()
-                    .find_map(ProtocolId::try_deserialize);
-
```
davidv1992, line 1440: Same as above, please split these refactors into a separate PR

ntp-proto/src/nts_record.rs
```diff
@@ -1534,28 +1625,43 @@ enum State {
...
+    /// # Errors
+    ///
+    /// Returns error if reading the TLS socket fails.
     pub fn read_socket(&mut self, rd: &mut dyn Read) -> std::io::Result<usize> {
         self.tls_connection.read_tls(rd)
     }
 
+    /// # Errors
+    ///
+    /// Returns error if getting the TLS bool `wants_write` fails.
```
davidv1992, line 1642: Doesn't actually error

ntp-proto/src/packet/extension_fields.rs
```diff
@@ -1,3 +1,4 @@
+#![allow(clippy::cast_possible_truncation)]
```
davidv1992, line 1: Note to self: Eliminate here, too general

ntp-proto/src/packet/mod.rs
```diff
@@ -305,129 +313,136 @@ impl<'a> NtpPacket<'a> {
...
-            3 => {
-                let (header, header_size) =
-                    NtpHeaderV3V4::deserialize(data).map_err(|e| e.generalize())?;
-                let mac = if header_size != data.len() {
-                    Some(Mac::deserialize(&data[header_size..]).map_err(|e| e.generalize())?)
-                } else {
-                    None
-                };
-                Ok((
-                    NtpPacket {
```
davidv1992, line 317: Good refactor on this match, but again needs more involved review. Could you please make this a separate PR?

ntpd/src/ctl.rs
```diff
@@ -122,28 +117,10 @@ impl NtpCtlOptions {
...
-        } else if self.validate {
-            self.action = NtpCtlAction::Validate;
-        } else if self.status {
-            self.action = NtpCtlAction::Status;
-        } else if self.force_sync {
-            self.action = NtpCtlAction::ForceSync;
-        } else {
-            self.action = NtpCtlAction::Help;
-        }
-    }
```
davidv1992, line 146: Needs fix: it is actually quite subtle, but removing this, in combination with the changes above, actually changes behaviour, and in an unwanted way. In particular, it changes behaviour of the -v and -h flags overriding all other actions, which is undesired. Please revert.

ntpd/src/daemon/ntp_source.rs
```diff
@@ -137,55 +137,38 @@ where
...
-            let actions = match selected {
-                SelectResult::Recv(result) => {
-                    tracing::debug!("accept packet");
-                    match accept_packet(result, &buf, &self.clock) {
-                        AcceptResult::Accept(packet, recv_timestamp) => {
-                            let send_timestamp = match self.last_send_timestamp {
-                                Some(ts) => ts,
-                                None => {
-                                    debug!(
-                                        "we received a message without having sent one; discarding"
```
davidv1992, line 149: Needs fix: Again, this is likely (part of) a good refactor, but too hard to review well in this context, please split this off into a separate PR.

davidv1992 (changes requested):
Thank you for picking this up. The changes make very clear that this was a good idea, even though it will still be some work to fully get these changes to land. Below I have made a number of comments. Note that most of these are notes to ourselves that we need to improve code in some places more and those need not be blockers for landing this in my eyes. I have marked those that do require more attention.

Please also note that given the wide scope of these changes, I am not comfortable merging these only on my opinion, so once this is further along there is going to be a review by a second person.

Again, thank you for doing the legwork on this. By the looks of it, you saved us quite some work. I hope you can bear with us while we make the push to get this in reviewable enough to merge it.

davidv1992:
@rnijveld As a headsup, just so you are aware of this PR, I am going to ask you to review this once it is a bit further along. Most of it is relatively straightforward, and hopefully we can split the parts that aren't into separate PRs. It does highlight a few places to where we should probably take a second look at code, particularly around casts.

van-sprundel:
Sorry for the radio silence, I'll take a look at the review and rebase the PR.

Am I safe to assume that all the reviews without "note to self" are something I can refactor?

davidv1992:
Yes, those are things that need attention before this can move forward.

van-sprundel:
While refactoring I did found that `test_deny_stops_poll` will not always succeed (even on main). Seems like a timing issue. Could it be that the cast to u32 is causing this?

https://github.com/pendulum-project/ntpd-rs/blob/419832f28b8560ae44fb178e54d2ba6328cc6da0/ntpd/src/daemon/ntp_source.rs#L541-L548

or something in the test startup that's failing it? my initial changes made it always fail, but now it sometimes fails on main

```sh
cargo test test_deny_stops_poll --target x86_64-unknown-linux-gnu --features run_tokio_rustls_tests
```

ntpd/src/ctl.rs
van-sprundel: Seems like this was refactored in-between time, nice

ntp-proto/src/packet/mod.rs
van-sprundel: Refactored already, looks good!

cikzh:
Thank you for great work! It looks like we severely underestimated this being a good first issue and the impact it would have. I've discussed this with @rnijveld and @davidv1992 and we think the best way forward is to hack this PR up into smaller, better reviewable chunks, so we can evaluate them piece-by-piece. Also, given the time this ended up taking, it will be easier to stay up-to-date with `main` that way.

We've created a tracking issue (#1818), so we can track the progress and manage the workload. If some commits of this branch are easily cherry-pickable into seperate PR's, we'd appreciate the help!

rnijveld:
Closing this PR for now as explained previously :)

PR #1697, opened 2024-11-02
Fallback to V4 should V5 not work for some reason.

davidv1992:


ntp-proto/src/source.rs
```diff
@@ -637,14 +653,16 @@ impl<Controller: SourceController> NtpSource<Controller> {
...
+                    self.protocol_version = ProtocolVersion::UpgradedToV5;
                 } else if tries_left == 0 {
                     debug!("Server does not support NTPv5, stopping the upgrade process");
                     self.protocol_version = ProtocolVersion::V4;
                 } else {
                     debug!(tries_left, "Server did not yet respond with upgrade code");
                     self.protocol_version = ProtocolVersion::V4UpgradingToV5 { tries_left };
                 };
+            } else if let ProtocolVersion::UpgradedToV5 = self.protocol_version {
+                self.protocol_version = ProtocolVersion::V5;
```
rnijveld, line 665: Do we need to check if the version in the packet we just got is actually v5 before we commit here?

rnijveld (approved):

PR #1698, opened 2024-11-02
Make ntp version on sources configurable.

davidv1992:


ntpd/src/daemon/config/ntp_source.rs
```diff
@@ -6,15 +6,34 @@ use std::{
...
+    }
+}
+
 #[derive(Deserialize, Debug, PartialEq, Eq, Clone)]
 #[serde(deny_unknown_fields)]
 pub struct StandardSource {
     pub address: NtpAddress,
+    #[cfg(feature = "unstable_ntpv5")]
+    #[serde(default, deserialize_with = "deserialize_ntp_version")]
+    pub ntp_version: Option<NtpVersion>,
```
rnijveld, line 36: I think we should probably add a test here that (a) tests the config parsing and (b) checks that the source is actually spawned with that `ProtocolVersion` specified

rnijveld (approved):

Issue #1696, opened 2024-10-31
Config option to disable colour in logs


mcpherrinm:
We feed ntpd-rs' logs into a log search product.

While it works, it contains ansi escape codes.

For example,

```
 #033[2m2024-10-31T18:38:56.000846Z#033[0m #033[32m INFO#033[0m #033[2mntp_proto::algorithm::kalman#033[0m#033[2m:#033[0m Offset: -0.098332274256776+-0.2337313649729611ms, frequency: 0.04341250045413119+-0.07468081620691763ppm
```

While this is tolerable, and we could probably even post-process the logs, it would be nice to be able to opt-out of this.

I think in `ntpd/src/daemon/tracing.rs`, in `tracing_init`, we'd want a way to set `.with_ansi(false)` in the same way as the level filter.

mcpherrinm:
Thanks!

PR #1726, opened 2024-11-25
Add option to disable ANSI codes in logs

michielp1807:
As suggested by #1696


davidv1992 (commented):
Looks good to me, but as this locks us into a configuration interface permanently, @ruben can you also review this?

docs/man/ntp.toml.5.md
```diff
@@ -179,6 +179,10 @@ ntp-metrics-exporter(8).
     daemon. Levels higher than the given log level are logged as well. If not
     set (the default), then logging will be completely disabled.
 
+`disable-ansi` = `true` | `false` (**false**)
```
rnijveld, line 182: I'm not sure if this is the best name for this option, I think we should try to not have double negatives (i.e. disable = false). Maybe a better option might be `ansi-colors` with a default value of true.

rnijveld (approved):

rnijveld (changes requested):
Sorry, I clicked the wrong button.

docs/man/ntp.toml.5.md
michielp1807: I agree, despite boolean values usually defaulting to false, it's probably better to not have negated options

docs/man/ntp.toml.5.md
```diff
@@ -179,6 +179,10 @@ ntp-metrics-exporter(8).
     daemon. Levels higher than the given log level are logged as well. If not
     set (the default), then logging will be completely disabled.
 
+`ansi-colors` = `true` | `false` (**true**)
+:   Disable ANSI escape codes in logs. By default, ANSI escape codes are used to
```
rnijveld, line 183: Maybe change this to something like `Can be used to disable [...]`?

rnijveld (approved):

PR #1768, opened 2024-12-13
Algorithm support for periodic sources.

### LINKED ISSUE #1631: Modify algorithm to deal with periodic time sources
Make modifications to the algorithm to support time sources that do not provide absolute time, but rather a reference point that reoccurs on a fixed interval, such as PPS sources.

davidv1992:


ntp-proto/src/algorithm/kalman/select.rs
```diff
@@ -38,30 +43,44 @@ pub(super) fn select<Index: Copy>(
...
 
     for (time, boundtype) in bounds.iter() {
         match boundtype {
-            BoundType::Start => cur += 1,
-            BoundType::End => cur -= 1,
-        }
-        if cur > max {
-            max = cur;
-            maxt = *time;
+            BoundType::Start => {
```
rnijveld, line 54: I think we should probably update the comment on top of this function (and/or maybe add a few inline comments), to explain why we now look for both the lower and higher bound.

ntp-proto/src/algorithm/kalman/select.rs
```diff
@@ -38,30 +43,44 @@ pub(super) fn select<Index: Copy>(
...
+                    maxhigh = cur;
+                    maxthigh = *time;
+                }
+                cur -= 1;
+            }
         }
     }
 
+    // Catch programming errors. If this ever fails there is high risk of missteering, better fail hard in that case
+    assert_eq!(maxlow, maxhigh);
```
rnijveld, line 72: Maybe add a third parameter for the panic message? Or at least explain 'this' in the comment?

ntp-proto/src/algorithm/kalman/select.rs
```diff
@@ -38,30 +43,44 @@ pub(super) fn select<Index: Copy>(
...
+    let max = maxlow;
+
     if max >= synchronization_config.minimum_agreeing_sources && max * 4 > bounds.len() {
```
rnijveld, line 75: I think we should probably update the documentation of minimum agreeing sources to indicate that it refers to non-periodic sources only?

rnijveld (dismissed):

ntp-proto/src/algorithm/kalman/select.rs
```diff
@@ -24,6 +24,11 @@ pub(super) fn select<Index: Copy>(
     let mut bounds: Vec<(f64, BoundType)> = Vec::with_capacity(2 * candidates.len());
 
     for snapshot in candidates.iter() {
+        if snapshot.period.is_some() {
```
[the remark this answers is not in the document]
davidv1992: I think we should see this as similar to having not enough sources for min_agreeing_sources, and handle this with a check and warning during the configuration check.

For right now, I don't think it is a good idea to have an option to use the current clock time for periodic sources. Although it is actually quite easy to implement, I don't see the usecase. We can always implement it later if needed

ntp-proto/src/algorithm/kalman/select.rs
davidv1992: Is in the documentation of the pps source PR

ntp-proto/src/algorithm/kalman/mod.rs
```diff
@@ -34,6 +34,7 @@ struct SourceSnapshot<Index: Copy> {
     state: KalmanState,
     wander: f64,
     delay: f64,
+    period: Option<f64>,
```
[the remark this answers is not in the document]
davidv1992: Added comment to clarify this.

ntp-proto/src/algorithm/kalman/select.rs
davidv1992: Added comments to clarify this

ntp-proto/src/algorithm/kalman/select.rs
davidv1992: Improved the comment significantly

cikzh (approved):

Issue #1633, opened 2024-08-30
Implement linux kernel pps reading


davidv1992:
Support reading time directly from a pps device. Requires #1631

PR #1776, opened 2024-12-19
Add PPS source support

michielp1807:
Builds on #1768 to be able to use Pulse Per Second devices in ntpd-rs using the [`pps-time` crate](https://github.com/pendulum-project/pps-time/), closing #1633.

Example config:
```toml
[[source]]
mode = "sock"
path = "/run/chrony.ttyAMA0.sock"
measurement_noise_estimate = 1e-6

[[source]]
mode = "pps"
path = "/dev/pps0"
measurement_noise_estimate = 1e-14
```

TODO:
- [x] Add documentation for using PPS sources in ntpd-rs
- [x] During testing I noticed it would run into [this debug assert](https://github.com/pendulum-project/ntpd-rs/blob/b5297848bb855c452f74cad4297a8579e89bfead/ntp-proto/src/algorithm/kalman/source.rs#L131) when using GPSD. It would probably be good to do some more testing :slightly_smiling_face: 
- [x] For some reason (likely precision problems?) things go wrong with low noise estimates (e.g. 1e-14), it does seem to work with a 1e-8 noise estimate -> could this be due to rounding errors with `NtpDuration`, where `to_seconds` / `from_seconds` are not exact?
- [x] Update `ntp-ctl status` to also show PPS/Sock sources


davidv1992 (commented):
These changes look good to me. However, I can't approve this as it also contains changes of me that have yet to be approved by someone.

davidv1992 (approved):
Based on the review of #1768 for the source I wrote.

Issue #1718, opened 2024-11-23
Should the reference_timestamp & reference_id be set in server only mode?


o087D:
I am attempting to configure `ntpd-rs` as a server for a system where the system time is being conditioned by `gpsd`.

My configuration is:

```
source = []

[observability]
log-level = "info"
observation-path = "./observe"

[[server]]
listen = "[::]:123"

[synchronization]
local-stratum = 1
```

The response packets from `ntpd` making a request are parsed by wireshark as being:

```
Reference ID: Unidentified reference source 'XNON'
Reference Timestamp: Jan  1, 1970 00:00:00.000000000 UTC
```

And the error I am getting from `ntpdate` is:

```
192.168.45.100: Server dropped: Server has gone too long without sync
```

(I know ntpdate is deprecated, but I believe it should still work?) The offending area in `ntpdate` is:

``` c
		if ((server->org.l_ui - server->reftime.l_ui)
		    >= NTP_MAXAGE) {
			if (debug)
				printf("%s: Server dropped: Server has gone too long without sync\n", 
				       ntoa(&server->srcadr));
			continue;	/* too long without sync */
		}
```
Where NTP_MAXAGE is set to 1 day. I do not think check is present in `ntpd` itself.

According to RFC5905:

```
   Reference Timestamp: Time when the system clock was last set or
   corrected, in NTP timestamp format.
```

Digging around I think the response builder (https://github.com/pendulum-project/ntpd-rs/blob/main/ntp-proto/src/packet/mod.rs#L233) is settings this to default and there is no other mechanism to configure it.

Would you be amenable to a stopgap PR which set the reference_timestamp to the receive_timestamp? (i.e. in the absence of an exact time the clock was last updated we can go with something very recent, rather than the epoch)

I would also like to add a configuration option to allow the reference ID word to be set in the [[server]] section of the configuration via a string.

Finally, if the server is running with established sources the reference_timestamp could be set to the actual time of the last completed poll event - but I have not looked in to the mechanics of that and expect there to be some cross-thread shenanigans required.

If these are of use I can find some time to do the work.

davidv1992:
We don't use the reference timestamp in ntpd-rs, as putting an accurate timestamp in that field leaks information that may make certain types of attacks significantly easier. I did not realize that this field was actually used by any implementation, as the ntpv4 standard does not.

[...]

PR #1815, opened 2025-02-03
Adds reference_id setting

o087D:
Enables end user to set the value for 'reference_id' if operating as a server in stratum 1, as discussed in https://github.com/pendulum-project/ntpd-rs/issues/1718


davidv1992:
@rnijveld can you take a short look at this, in particular if you agree with the new setting name and placement?

rnijveld:
Thanks for the contribution!

PR #1819, opened 2025-02-07
Switch to precision field for specifying pps and sock noise.

davidv1992:
This involves deprecating a field, we should check that this is the way we want to proceed.

Reasoning is that precision as given in a 1-stddev bound is a much more intuitive notion than measurement variance. These numbers can be interpreted as (parts of) seconds, which makes their values a  lot nicer.


rnijveld (approved):
The precision field makes a lot of sense to me as well. I think this way of deprecating the old field is fine, it's probably unlikely to have been used much in the wild anyway. We should make sure to note this in the release notes of the next version though.

PR #1836, opened 2025-02-26
Replace rustls-native-certs with rustls-platform-verifier

rnijveld:


rnijveld:
Looks like this runs into https://support.apple.com/en-us/102028

davidv1992 (approved):

PR #1873, opened 2025-04-21
Code coverage for the fuzzers

folkertdev:
Backporting lessons from zlib-rs

The comment below has a table with the flags (see dropdown). The coverage percentage itself is not too interesting, but the delta may be in the future. The real advantage is that you can now see whether the fuzzer actually hits the logic you expect.

Doing some clicking through the interface, this is the coverage for `fuzz-packet_parsing_sound`: https://app.codecov.io/github/pendulum-project/ntpd-rs/commit/b560a6393de61168dde532292c01bf5480e22af9/tree/ntp-proto/src?flags%5B0%5D=fuzz-packet_parsing_sound&dropdown=coverage

E.g. here you can see  that the fuzzer is unable to generate a valid `NtsEncryptedField`:

https://app.codecov.io/github/pendulum-project/ntpd-rs/commit/b560a6393de61168dde532292c01bf5480e22af9/blob/ntp-proto/src/packet/extension_fields.rs?flags%5B0%5D=fuzz-packet_parsing_sound&dropdown=coverage#L660

At this point I'm not that familiar with the details of the codebase, so you'll have to look at the results here to see if anything jumps out. It's straightforward to add more fuzzers, and they all run in parallel so CI time does not increase if you do so.

### DIFF
--- a/.github/workflows/checks.yaml
+++ b/.github/workflows/checks.yaml
@@ -270,6 +270,56 @@ jobs:
             cargo fuzz run ${{ matrix.features }} $target -- -max_total_time=10
           done
 
+  fuzz-code-coverage:
+    name: Fuzz with code coverage
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        include:
+          - fuzz_target: packet_parsing_sound
+            corpus: ""
+            features: ''
+          - fuzz_target: record_encode_decode
+            corpus: ""
+            features: ''
+    steps:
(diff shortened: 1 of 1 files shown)


davidv1992:
Thanks for doing this for us. Integration is going to take me a bit more time than I currently can spend, but this is definitely going to come in handy in the future.

rnijveld:
#1902 was merged instead

PR #1877, opened 2025-04-28
Changed kalman offset message to debug level

jpds:
After deploying `ntpd-rs` on my systems, the system logs are peppered with:

```
ntp-daemon[2022]: 2025-04-28T22:16:44.902268Z  INFO ntp_proto::algorithm::kalman: Offset: -9.71569488420898+-5.958433943105912ms, frequency: -0.029784204729893793+-1.5403756385090928pp
```

Which isn't useful to the average system admin, hence reduce this message to `debug`.

### DIFF
--- a/ntp-proto/src/algorithm/kalman/mod.rs
+++ b/ntp-proto/src/algorithm/kalman/mod.rs
@@ -143,7 +143,7 @@ impl<C: NtpClock, SourceId: Hash + Eq + Copy + Debug> KalmanClockController<C, S
         );
 
         if let Some(combined) = combine(&selection, &self.algo_config) {
-            info!(
+            debug!(
                 "Offset: {}+-{}ms, frequency: {}+-{}ppm",
                 combined.estimate.offset() * 1e3,
                 combined.estimate.offset_variance().sqrt() * 1e3,


rnijveld:
I do think that the messages themselves might a little too spammy though: the formatting isn't great (way to many digits, most of them realistically completely useless information) and the number of messages might be too high especially when you have many sources (even more so with a low max poll interval) given that we get these messages most of the time when we receive a new packet from a source.

davidv1992:
Perhaps the better solution here is to make those jumps a warning. Thinking about this, that makes a certain amount of sense to me as it is something that may cause issues (not all software likes the system time jumping around) although it can be unavoidable. What do you think @rnijveld

rnijveld:
Sorry for the slow response, we've been discussing this a little internally, but haven't quite figured out a way forward yet. We do agree that the number of messages is a concern at the very least, but we also think that just moving this log message to a lower level isn't the way forward. I've summed up our current considerations in #1912, if you have any ideas we'd love to hear them in there, but I'll close this PR for now until we've found a good solution going forward. Thanks for all the effort you've put into this so far though!

Issue #1599, opened 2024-08-08
Metrics exporter documentation


cikzh:
We should improve the documentation around the metrics exporter. 
For example:
- [x] how to change the address the exporter will listen to
- [ ] In-depth usage guidance for setup.

PR #1878, opened 2025-04-28
docs: Added reference to metrics-exporter-listen option

jpds:
Fixes: #1599


cikzh (approved):
Thank you for your contribution! This is definitely an improvement, but the issue is not ready to be closed yet. We still want to make additional improvements to this section of the documentation.

PR #1879, opened 2025-04-29
metrics: Drop high cardinality 'id' label from metric collection

jpds:
This creates a new time series record for every single request ID that's sent to every NTP server: https://prometheus.io/docs/practices/naming/#labels

### DIFF
--- a/ntpd/src/metrics/mod.rs
+++ b/ntpd/src/metrics/mod.rs
@@ -101,7 +101,6 @@ macro_rules! collect_sources {
             let labels = vec![
                 ("name", $ident.name.clone()),
                 ("address", $ident.address.clone()),
-                ("id", format!("{}", $ident.id)),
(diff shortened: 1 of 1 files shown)


rnijveld:
Are you getting changed ids every poll? These ids should be our internal source ids, they should not change unless your sources change (i.e. if a source in a pool isn't reachable anymore you'll get a new one), and should be almost as low as the address and name in cardinality.

jpds:
Yes, I am getting new IDs if the source cannot be reached.

rnijveld:
Hmm interesting! I'd agree with your assessment here, it is a likely cause. You'll probably also find in your metrics that you have missed responses from that source as well. We would prefer to resolve the underlying cause here. We'll try and look into it as soon as we have some time. For now I'll close this PR.

PR #1887, opened 2025-05-15
Fix servers root dispersion not properly updating.

davidv1992:


davidv1992:
@squell added the comment.

rnijveld (approved):

PR #1889, opened 2025-05-15
Configure ntpv5 at runtime for server and client (and remove the ntpv5 feature flag)

rnijveld:
- Removes the ntpv5 feature flag
- Uses a custom deserialize for `ProtocolVersion` for the `ntp-version` field of the source config. This value can be set to `4`, `5` and `"auto"`, this last variant maps to `V4UpgradingToV5` with a default number of tries. It is set to `4` by default.
- `NtpVersion` is now returned from the version function of a packet, this did require adding the `V3` version (given that we can handle those versions); this made the server filter based on NTP version relatively simple.
- Server filtering can be configured using the `accept-ntp-versions` config flag (as an array of NTP versions as numbers), it is set to `[3, 4]` by default.


ntp-proto/src/nts_record.rs
```diff
@@ -217,33 +254,23 @@ impl NtsRecord {
 
     #[allow(unused_variables)]
```
davidv1992, line 255: Is this one still necessary?

ntp-proto/src/nts_record.rs
```diff
@@ -217,33 +254,23 @@ impl NtsRecord {
...
     pub fn client_key_exchange_records(
-        ntp_version: Option<NtpVersion>,
+        ntp_version: ProtocolVersion,
         denied_servers: impl IntoIterator<Item = String>,
     ) -> Box<[NtsRecord]> {
         let mut base = vec![
-            #[cfg(feature = "ntpv5")]
             NtsRecord::DraftId {
                 data: crate::packet::v5::DRAFT_VERSION.as_bytes().into(),
             },
```
davidv1992, line 263: This record should only be included if we ask for ntpv5

ntp-proto/src/nts_record.rs
```diff
@@ -1450,7 +1457,6 @@ impl KeyExchangeServerDecoder {
                 Continue(state)
             }
             NextProtocol { protocol_ids } => {
-                #[cfg(feature = "ntpv5")]
                 let selected = if state.allow_v5 {
```
davidv1992, line 1460: An additional check here is needed if the server configuration allows NTPv5

ntp-proto/src/packet/v5/extension_fields.rs
```diff
@@ -5,6 +5,7 @@ use crate::packet::ExtensionField;
 use std::borrow::Cow;
 use std::convert::Infallible;
 
+#[allow(dead_code)]
```
davidv1992, line 8: Why is this needed?

ntp-proto/src/packet/v5/extension_fields.rs
```diff
@@ -21,6 +22,7 @@ pub enum Type {
 }
 
 impl Type {
+    #[cfg(test)]
```
davidv1992, line 25: And why is this needed?

ntp-proto/src/packet/v5/mod.rs
```diff
@@ -39,12 +37,12 @@ impl NtpMode {
         self as u8
     }
 
-    #[allow(dead_code)]
+    #[cfg(test)]
```
davidv1992, line 40: Why is this and the next not used?

docs/man/ntp.toml.5.md
```diff
@@ -133,6 +133,14 @@ sources.
     the number of seconds (i.e. two to the power of the interval). The default
     value of 4 results in an interval of 16 seconds.
 
+`ntp-version` = `4` | `5` | `"auto"` (**4**)
+:   Which NTP version to use for this source. By default this uses NTP version
+    4. You can use `5` to set the protocol version to the draft NTPv5
+    specification, but your source must support the same draft. To use the
+    experimental upgrade protocol set the NTP version to `"auto"`. This uses a
+    specially constructed NTPv4 packet to attempt an upgrade to the supported
+    draft NTPv5 version if the source also supports the same draft version.
```
davidv1992, line 142: Perhaps we should add that NTPv5 support is still considered in beta, and that things might not be backwards compatible with future versions.

docs/man/ntp.toml.5.md
```diff
@@ -198,6 +206,12 @@ time.
     `DENY` kiss code. When set to `false` (the default), normal NTP messages are
     also allowed.
 
+`accept-ntp-versions` = [ `3` | `4` | `5`, .. ] (**[3, 4]**)
+:   An array of NTP versions that are accepted by the server. By default only
+    NTPv3 and NTPv4 packets are supported. To also allow draft NTPv5 packets,
+    you can set this value to `[3, 4, 5]`.
```
davidv1992, line 212: Perhaps we should add that NTPv5 support is still considered in beta, and that things might not be backwards compatible with future versions.

davidv1992 (changes requested):
Missing checks in various parts of the key exchange, and a bit of warnings in the docs, but other than that this looks solid.

ntp-proto/src/nts_record.rs
rnijveld: Fixed in eff750520a38376b666d24858b8630cdd66fb8be

ntp-proto/src/nts_record.rs
rnijveld: It is, but I've narrowed the attr using cfg-attr in d58d7086c70012189c25a9a869d0c5715cd03191

ntp-proto/src/packet/v5/extension_fields.rs
rnijveld: This function is only used in tests, in normal usage we just use the `fn from_type_id` in the regular `extension_fields.rs`, so I do think we probably need to do some more cleanup here, but I considered that out of scope for this PR (it is big enough already)

ntp-proto/src/packet/v5/mod.rs
rnijveld: I've only seen usages in test code, which is why I replaced the `allow(dead_code)` with `cfg(test)`. I think we should probably figure out why there is so much apparently unneeded code in the v5 parts of the codebase, but I would propose we do this in another PR to keep this one somewhat reasonably sized.

docs/man/ntp.toml.5.md
rnijveld: Fixed in a3c075476f93ee85b1735790d2ac90df37f2d190

docs/man/ntp.toml.5.md
rnijveld: Fixed in a3c075476f93ee85b1735790d2ac90df37f2d190

davidv1992 (approved):

PR #1905, opened 2025-06-14
Update CHANGELOG.md

Joysephk:
just a dot adder

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,38 +3,38 @@
 ## [1.5.0] - 2025-02-28
 
 ### Added
-- Support for PPS based sources (this support can be disabled with a compile flag for now)
-- Per source configuration of poll intervals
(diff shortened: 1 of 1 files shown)


rnijveld:
Hi! Thanks for doing this, but given that our changelog format is based on https://keepachangelog.com, and even on their website they aren't very strict with this, I can't really justify us being very strict with this. Different people release the software, and this is one of those things that just happens and would take extra time and tooling to prevent. Investing too much time in that is hard to justify with all the other stuff we also could be doing.

I'd also prefer not to change the changelog for older entries, instead treating the file as mostly append-only (or maybe prepend only given we add new entries at the top :sweat_smile:). But I'll try to make sure that future releases are a little more consistent in their changelog format! Very glad someone is at least reading the changelog file and noticing these details!

PR #1911, opened 2025-06-18
Rework nts

davidv1992:


ntp-proto/src/nts/messages.rs
```diff
@@ -0,0 +1,1849 @@
...
+    },
+    #[cfg(feature = "nts-pool")]
+    Support {
+        wants_protocols: bool,
+        wants_algorithms: bool,
+    },
+}
+
+impl Request<'_> {
+    pub async fn parse(mut reader: impl AsyncRead + Unpin) -> Result<Self, NtsError> {
```
rnijveld, line 37: I don't really like how different this function is with or without the nts-pool cfg, but I don't really have an alternative either. Maybe you can put the section after the loop in its own function?

davidv1992: The differences between with or without pool are quite unavoidable. As for the length of the function, I'll have a think about refactoring this further to reduce that, but I don't think splitting the loop and the part after is the right way to go given the amount of data flowing between the two. But let's do that after merging this.

PR #1968, opened 2025-09-22
Fix compilation failure with "pps" feature disabled

decathorpe:
The `NtpSourceConfig::Pps` enum variant is referenced twice in `ntpd/src/daemon/config/mod.rs`, but only properly cfg-gated `#[cfg(feature = "pps")]` once. This PR fixes this issue and the compilation failure if the "pps" feature is not enabled.

See also https://github.com/pendulum-project/pps-time/issues/1 where it was suggested to build ntpd-rs to avoid potential issues with the pps-time crate.


davidv1992:
Thanks for spotting this and providing a fix.

PR #1982, opened 2025-10-01
Implemented token authentication for nts pool.

davidv1992:


rnijveld (commented):
Looks pretty good to me. One thing does bother me a little: calling it 'authentication'/'authentications' (in variable names and config etc) whereas they are more authentication tokens or keys. I would consider authentication more of an action you do and less of a thing you have.

rnijveld (approved):

PR #2007, opened 2025-10-28
Cap control socket payload size

lblommesteyn:
## Summary
- cap the control/metrics Unix socket JSON payload size at 1 MiB so untrusted clients can’t force arbitrarily large allocations
- add a checked `u64 → usize` conversion so bogus length prefixes fail fast with `InvalidInput`
- cover the new behaviour with `oversized_messages_are_rejected`

## Testing
- cargo +1.88.0 test -p ntpd oversized_messages_are_rejected


davidv1992:
Thank you for providing this. Would it be possible to fix the formatting issues identified by the CI?

PR #2013, opened 2025-10-31
Implement seccomp and drop unused capabilities.

ghost:
Hey! This PR Implements seccomp syscall filtering and dropping of unused capabilities. 

Before a tool runs, firstly uneeded capabilities are dropped. This Is already done In the systemd service file of ntpd-rs and ntp-metrics-exporter. This Is particularly useful for non-systemd distributions. Then the seccomp filter Is initialized with the default action being to kill the process (by SIGSYS) If a bad syscall Is made. Syscalls used by all three programs are initialized in the SHARED_SYSCALL array, syscalls which are unique to each tool are additionally passed to the function during initialization. In seccomp_init() these two are concatted together and allowed. I did it this way so that there Isn't repetitive code. Since all of this Is well managed In one file, implementing landlock In the future should be easier.

### DIFF
--- a/ntpd/src/security.rs
+++ b/ntpd/src/security.rs
@@ -67,11 +67,12 @@ pub(crate) fn drop_caps(needed: Option<&[Cap]>) {
     current.inheritable = capset.clone();
     current.effective = capset.clone();
     if let Err(e) = current.set_current() {
-        eprintln!("Failed to set current capabilities: {:?}", e);
-        process::exit(1);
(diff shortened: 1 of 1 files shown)


davidv1992:
Thank you for the contribution. For ease of review and merging, would it be possible for you to split the PR. Seccomp syscall filtering is notoriously finicky, with parts of it depending on library versions and such. So I would prefer to handle merging that separately.

Edit: I mean splitting it into a separate PR for the seccomp stuff and a separate PR for the capability control stuff.

PR #2020, opened 2025-11-05
Keep alive support

davidv1992:
This patch set implements keep alive support for ntpd-rs.

The final design restricts longlived connections to a small number through a separate permit pool. Several callbacks have been added to the ntp-proto key exchange server to facilitate dealing with these permits, and to assist with keyset updates during the life of the longlived connections.

These callbacks are deliberatly not async functions, as they should complete quickly, and, in the case of the callback giving a permit, should fail if they cannot be completed quickly.

### DIFF
--- a/ntp-proto/src/nts/mod.rs
+++ b/ntp-proto/src/nts/mod.rs
@@ -480,6 +480,8 @@ impl KeyExchangeServer {
         mut io: tokio_rustls::server::TlsStream<T>,
         mut get_keyset: impl FnMut() -> U,
     ) -> Result<(), NtsError> {
+        tracing::debug!("Longterm handling started for connection");
+
         loop {
             let request = match Request::parse(&mut io).await {
                 Ok(request) => request,
@@ -527,6 +529,8 @@ impl KeyExchangeServer {
                     protocol,
                     keep_alive,
                 } => {
+                    tracing::debug!("Received fixedkey request on longterm connection");
+
                     let cookie = DecodedServerCookie {
                         algorithm,
                         s2c: s2c_key,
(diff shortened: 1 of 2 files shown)


ntpd/src/daemon/config/server.rs
```diff
@@ -191,27 +191,66 @@ impl From<ServerConfig> for ntp_proto::ServerConfig {
...
+        let raw = NtsKeConfigRaw::deserialize(deserializer)?;
+        Ok(NtsKeConfig {
+            certificate_chain_path: raw.certificate_chain_path,
+            private_key_path: raw.private_key_path,
+            accepted_pool_authentication_tokens: raw.accepted_pool_authentication_tokens,
+            key_exchange_timeout_ms: raw.key_exchange_timeout_ms,
+            concurrent_connections: raw.concurrent_connections,
+            longlived_connections: raw
+                .longlived_connections
+                .unwrap_or(5.min(raw.concurrent_connections / 10)),
```
rnijveld: I'm not sure I agree with doing it this way, it feels like a very opaque way of doing it (even if documented properly)

docs/man/ntp.toml.5.md
```diff
@@ -301,9 +301,16 @@ not tampered with.
     is exceeded the connection will be dropped.
 
 `concurrent-connections` = *number* (**512**)
-:   Maximum number of concurrent connections the key exchange server will handle.
+:   Maximum number of total concurrent connections the key exchange server will handle.
     Any connections above the threshold will be held in an OS level queue.
 
+`longlived-connections` = *number* (**5**, see text)
+:   Maximum number of concurrent longlived connections from nts pools the key
+    exchange server will handle. The default value is the smaller of 5
```
rnijveld: I cannot succesfully parse the next sentence such that it makes sense to me.

rnijveld (changes requested):

rnijveld (approved):

PR #2034, opened 2025-11-27
Fix Doc example location

RootUser-GeetanshAditya:
In the installation page, there was a small error on location of the example config file. It is now placed in docs/examples/cfg instead of just docs/examples


davidv1992:
Thank you for noticing and providing the fix.

PR #2039, opened 2025-12-04
Correct warn_on_jump to warn-on-jump in log message

arp242:
I got this log message:

	Jumped offset by 605.9385589814515ms. This may cause problems
	for other software. If this is not a problem for your system,
	you can reclassify this warning as an informative message
	through the `synchronization.warn_on_jump` setting in ntp.toml.

But it won't work with underscores: it needs to be kebab-case.


davidv1992 (approved):

davidv1992:
Woeps, thank you for noticing.

PR #2079, opened 2026-01-15
Explicitly allow redundant_else clippy lint.

davidv1992:
This is motivated by the fact that I prefer the clarity of either A or B happens in the situations where this lint triggers. The tradeoff of slightly more indentation is in my opinion not problematic.

Note: Commit message is formulated to reflect that this is shared consensus after accepting this PR


cikzh (dismissed):
I checked the code this lint triggers and I agree that the explicit else blocks improve the code. Maybe it's wise @rnijveld also agrees before merging

rnijveld:
I don't really see the issue, its only a couple of tiny changes, most of them make a lot of sense to me, see this commit: 0d68970113b5c1dbe117ed63992492101c46c2a3

PR #2126, opened 2026-02-18
Reorganization of communication within the daemon

davidv1992:


ntpd/src/force_sync/mod.rs
```diff
@@ -1,4 +1,4 @@
-use std::{
+/*use std::{
```
squell, line 1: ?

davidv1992: This temporarily disables the force sync subcommand until we have everything quieted down a bit.

rnijveld (dismissed):

rnijveld (approved):

PR #2133, opened 2026-02-27
Update certificates for another year

rnijveld:


vcunat:
/cc ntpd-rs maintainers in nixpkgs: @fpletz @getchoo
(It's unclear to me how exactly solve this in there, too.)

PR #2139, opened 2026-03-03
Various CI improvements suggested by zizmor

bjorn3:
https://zizmor.sh/ is a static analysis tool for github actions. @hashcatHitman suggested using this.

### DIFF
--- a/.github/workflows/packaging.yaml
+++ b/.github/workflows/packaging.yaml
@@ -127,7 +127,7 @@ jobs:
           test "$release_notes_version" == "${release_version}"
           echo "$release_notes_body" > release_notes.md
 
-      - name: Create release
+      - name: Create a draft release
         env:
           GH_TOKEN: ${{ github.token }}
         run: |


.github/workflows/packaging.yaml
```diff
@@ -21,10 +20,8 @@ jobs:
           - i686-unknown-linux-gnu
     steps:
       - name: Setup packaging tools for cross compiled artifacts
-        uses: awalsh128/cache-apt-pkgs-action@acb598e5ddbc6f68a970c5da0688d2f3a9f04d05 # v1.6.0
```
davidv1992: @rnijveld was there a specific reason we used caching here?

.github/workflows/packaging.yaml
```diff
@@ -21,10 +20,8 @@ jobs:
...
-        with:
-          packages: musl-tools qemu-user-static crossbuild-essential-armhf crossbuild-essential-arm64 crossbuild-essential-i386
-          version: 1
+        run: |
+          sudo apt install --update musl-tools qemu-user-static crossbuild-essential-armhf crossbuild-essential-arm64 crossbuild-essential-i386
```
davidv1992: I think we should call `sudo apt update` before the install here to ensure this keeps functioning, even if the OS image is somewhat out of date.


```suggestion
        run: |
          sudo apt update
          sudo apt install --update musl-tools qemu-user-static crossbuild-essential-armhf crossbuild-essential-arm64 crossbuild-essential-i386
```

PR #2189, opened 2026-04-08
Fix codeql-action/upload-sarif version number being off.

davidv1992:


squell (approved):
I have confirmed that this is the correct version number for the hash: https://github.com/github/codeql-action/releases/tag/v4.35.1

PR #2192, opened 2026-04-09
refactor: move entry points to cargo standard paths

### LINKED ISSUE #2176: Consider moving entry points of binaries to standard location
The [standard layout](https://doc.rust-lang.org/cargo/guide/project-layout.html) will compile binaries in the right location, which currently would just mean moving the `bin` folder into `src`.

If you want, you could remove the `[[bin]]` settings from `Cargo.toml` as well.

idealseal:
This makes the project more discoverable by using the standard paths for binary entry points as expected by cargo's
[package layout](https://doc.rust-lang.org/cargo/guide/project-layout.html). It also removes the necessity to specify the files by hand in Cargo.toml.

Closes: https://github.com/pendulum-project/ntpd-rs/issues/2176


davidv1992 (approved):
Thank you

PR #2193, opened 2026-04-09
feat: notify systemd about service startup

idealseal:
Closes: https://github.com/pendulum-project/ntpd-rs/issues/2177

### DIFF
--- a/ntpd/src/daemon/tracing.rs
+++ b/ntpd/src/daemon/tracing.rs
@@ -108,6 +108,7 @@ impl LogReloadTaskStarter {
 
             loop {
                 stream.recv().await;
+
                 let new_file = match std::fs::File::create(&self.path) {
                     Ok(new_file) => new_file,
                     Err(e) => {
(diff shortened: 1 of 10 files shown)


davidv1992 (dismissed):

davidv1992:
Ignore the approval, that was me being a bit dumb trying to approve running the workflows.

ntpd/src/daemon/tracing.rs
```diff
@@ -108,6 +108,17 @@ impl LogReloadTaskStarter {
...
+                #[cfg(target_os = "linux")]
+                {
+                    use sd_notify::NotifyState;
+
+                    let _ = sd_notify::notify(&[
+                        NotifyState::Reloading,
+                        NotifyState::monotonic_usec_now().unwrap(),
+                    ]);
+                }
+
```
davidv1992: This is not really a full reload of configuration. For now we probably shouldn't report this to systemd.

davidv1992 (commented):
Thank you for providing this. It looks really quite nice already, would you be willing to do the rebase it needs and look at the one comment below?

davidv1992 (approved):
Thank you for providing this

PR #2235, opened 2026-04-22
Update rustls-webpki to 0.103.13

davidv1992:
This is necessary because of a (for us luckily unreachable) panic (RUSTSEC-2026-0104)


squell (approved):

xtqqczze:
I've submitted https://github.com/pendulum-project/ntpd-rs/pull/2237 to more proactive about RUSTSEC advisories.

PR #2236, opened 2026-04-22
Use package manager frontends instead of the lower-level tools for installing packages.

davidv1992:
Closes #2186


squell:
I didn't know you could install .deb files using `apt`. Did you verify that this works?

squell (approved):

PR #2238, opened 2026-04-22
Update ntpv5

### LINKED ISSUE #2092: Bump from NTPv5 draft 6 to draft 7
https://datatracker.ietf.org/doc/html/draft-ietf-ntp-ntpv5-07

davidv1992:
Bunch of minor fixes to packet parsing, mostly around ntpv5. Also bumps to the latest draft.

Closes #2092

### DIFF
--- a/ntp-proto/src/packet/v5/server_reference_id.rs
+++ b/ntp-proto/src/packet/v5/server_reference_id.rs
@@ -2,7 +2,6 @@ use crate::packet::v5::NtpClientCookie;
 use crate::packet::v5::extension_fields::{ReferenceIdRequest, ReferenceIdResponse};
 use rand::distributions::{Distribution, Standard};
 use rand::{Rng, thread_rng};
-use std::array::from_fn;
 use std::fmt::{Debug, Formatter};
 
 #[derive(Copy, Clone, Debug)]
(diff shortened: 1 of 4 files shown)


ntp-proto/src/packet/v5/server_reference_id.rs
```diff
@@ -48,10 +47,18 @@ pub struct ServerId([U12; 10]);
 impl ServerId {
     /// Generate a new random `ServerId`
     pub fn new(rng: &mut impl Rng) -> Self {
-        // FIXME: sort IDs so we access the filters predictably
-        // FIXME: check for double rolls to reduce false positive rate
-
-        Self(from_fn(|_| rng.r#gen()))
+        loop {
```
squell, line 53:
```suggestion
        // To make optimal use of the Bloom filter used for loop detection, which we index with 10 12-bit strings derived from our identifier,
        // we require that the server ID consist of 10 unique 12bit identifiers. For optimization of access patterns, it's also beneficial to
        // have these sorted anyway, which makes this check cheap.
        loop {
```

squell (dismissed):

ntp-proto/src/packet/v5/server_reference_id.rs
squell: Sadly I can't quickly find a standardized taxonomy to describe a "k-times-n bit Bloom filter" or some such, so this needs to be a bit more verbose than I'd ideally like

rnijveld (approved):

PR #2240, opened 2026-04-22
clippy: fix enum_glob_use lint

xtqqczze:
https://rust-lang.github.io/rust-clippy/rust-1.95.0/index.html#enum_glob_use

### DIFF
--- a/ntp-proto/src/lib.rs
+++ b/ntp-proto/src/lib.rs
@@ -28,7 +28,7 @@
 //FIXME: Enable #![warn(clippy::doc_markdown)]
 #![warn(clippy::elidable_lifetime_names)]
 #![warn(clippy::empty_enum)]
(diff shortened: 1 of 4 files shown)


davidv1992:
Hmm, I think this is one where we should make the decision to not include it in the big list. I'm not sure I like the impact this one has on our code base.

xtqqczze:
I think that the lint’s suggested fix should favor using the fully qualified enumeration variant name instead of importing the variants into scope, see https://github.com/rust-lang/rust-clippy/issues/16900.

davidv1992 (approved):

davidv1992:
@xtqqczze I didn't realize you had already made the changeover to the other type of fix. This way indeed looks much better and I would like to merge this. Are there reasons you still kept it as draft?

davidv1992:
Thank you for providing this :)

PR #2278, opened 2026-05-28
Make dependency on AWS-LC optional and allow OpenSSL as an alternative

### LINKED ISSUE #2223: Support openSSL as crypto provider for TLS
This is desired by some distributions, as they then only have to manage one set of crypto primitives.

squell:
This is achieved in three steps (hence three commits)

- First, the hickory resolver is gated behind a feature flag. It has a hard dependency on AWS-LC or `ring` (but using that is not an option)
   - I note with interest that all the unit tests keep working if the `srv` feature flag is disabled; I don't think it's the job of this PR to do anything about that.
   
- Remove the use of `aws-lc` as the CryptoProvider. After this step, one can still (surprising to me) compile everything, but the moment you try to play around with TLS connections without the `aws-lc` feature you hit a `panic!` inside RusTLS.

- Add a gated dependency on `rustls_openssl` and install it very early as the default cryptoprovider.

### DIFF
--- a/ntpd/src/daemon/dns.rs
+++ b/ntpd/src/daemon/dns.rs
@@ -19,17 +19,30 @@ pub(crate) struct KeResolutionResult {
     pub(crate) srv_record_name: Option<String>,
 }
 
+#[cfg(not(feature = "srv"))]
+pub(crate) async fn resolve_ke(
+    addr: &NormalizedAddress,
+) -> Result<impl Iterator<Item = KeResolutionResult>, std::io::Error> {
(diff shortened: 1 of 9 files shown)


ntpd/src/daemon/dns.rs
rnijveld, line 37: Personally not an amazing fan of cfg-ing away almost the entire body of a function, maybe create this function twice, once for the SRV variant, and once without (looks like the variant without is only a single expression anyway) and add the cfg on that level?

squell: 🤩

rnijveld (approved):

PR #2287, opened 2026-06-04
Openssl crypto

### LINKED ISSUE #2224: Support openSSL as alternative to some of the rust-crypto crates we use.
Reduces complexity for some packagers.

squell:
Notes:

- MD5 in OpenSSL can return an error (!?); I've made this into an unwrap since any error handling would also result in ntpd-rs exiting.

- First other few commits decouple some of the RustCrypto details from `packet/crypto.rs`

- `crypto/openssl_defs.rs` contains some helper definitions that will allow a lot of source-level code sharing between the Rustcrypto and OpenSSL side by mimicking some types. I do not use `GenericArray` here because I don't like that particular construction. Sadly Rust doesn't allow sizes of types to depend on trait constants but I found a way around that.

### DIFF
--- a/ntp-proto/src/packet/crypto/openssl_defs.rs
+++ b/ntp-proto/src/packet/crypto/openssl_defs.rs
@@ -1,0 +1,67 @@
+//! This file contains file types that mimic those in RustCrypto, so *at a source level* code sharing can be maximized.
+
+use std::ops::Deref;
+
+#[derive(Debug, Clone, Copy, Default)]
+pub struct Aes128Siv;
+
+#[derive(Debug, Clone, Copy, Default)]
+pub struct Aes256Siv;
+
+#[repr(transparent)]
+pub struct Key<T>(<Key<T> as Deref>::Target)
+where
+    Key<T>: Deref;
+
+impl Deref for Key<Aes128Siv> {
+    type Target = [u8; 32];
+    fn deref(&self) -> &Self::Target {
+        &self.0
+    }
+}
+
+impl Deref for Key<Aes256Siv> {
+    type Target = [u8; 64];
+    fn deref(&self) -> &Self::Target {
+        &self.0
+    }
+}
+
+impl Default for Key<Aes128Siv> {
+    fn default() -> Self {
+        Self([0; _])
+    }
+}
+
+impl Default for Key<Aes256Siv> {
+    fn default() -> Self {
+        Self([0; _])
+    }
+}
+
+impl<T> FromIterator<u8> for Key<T>
+where
+    Key<T>: Deref + Default,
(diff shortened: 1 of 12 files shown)


ntp-proto/src/packet/crypto/openssl_defs.rs
```diff
@@ -0,0 +1,194 @@
...
+pub struct Aes128Siv;
+
+#[derive(Debug, Clone, Copy, Default)]
+pub struct Aes256Siv;
+
+impl SSLName for Aes128Siv {
+    // NOTE: the argument is used to prevent coding mistakes at compile time
+    fn name() -> &'static str {
+        "AES-128-SIV"
+    }
```
davidv1992: There is no argument, what is the comment trying to say?

ntp-proto/src/packet/crypto/openssl_defs.rs
```diff
@@ -0,0 +1,194 @@
...
+impl Default for Key<Aes256Siv> {
+    fn default() -> Self {
+        Self([0; _])
+    }
+}
+
+#[repr(transparent)]
+pub struct Key<T>(<Key<T> as Deref>::Target)
+where
+    Key<T>: Deref;
```
davidv1992: This construct is a bit funky. Perhaps we should comment what is going on here and why.

davidv1992 (changes requested):
Mostly looks good. Two minor things around the new openssl specific traits.

Also, this will need some small changes to the CI created in #2294

ntp-proto/src/packet/crypto/openssl_defs.rs
squell: That is a stray, there used to be an argument here to have a double-check that keysize and cipher are matched but I removed it as we only need it in one place.

ntp-proto/src/packet/crypto/openssl_defs.rs
squell: Actually in the end I did need a custom trait so I think I can simplify this code and make it more pedestrian. Sad. I like leaving funky nuggets in production source code bases... :(

davidv1992 (dismissed):

davidv1992:
@squell changes look good, but can't merge because of the conflict. Can you do a quick rebase and regenerate the cargo.lock?

squell:
> @squell changes look good, but can't merge because of the conflict. Can you do a quick rebase and regenerate the cargo.lock?

Rebased and fixed the Cargo.lock file

davidv1992 (dismissed):

squell (commented):
I approve!

rnijveld (dismissed):

davidv1992 (dismissed):

davidv1992 (approved):

PR #2295, opened 2026-06-10
Add missing openssl entry to cargo lock file.

davidv1992:


davidv1992:
Already fixed by dependabot

PR #2302, opened 2026-06-16
Add AppArmor profile

### LINKED ISSUE #2222: Implement apparmor support/provide apparmor example
This needs investigation, but would be nice to have.

squell:
Closes #2222 

Note that `dac_read_search` capability is needed to prevent this denial:

```
apparmor="DENIED" operation="capable" class="cap" profile="pendulum" pid=5342 comm="tokio-rt-worker" capability=2  capname="dac_read_search"
```
I don't really know why tokio-rt-worker triggers this. In any case it's a capability seen in various AppArmor profiles; file access is being very narrowly restricted by the AppArmor config so it should be fine (if no cracks in the armor 😅)

### DIFF
--- a/pkg/apparmor/pendulum
+++ b/pkg/apparmor/pendulum
@@ -1,0 +1,38 @@
+#abi <abi/4.0>,
+
+include <tunables/global>
+
+#NOTE: this should probably be written as an attached profile, in the form of:
+#
+# /usr/sbin/ntp-daemon flags=(enforce) {
+#
+# but for development this profile is easier to work with.
+#
+# Load it using `apparmor_parser -r pkg/apparmor/pendulum`
+# and then apply it using `aa-exec -p pendulum -- <ntpd-rs command>`
+#
+profile pendulum flags=(enforce) {
+  include <abstractions/base>
+  include <abstractions/nameservice>
+  include <abstractions/user-tmp>
+
+  capability sys_time,
+  capability dac_read_search,
+
+  /dev/pps[0-9]*        rw,
+
+  @{PROC}/@{pid}/cgroup r,
+  /sys/fs/cgroup/**/cpu.max  r,
+
+  /etc/ntpd-rs/** r,
+
+  /run/ntpd-rs/**     rw,
+
+  # NOTE: this is for development only, assuming ntpd-rs is checked out in your home folder
+  @{HOME}/ntpd-rs/**   r,
+  @{HOME}/**/ntpd-rs/**   r,
+
+  # NOTE: site-specific additions and overrides; this should probably also be something like
+  # local/home.usr.sbin.ntp-daemon if we turn this into an attached profile
+  # include <local/pendulum>
+}
--- a/pkg/apparmor/pendulum
+++ b/pkg/apparmor/pendulum
@@ -17,7 +17,6 @@ profile pendulum flags=(enforce) {
   include <abstractions/user-tmp>
 
   capability sys_time,
-  capability dac_read_search,
 
   /dev/pps[0-9]*        rw,
 


pkg/apparmor/pendulum
```diff
@@ -0,0 +1,38 @@
...
+#
+# but for development this profile is easier to work with.
+#
+# Load it using `apparmor_parser -r pkg/apparmor/pendulum`
+# and then apply it using `aa-exec -p pendulum -- <ntpd-rs command>`
+#
+profile pendulum flags=(enforce) {
+  include <abstractions/base>
+  include <abstractions/nameservice>
+  include <abstractions/user-tmp>
```
squell, line 17: Note: this would basically be unnecessary if we don't need to access `/tmp`; I do see that the ntpd-rs metrics exporter does this (and perhaps other things as well)

pkg/apparmor/pendulum
```diff
@@ -0,0 +1,38 @@
...
+#
+# /usr/sbin/ntp-daemon flags=(enforce) {
+#
+# but for development this profile is easier to work with.
+#
+# Load it using `apparmor_parser -r pkg/apparmor/pendulum`
+# and then apply it using `aa-exec -p pendulum -- <ntpd-rs command>`
+#
+profile pendulum flags=(enforce) {
+  include <abstractions/base>
```
squell, line 15: The `base` abstraction probably contains various things we don't need (but also various we do). Sifting through all that is probably a lot more work; I notice the `usr.sbin.ntpd` also uses this abstraction so it's probably fine *for now*.

pkg/apparmor/pendulum
```diff
@@ -0,0 +1,38 @@
...
+  include <abstractions/nameservice>
+  include <abstractions/user-tmp>
+
+  capability sys_time,
+  capability dac_read_search,
```
davidv1992: Hmmm, I know this somehow is needed for tokio, but it feels very strange that this is something that is needed. Best I can tell from the AppArmor documentation, these capabilities map onto the regular linux kernel capabilities, and we don't need this without AppArmor.

Given that, I would love to better understand this before we blindly add this as needed to the profile.

Looking at the file rights below, my best guess is that we are incorrectly not indicating that it should have directory listing rights, which might be what this capability is allowing us to then avoid needing.

davidv1992 (commented):
Mostly looks good, eventually we want to use this in our own packages as well I would guess, but we can do that integration later in a separate PR.

pkg/apparmor/pendulum
rnijveld: This could also be caused by tokio renaming worker threads to `tokio-rt-worker`, which requires filesystem access to `/proc/self/task/*/comm`.

squell: Note, it only occurs if using ntpd-rs as a server:

```
Jun 29 14:21:21 fluorine kernel: audit: type=1400 audit(1782735681.188:4454): apparmor="DENIED" operation="capable" class="cap" profile="pendulum" pid=9536 comm="tokio-rt-worker" capability=2  capname="dac_read_search"
```

squell: Most other errors leave a more clear error in the log, which led to e.g. the explicit permissions for /sys/fs/cgroup/**/cpu.max etc. Even giving ntpd-rs full permissions on the entire file system isn't solving the issue for dac_read_search though

squell: After a bit of fiddling: it's only necessary if ntpd-rs is run as root. If I just run it with `setcap cap_sys_time,cap_net_bind_service+ep` it runs fine within the pendulum profile.

squell: Alright, problem solved; without dac_read_search root simply didn't have permissions to open the file and so needed to fallback to the classic "root can do whatever the hell it likes". That causes the dac_read_search or dac_override request.

Since we don't need it (we assume permissions are configured correctly) I've removed it.

Hat tip to @bjorn3 for putting in the final puzzle piece.

davidv1992 (approved):

Issue #2210, opened 2026-04-16
Implement socket management for CSPTP and PTP


davidv1992:
Because CSPTP must use port 319 and 320 also for sending requests, and these ports are also needed for PTP, we cannot use one socket per CSPTP source or server. Therefore, we need something central to manage the UDP sockets for these ports and the traffic over them.

Issue #1880, opened 2025-04-30
Support for multiple IP addresses per interface and multihoming


demsbjf8:
**Multiple IP addresses per interface**
Imagine ntpd-rs as a server on a machine that has multiple IP addresses on one interface. If a request comes in for an IP address that is not the "primary"* one, but a "secondary" one, the outgoing response should have that secondary IP address as source address that was the destination IP address on the incoming request. 

Instead, as far as I could see so far for IPv6, the response will have the "primary" IPv6 address as source IP address

**Multihoming (multiple interfaces)**
Imagine ntpd-rs as server on a machine that has two (or more) interfaces in different subnets, e.g., a router. Now, when a packet comes in on one interface, but is destined to the IP address of another interface, the response should have that other interface's address as source. I.e., the destination address on the incoming request should be used as source address on the outgoing response.

Instead, as far as I could see so far, the outgoing response will have the IPv6 address of the outgoing interface as source address, not the one that was the destination IP address of the incoming request. 

**Current issue with both multiple IP addresses per interface as well as multihoming**
Some clients seem to still accept the response somehow, e.g., NTPsec's ntpdate implementation (based on ntpdig, I think). Many are not, and actually should not. 

At the latest when the client is behind a NAT device, or stateful firewall, those will drop the response packets as they don't match the state created from the request packet. So the client will not even see the response packet, and will again not be able to sync to the server.

**Potential solution**
ntpd-rs as UDP server should explicitly remember the destination address on an incoming request packet, and explicitly set that as source address on the corresponding outgoing response packet. Right now, it seems the choice of IP address for outgoing responses is left to the OS. Which is fine for outgoing requests, but not for outgoing responses.

This is with ntpd-rs v. 1.5.0-1 on Ubuntu 24.04 on AMD64, installed from your release's artifacts. I currently am using IPv6 only, so not sure whether similar issues are present for IPv4 as well.

If the above is not clear, I could try to come up with a picture.

* With IPv4 on Linux, the system makes the distinction between a primary address and secondary addresses explicit, e.g., in the output of `ip -4 address show`. With IPv6, Linux will typically pick the last address that was added to the interface. But depending on _how_ the addresses were configured, this may not be easily visible to a user.

squell:
Closed by #2320

PR #2317, opened 2026-07-01
Update timestamped-socket to 0.3.0

davidv1992:
This updates timestamped socket to the new API, which is needed for #2210 as well as the solution to #1880. In draft until we publish a definitive version of timestamped-socket 0.3.0


davidv1992:
Apologies, accidentally merged #2320 into this, so that is now also included.

rnijveld (approved):

PR #2330, opened 2026-07-08
Cleanup several clippy lints

stefanvi:
This PR uncomments and resolves several clippy lints found in `lib.rs` files throughout the project. I put each lint in a separate commit, so that they can be dropped easily if not desirable and they can be easily reviewed in github by walking through the commits.

Some notable changes:
- The [`ip_constant`](https://rust-lang.github.io/rust-clippy/master/index.html#ip_constant) lint exists from version 1.89, which is greater than the MSRV (1.88). I commented this out.
- I resolved [`unchecked_duration_subtraction`](https://rust-lang.github.io/rust-clippy/master/index.html#unchecked_time_subtraction) using a saturating sub in [`ntpd/src/daemon/spawn/mod.rs`](https://github.com/pendulum-project/ntpd-rs/blob/ef16c2277cd0efafdb3712e43b4fed5962a66ed2/ntpd/src/daemon/spawn/mod.rs#L251)


ntpd/src/daemon/clock.rs
```diff
@@ -42,7 +42,7 @@ impl NtpClock for NtpClockWrapper {
         let (seconds, nanos) = offset.as_seconds_nanos();
         self.0
             .step_clock(TimeOffset {
-                seconds: seconds as _,
+                seconds: seconds.into(),
```
stefanvi: I see now that this fails on 32 bit. Maybe we can do (and/or check `target_pointer_width`):

```rust
let (seconds, nanos) = offset.as_seconds_nanos();
#[cfg(not(target_arch = "aarch64"))]
let seconds = seconds.into();
self.0
    .step_clock(TimeOffset { seconds, nanos })
    .map(convert_clock_timestamp)
```

davidv1992: Perhaps for now separate the commit causing this into a separate commit, then we can merge the rest. Also, would it be an option to do the cargo fmt on each individual commit, instead of once at the end?

stefanvi: Done and done! (bless interactive rebase)

ntpd/src/daemon/ntp_source.rs
```diff
@@ -531,11 +531,13 @@ mod tests {
...
 
-    async fn test_startup<T: Wait>() -> (
+    type ProcessInfo<T> = (
         SourceTask<TestClock, TwoWaySourceControllerWrapper<TwoWayKalmanSourceController>, T>,
         Socket<SocketAddr, Open>,
         mpsc::Receiver<MsgForSystem>,
-    ) {
+    );
+
+    fn test_startup<T: Wait>() -> ProcessInfo<T> {
```
davidv1992: I think we should do an exception to the return type lint that triggers here instead of adding the custom type in testing code.

davidv1992 (commented):
Please remove (and make separate prs where relevant) the following commits:
- clippy::enum_globs (redundant with earlier PR)
- clippy::unnecessary_wraps (needs discussing)
- clippy::if_not_else  (needs discussing)
- Fix clippy warnings on MSVR

And please combine the two clippy::ignored_unit_patterns commits.

In general, please check if you enabled a lint in one crate, that it now is enabled in all crates.

stefanvi:
I have separated the relevant commits, replaced the `ProcessInfo` complex return type with an exception and checked that lints are present in all `lib.rs` files. For `cast_precision_loss` this was already not the case before this PR. I had initially added a commit that added this lint to three out of four packages, but the fourth (`ntp-proto`) proved to be difficult. I decided to drop this commit as well

davidv1992 (approved):

PR #2332, opened 2026-07-08
Add clippy::cast_lossless

stefanvi:
The `clippy::cast_lossless` lint was originally included in #2330, but deserves a more thorough look. The lint suggests the following to convert an i32 to an i64:

```diff
self.0
    .step_clock(TimeOffset {
-        seconds: seconds as _,
+        seconds: seconds.into(),
        nanos,
    })
```

However, on 32 bit systems, the `seconds` are also an i32 and thus clippy complains that the call to `.into()` is a noop. I propose gating the `into()` call behind a `target_arch` feature.

Edit: `target_arch` still caused issues in CI, so I switched to `target_pointer_width`

### DIFF
--- a/ntpd/src/daemon/clock.rs
+++ b/ntpd/src/daemon/clock.rs
@@ -40,11 +40,10 @@ impl NtpClock for NtpClockWrapper {
         offset: ntp_proto::NtpDuration,
     ) -> Result<ntp_proto::NtpTimestamp, Self::Error> {
         let (seconds, nanos) = offset.as_seconds_nanos();
+        #[cfg(not(target_pointer_width = "32"))]
+        let seconds = seconds.into();
         self.0
-            .step_clock(TimeOffset {
-                seconds: seconds as _,
-                nanos,
-            })
+            .step_clock(TimeOffset { seconds, nanos })
             .map(convert_clock_timestamp)
     }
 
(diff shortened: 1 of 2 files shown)


xtqqczze:
Alternatively, do not gate the `into()` call, and use something like:

```rust
#[cfg_attr(target_pointer_width = "32", expect(clippy::useless_conversion))]
```

ntpd/src/daemon/clock.rs
```diff
@@ -40,11 +40,10 @@ impl NtpClock for NtpClockWrapper {
...
     ) -> Result<ntp_proto::NtpTimestamp, Self::Error> {
         let (seconds, nanos) = offset.as_seconds_nanos();
+        #[cfg(not(target_pointer_width = "32"))]
+        let seconds = seconds.into();
         self.0
-            .step_clock(TimeOffset {
-                seconds: seconds as _,
-                nanos,
-            })
+            .step_clock(TimeOffset { seconds, nanos })
```
davidv1992: I think we are better off if we do the conversion unconditionally, and just add a specific allow here to allow the useless into here, with a reason stating that it is to deal with the difference between 32 and 64 bit time types.

davidv1992 (changes requested):
Thank you for providing this. I think for the one case still triggering on this, we should want a different solution, see below.

davidv1992 (approved):

davidv1992 (approved):

PR #2336, opened 2026-07-08
Add clippy::if_not_else

stefanvi:
This lint should be discussed


davidv1992:
We've discussed this internally, and decided we don't like what this lint does to our code, so we are going to disable it. Leaving this PR open as a reminder to do that.

rnijveld:
Replaced by #2346

PR #2337, opened 2026-07-08
Fix clippy warnings on MSVR

stefanvi:
When building the project with version 1.88, clippy warns about some of the lints. This PR
- adds `#![allow(renamed_and_removed_lints)]` to `statime-wire`, allowing for the old names of `empty_enum` and `unchecked_duration_subtraction`.
- Removes the `ip_constant` lint, which is only available from version 1.89. This is commented out.
- Fixes some instances of `uninlined_format_args` that 1.88 seems to be stricter about


davidv1992:
No longer relevant as we bumped the MSRV

Issue #1913, opened 2025-06-21
Completely unusable on Debian 12


oittaa:
setting /etc/ntpd-rs/ntp.toml to somethging like `path = "/var/run/ntpd-rs.satpulse.sock`

results in:

```
# systemctl status ntpd-rs.service
× ntpd-rs.service - Network Time Service (ntpd-rs)
     Loaded: loaded (/lib/systemd/system/ntpd-rs.service; enabled; preset: enabled)
     Active: failed (Result: signal) since Sat 2025-06-21 23:35:56 CEST; 8min ago
   Duration: 3ms
       Docs: https://github.com/pendulum-project/ntpd-rs
    Process: 1489 ExecStart=/usr/bin/ntp-daemon (code=killed, signal=ABRT)
   Main PID: 1489 (code=killed, signal=ABRT)
        CPU: 3ms

Jun 21 23:35:56 time systemd[1]: Started ntpd-rs.service - Network Time Service (ntpd-rs).
Jun 21 23:35:56 time ntp-daemon[1489]: 2025-06-21T21:35:56.447365Z  INFO ntpd::daemon::config: using config file at default location `"/etc/ntpd-rs/ntp.toml"`
Jun 21 23:35:56 time ntp-daemon[1489]: 2025-06-21T21:35:56.448275Z  INFO ntpd::daemon::system: new source source_id=SourceId(1) addr="/var/run/ntpd-rs.satpulse.sock" spawner=SpawnerId(2)
Jun 21 23:35:56 time ntp-daemon[1489]: thread 'tokio-runtime-worker' panicked at ntpd/src/daemon/sock_source.rs:218:50:
Jun 21 23:35:56 time ntp-daemon[1489]: Could not create socket: Os { code: 13, kind: PermissionDenied, message: "Permission denied" }
Jun 21 23:35:56 time ntp-daemon[1489]: note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
Jun 21 23:35:56 time systemd[1]: ntpd-rs.service: Main process exited, code=killed, status=6/ABRT
Jun 21 23:35:56 time systemd[1]: ntpd-rs.service: Failed with result 'signal'.

```

garyemiller:
[...]

PR #2338, opened 2026-07-09
misaddress #1913 by basically including "chgrp 123" & "chmod g+w" in sock_sources...

JamesB192:
This is the toy model.

I expect/welcome constructive critique/pushback.

I think it needs to be configurable.

I wonder which of the following is least objectionable.

```toml
[[source]]
mode = "sock"
path = "/tmp/chrony.ttyS0.sock"
precision = 1e-3
template_from = "/etc/ntpd-rs/sock"

[[source]]
mode = "sock"
path = "/tmp/chrony.ttyS1.sock"
precision = 1e-3
group= 123
mode = 0o620 

[[source]]
mode = "sock"
path = "/tmp/chrony.ttyS2.sock"
precision = 1e-3
group = 123
mode_set = 0o020
model_clear = 0o200
```

Performing group ID lookups by name without shunting into C seems problematic, so I won't.

### DIFF
--- a/docs/examples/conf/ntpd-rs.service
+++ b/docs/examples/conf/ntpd-rs.service
@@ -13,6 +13,7 @@ Environment="RUST_LOG=info"
 RuntimeDirectory=ntpd-rs
 User=ntpd-rs
 Group=ntpd-rs
+SupplementaryGroups=123
 AmbientCapabilities=CAP_SYS_TIME CAP_NET_BIND_SERVICE
 
 [Install]
--- a/ntpd/src/daemon/sock_source.rs
+++ b/ntpd/src/daemon/sock_source.rs
@@ -1,3 +1,4 @@
+use std::os::unix::fs::PermissionsExt;
 use std::path::PathBuf;
 use std::{fmt::Display, path::Path};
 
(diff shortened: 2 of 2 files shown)


davidv1992:
Can you provide more context on what this is supposed to do? It is very unclear to me what this is trying to fix, and how it is trying to do that at the moment.

JamesB192:
Probably not well, but I'll try.

This MR changes the ownership and permissions of the Unix datagram sockets used by the sock source. This allows a gpsd (or ntp-refclock) instance started without root privileges to communicate with ntpd-rs. It (generally) requires that gpsd is started in groups 123 and dialout so that the samples may be forwarded once processed from the source. Also, ntpd-rs should have 123 as an additional group.

davidv1992:
I am not sure this addresses the underlying concern of #1913. The problem identified there is that gpsd wants the sockets at a very specific path which is only writable for root, and which cannot (securely) be made easily writable by the unprivileged ntpd-rs user. That socket is then opened by gpsd, which is always running as root.

The patch here seems primarily useful when running gpsd not as root, which the maintainers of that project have strongly indicated they see as completely unreasonable, so there seems to be little incentive for us to support it, at least for the purpose of interacting with gpsd.

Given that, I am missing the usecase for this a bit. Did you have an idea of a usecase outside of gpsd?

JamesB192:
Ah, you're one of those. nevermind

Issue #2326, opened 2026-07-03
Implement proper listener binding for CSPTP servers


davidv1992:
We may want to combine this with larger refactors of how we handle servers, so that some of the server functionality from ntp can be reused

Issue #2324, opened 2026-07-03
Figure out a new configuration structure


davidv1992:
The current configuration structure contains a bunch of aspects that are fairly NTP specific, in places that on their own aren't. This needs fixing.

PR #2339, opened 2026-07-09
Integration of CSPTP into the time daemon

davidv1992:
This contains the final bits of infrastructure to integrate csptp into ntpd-rs.

It has the following limitations still:
Servers always bind to all interfaces and all IP-addresses. (open as issue #2326)
The global (cross-source/server) configuration for csptp is somewhat awkwardly bolted on to the current configuration. (open as issue #2324)


ntpd/src/daemon/spawn/csptp.rs
```diff
@@ -0,0 +1,223 @@
...
+            return Ok(());
+        };
+        action_tx
+            .send(SpawnEvent::new(
+                self.id,
+                SpawnAction::Create(SourceCreateParameters::Csptp(CsptpSourceCreateParameters {
+                    id: ClockId::new(),
+                    addr,
+                    config: self.config.clone(),
+                })), /*SpawnAction::create_ntp(
```
rnijveld: Comment can be removed?

ntpd/src/daemon/config/mod.rs
```diff
@@ -290,6 +290,50 @@ impl TimestampMode {
     }
 }
 
+#[cfg(target_os = "linux")]
```
rnijveld, line 293: We should probably put all of this into its own module eventually so we don't sprinkle `#[cfg(target_os = "linux"]` all over all the time, but since we're probably going to rework this, maybe not right now?

ntpd/src/daemon/system.rs
```diff
@@ -131,6 +158,14 @@ struct SystemTask<C: NtpClock, Controller: TimeSyncController<Clock = C>> {
     controller: Arc<Controller>,
     ntp_manager: Arc<NtpManager>,
 
+    #[cfg(target_os = "linux")]
+    ptp_networking_ipv4: Option<&'static statime_netptp::NetworkManager<Ipv4Addr>>,
```
rnijveld: If there is a specific reason why these are leaked boxes instead of just `Option<Box<_>>` I think we should document that. If there is no reason then we should just use `Option<Box<_>>`.

rnijveld (changes requested):

rnijveld (approved):

Issue #2131, opened 2026-02-26
IPv4 addresses in allowlist don't work when listening on IPv6


tohojo:
When using a config like the following:

```toml
[[server]]
listen = "[::]:123"
[server.allowlist]
filter = ["10.0.0.0/16"]
action = "ignore"
```

all replies from clients in the allowed subnet are ignored. Switching the config to using IPv4-mapped IPv6 addresses instead makes things work as expected:

```toml
[[server]]
listen = "[::]:123"
[server.allowlist]
filter = ["::ffff:10.0.0.0/112"]
action = "ignore"
```

This is somewhat surprising, and hard to debug (filtering decisions do not seem to appear in the logs, even with `log-level = "trace"`). I would expect ntpd-rs to be smart enough to recognise the mapped addresses as v4 addresses and correctly apply the allowlist rules.

upmcplanetracker:
Can you break them out into 2 separate pieces?

```
[[server]]
listen = "0.0.0.0:123"
[server.allowlist]
filter = ["10.0.0.0/16"]
action = "ignore"

[[server]]
listen = "[::]:123"
[server.allowlist]
filter = ["::1/128"] # or your IPv6 range
action = "ignore"
```

Also maybe nothing is being logged because the packets are being rejected by the os/kernel so it isn't even hitting ntpd-rs

Try
`sudo tcpdump -ni any port 123`
And watch the traffic.

rnijveld:
Note: This same problem also exists for blocklists, where the specific issue could also cause that an admin thinks they've excluded a specific address, but that address is still able to reach the server via an IPv6 mapped IPv4 address. We would very much welcome any contributions to make this a little bit less of a paper cut. It should be relatively easy to check if we're dealing with a mapped IPv6 address and check for the 'unmapped' IPv4 address as well when we encounter one.

upmcplanetracker:
I think I have a work around. Will try a pr in a bit. 

edit - https://github.com/pendulum-project/ntpd-rs/pull/2350

PR #2350, opened 2026-07-16
Proposed fix IPv4-mapped IPv6 addresses bypassing allowlist/denylist filters (issue #2131)

upmcplanetracker:
Proposed fix for issue #2131 

Problem: IpFilter keeps two independent match structures, `ipv4_filter` and `ipv6_filter`, and `is_in()` dispatched purely on the `IpAddr enum` variant. When a server listens on a dual-stack IPv6 socket `(listen = "[::]:123")`, incoming IPv4 client packets are delivered as IPv4-mapped IPv6 addresses `(::ffff:a.b.c.d)` which is a V6 variant. A plain `10.0.0.0/16` entry in `[server.allowlist]` lives in `ipv4_filter` and was never consulted for such addresses, so the allowlist (and equally, the denylist, per @rnijveld's comment on the issue) silently failed to match.

Proposed fix: `IpFilter::is_in` now checks a V6 address in two ways and ORs the result:
1. as given, against the IPv6 tree - unchanged, so existing configs that already use the `::ffff:10.0.0.0/112`-style workaround mentioned in the issue keep working exactly as before.
2. If `Ipv6Addr::to_ipv4_mapped()` returns `Some` (i.e. it really is an IPv4-mapped address), also check the unmapped IPv4 form against the IPv4 tree.

this matches @rnijveld's [suggestion](https://github.com/pendulum-project/ntpd-rs/issues/2131#issuecomment-4990595267).

An earlier version of this fix collapsed the address to one canonical form via `IpAddr::to_canonical()` before a single lookup. That's simpler, but it regressed the existing `test_filter` test: a user who already works around this bug today by writing filter entries as `::ffff:.../mask` would have that entry silently stop matching, because canonicalizing the incoming address routes the lookup to the IPv4 tree only. The OR-based approach avoids that regression.

since `Server::intended_action` calls both `denyfilter.is_in()` and `allowfilter.is_in()` through this same method, this one change fixes the paper cut for both the allowlist and the denylist.

I also changed: the fuzz reference implementation `(fuzz::contains)` matched strictly by `IpAddr` variant too. It now mirrors the same OR-based logic — a (V4 filter entry, V6 address) pair is checked via `to_ipv4_mapped()` so the fuzz equivalence assertion `(is_in(addr) == any_contains(nets, addr))` stays valid instead of spuriously failing on mapped addresses.

added `test_ipv4_mapped_address_matches_plain_ipv4_filter`,reproducing the exact config shape from the issue (a plain
`10.0.0.0/16` allowlist entry, no explicit mapped-notation entry) and confirming a mapped address now matches while an unrelated one, and a genuine IPv6 address, still don't.

Added `test_ipv4_mapped_address_matches_plain_ipv4_denylist_style_filter` covering the denylist-shaped case @rnijveld flagged. All pre-existing tests in `ipfilter.rs` still pass, including `test_filter`, which exercises the `::ffff:.../mask`-workaround
path this fix is careful not to break. 

I ran the module's tests (plus a fuzz-equivalence grid check) against a minimal standalone reproduction of this file on an older Rust toolchain, since I didn't have the project's pinned MSRV toolchain available locally. I have not yet run this on the project's actual CI/MSRV toolchain or its real fuzz target.

the issue (#2131 ) also mentions that filtering decisions don't show up in logs even at trace level, making this hard to debug. I've left this request out of this PR to keep the diff focused.

### DIFF
--- a/ntp-proto/src/ipfilter.rs
+++ b/ntp-proto/src/ipfilter.rs
@@ -199,11 +199,24 @@ impl IpFilter {
     }
 
     /// Check whether a given ip address is contained in the filter.
+    ///
+    /// For an IPv4-mapped IPv6 address (`::ffff:a.b.c.d`), this checks both
+    /// the address as given (so existing filter entries written in mapped
+    /// notation, e.g. `::ffff:10.0.0.0/112`, keep working) *and* its
+    /// unmapped IPv4 form (so plain IPv4 filter entries, e.g.
+    /// `10.0.0.0/16`, also match). Without the latter, a dual-stack
+    /// listener (`listen = "[::]:123"`) would receive IPv4 clients as `V6`
+    /// addresses that never match filter entries written in ordinary IPv4
+    /// CIDR notation, even though that's what an operator would expect.
+    /// See <https://github.com/pendulum-project/ntpd-rs/issues/2131>.
+    ///
     /// Complexity: O(1)
     pub fn is_in(&self, addr: &IpAddr) -> bool {
         match addr {
             IpAddr::V4(addr) => self.is_in4(addr),
-            IpAddr::V6(addr) => self.is_in6(addr),
+            IpAddr::V6(addr) => {
+                self.is_in6(addr) || addr.to_ipv4_mapped().is_some_and(|v4| self.is_in4(&v4))
+            }
         }
     }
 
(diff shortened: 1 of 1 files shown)


upmcplanetracker:
oh boy, I didn't mean to do that.  I didn't know that deleting the head repo would close the pull request.

PR #2354, opened 2026-07-17
fix(ipfilter): apply IPv4 filter entries to IPv4-mapped IPv6 sources

### LINKED ISSUE #2131: IPv4 addresses in allowlist don't work when listening on IPv6

ibondarenko1:
Fixes #2131.

On a dual-stack `[::]` listen socket, Linux delivers IPv4 packets with an IPv4-mapped IPv6 source (`::ffff:a.b.c.d`). `IpFilter::is_in` routes those to the v6 tree, so an IPv4 allow/deny entry never matches them and the rule silently does not apply (allowlisted clients get ignored, denylisted clients get served).

Fix: in `IpFilter::new`, mirror each IPv4 subnet entry as `::ffff:net/(96+mask)` in the v6 tree, so the same address matches whether it arrives as IPv4 or IPv4-mapped IPv6. Existing v6 entries and behavior are unchanged.

Added `test_ipv4_mapped_matches_ipv4_entry`. `cargo test -p ntp-proto`, `cargo fmt` and `cargo clippy -p ntp-proto` are clean.

### DIFF
--- a/ntp-proto/src/ipfilter.rs
+++ b/ntp-proto/src/ipfilter.rs
@@ -182,10 +182,14 @@ impl IpFilter {
 
         for subnet in subnets {
             match subnet.addr {
-                IpAddr::V4(addr) => ipv4list.push((
-                    (u32::from_be_bytes(addr.octets()) as u128) << 96,
-                    subnet.mask,
-                )),
+                IpAddr::V4(addr) => {
+                    let mapped = u32::from_be_bytes(addr.octets()) as u128;
+                    ipv4list.push((mapped << 96, subnet.mask));
+                    // Also match this address when it arrives IPv4-mapped
+                    // (::ffff:a.b.c.d) on a dual-stack `[::]` socket, so an
+                    // IPv4 filter entry still applies. See #2131.
+                    ipv6list.push(((0xffff_u128 << 32) | mapped, subnet.mask + 96));
+                }
                 IpAddr::V6(addr) => {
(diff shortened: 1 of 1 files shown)


rnijveld:
Hi, thanks for this! After looking at the code I decided that there was a little bit better way to handle this by canonicalizing all the IP addresses. That prevents duplicating any IP addresses in the in memory storage for the filters. I quickly implemented this in #2359. Let me know if you are ok with that solution as well!

ibondarenko1:
Yes, #2359 is the better fix. Canonicalizing on both the build and lookup paths keeps a single entry per address instead of mirroring every v4 subnet into the v6 tree like mine did, and it also handles the reverse case (a `::ffff:` entry in the config) plus the mapped-mask validation, which my patch didn't. Happy to close this in favor of #2359.

PR #2359, opened 2026-07-19
Use canonical addresses for the IP filter

### LINKED ISSUE #2131: IPv4 addresses in allowlist don't work when listening on IPv6

rnijveld:
Alternative for #2354 and #2353, fixes #2131 

This implementation canonicalizes both the filter IP addresses as well as the IP addresses that are being tested against.

Thanks to @ibondarenko1 and @upmcplanetracker for the inspiration.

### DIFF
--- a/fuzz/fuzz_targets/ipfilter.rs
+++ b/fuzz/fuzz_targets/ipfilter.rs
@@ -22,8 +22,13 @@ impl<'a> Arbitrary<'a> for ASubnet {
             let addr = IpAddr::V4(Ipv4Addr::from(u.arbitrary::<[u8; 4]>()?));
             Ok(ASubnet(IpSubnet { mask, addr }))
         } else {
-            let mask: u8 = u.int_in_range(0..=128)?;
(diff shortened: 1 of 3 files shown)


davidv1992 (approved):

upmcplanetracker:
I appreciate the hard work you guys do on this project

PR #2360, opened 2026-07-19
Allow running without updating system clock (issue 2322)

upmcplanetracker:
addresses issue 2322.  Unclear if developers truly want to add this function, but it was flagged as ok for a PR, so I figured I'd try my hand at it.

tested on Ubuntu 26.04 system on the newest version of ntpd-rs.  I got it to compile with cargo 1.97.1.
I ran it as normal and then with the new flag set to true. I was able to see zero writes using strace using a fake time of 01-01-2020 12:00:00 am. it didn't break it running as normal with the correct time (with the flag `update-system-clock` set as true or omitted), and it was syncing to time servers per usual, so that's a victory in itself.  And when I ran it with the new flag set to false and a fake time (1-1-2020 12:00:00 am) it held that time (and that's when I did the strace writes check).

I added `update-system-clock = bool` (default true) under `[synchronization]` as the issue author suggested.
When set to false:
The daemon never calls `adjtimex`, `clock_adjtime`, `clock_settime`, or `settimeofday` with adjustment flags
The Kalman filter continues to track offset and frequency normally
ServerTask applies the tracked offset to receive timestamps before building NTP responses, so clients still get corrected time
A config warning is emitted on startup if the option is disabled
This allows running ntpd-rs in environments without `CAP_SYS_TIME` while still serving accurate time.

I changed these files-
`ntpd/src/daemon/config/mod.rs` I added config option and validation warning
`ntpd/src/daemon/clock.rs` I added a new SoftClock wrapper and TrueTimeClock trait
`ntpd/src/daemon/server.rs` causes server responses `use to_true_time()`
`ntpd/src/daemon/system.rs` now generic `spawn()` accepts TrueTimeClock bound
`ntpd/src/daemon/observer.rs` this exposes `system_clock_adjustment` in status
`ntpd/src/daemon/mod.rs` this wires SoftClock into main loop

### DIFF
--- a/ntpd/src/daemon/observer.rs
+++ b/ntpd/src/daemon/observer.rs
@@ -19,6 +19,8 @@ pub struct ObservableState {
     pub system: SystemSnapshot,
     pub sources: Vec<ObservableSourceState>,
     pub servers: Vec<ObservableServerState>,
+    pub system_clock_adjustment: bool,
+    pub accumulated_offset_ms: Option<f64>,
 }
 
 #[derive(Debug, Serialize, Deserialize)]
@@ -74,12 +76,13 @@ pub fn spawn<C: 'static + NtpClock + Send>(
     server_reader: tokio::sync::watch::Receiver<Vec<ServerData>>,
     system_reader: tokio::sync::watch::Receiver<SystemSnapshot>,
     clock: C,
+    update_system_clock: bool,
 ) -> JoinHandle<std::io::Result<()>> {
     let config = config.clone();
     tokio::spawn(
         (async move {
             let result =
-                observer(config, sources_reader, server_reader, system_reader, clock).await;
+                observer(config, sources_reader, server_reader, system_reader, clock, update_system_clock).await;
             if let Err(ref e) = result {
                 warn!("Abnormal termination of the state observer: {e}");
                 warn!("The state observer will not be available");
@@ -96,6 +99,7 @@ async fn observer<C: 'static + NtpClock + Send>(
     server_reader: tokio::sync::watch::Receiver<Vec<ServerData>>,
     system_reader: tokio::sync::watch::Receiver<SystemSnapshot>,
     clock: C,
+    update_system_clock: bool,
 ) -> std::io::Result<()> {
     let start_time = Instant::now();
     let timeout = std::time::Duration::from_millis(500);
@@ -172,6 +176,7 @@ async fn handle_connection(
     server_reader: tokio::sync::watch::Receiver<Vec<ServerData>>,
     system_reader: tokio::sync::watch::Receiver<SystemSnapshot>,
     now: NtpTimestamp,
+    update_system_clock: bool,
 ) -> std::io::Result<()> {
     let observe = ObservableState {
         program: ProgramData::with_dynamics(start_time.elapsed().as_secs_f64(), now),
@@ -183,6 +188,8 @@ async fn handle_connection(
             .collect(),
         system: *system_reader.borrow(),
         servers: server_reader.borrow().iter().map(Into::into).collect(),
+        system_clock_adjustment: update_system_clock,
+        accumulated_offset_ms: None, // TODO: wire through SoftClock when available
     };
 
     super::sockets::write_json(stream, &observe).await?;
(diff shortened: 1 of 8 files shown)


signalsforgranted:
> Unclear if developers truly want to add this function

Hi, yes, hello, I would like this please. Specifically because over a weekend I have developed a very dodgy [interop testing harness](https://github.com/signalsforgranted/zeitwaage) of NTP implementations, and when all you care about is the shape of the packets, having multiple NTP clients constantly move the system clock it's a bad time for everyone involved. As part of that, @davidv1992 very quickly and generously made [this patch](https://github.com/signalsforgranted/zeitwaage/blob/main/implementations/ntpd-rs/0001-Prototype-for-not-touching-the-system-clock.patch) to meet my needs, but I would propose this would be useful for others (for example, `chronyd` has the `-x` option which similar, which can't have been put there on a whim).

upmcplanetracker:
It is a hot, wildfire smoke filled Sunday here in the North Eastern US so I
was just excited for a challenge. Ive been thinking about this for a while
and figured it wouldn't be too hard. (Not like the feature request someone
else asked for to have ntpd-rs update the config without restarting...)
So if the devs like @davidv1992 want to use mine as a base or his patch (or
just delete this PR) I'm fine any which way. It kept me busy today.

upmcplanetracker:
> > Unclear if developers truly want to add this function
> 
> Hi, yes, hello, I would like this please. Specifically because over a weekend I have developed a very dodgy [interop testing harness](https://github.com/signalsforgranted/zeitwaage) of NTP implementations, and when all you care about is the shape of the packets, having multiple NTP clients constantly move the system clock it's a bad time for everyone involved. As part of that, @davidv1992 very quickly and generously made [this patch](https://github.com/signalsforgranted/zeitwaage/blob/main/implementations/ntpd-rs/0001-Prototype-for-not-touching-the-system-clock.patch) to meet my needs, but I would propose this would be useful for others (for example, `chronyd` has the `-x` option which similar, which can't have been put there on a whim).

Also would you mind pulling ntpd-rs from my forked repo and building it and seeing if it does what you want it to do.  I think that would be important feedback if the devs move forward because if it does what you want it to do then they wouldn't have to make many changes. If what I built doesn't break anything but is worthless for your use case, then that's good for me and the devs to know. 

I was able to pull and build it on my Ubuntu 26.04 system without much of an issue. Of course, even though 26.04 is new-ish their version of cargo is too old for ntpd-rs so I had to update it first to a newer version. 😄 Sometimes Ubuntu releases are crazy up to date with components at the time of release (like podman), other times they're still using much older versions of things...

davidv1992:
Thank you for providing this. We will take a deeper look next week, for now I have approved running the CI. Feel free to already try to resolve those issues, as that will be needed for merging anyway.

davidv1992 (changes requested):
Structurally, this is looking good to me. However, it needs a few fixes to make it past CI before we can move forward with this. Could you make those?

upmcplanetracker:
I think I fixed it, missed a few lines.

ntpd/src/daemon/observer.rs
```diff
@@ -183,6 +189,8 @@ async fn handle_connection(
             .collect(),
         system: *system_reader.borrow(),
         servers: server_reader.borrow().iter().map(Into::into).collect(),
+        system_clock_adjustment: update_system_clock,
+        accumulated_offset_ms: None, // TODO: wire through SoftClock when available
```
davidv1992, line 193: There is a TODO remaining here.

davidv1992 (commented):
Took another look, moslty looks good, but still fails CI, and there is an open TODO that at least needs some discussion for us. Thank you for taking the time to do this.

Tip for faster iteration, ensure that at least cargo test, cargo clippy and cargo fmt run succesfully locally. That doesn't catch everything but this is platform independent so should get most of the stuff that is still a problem.

@rnijveld I think this is a good approach, at the very least a good base for us to build on once the larger refactor gets here. But can you take a quick look to see you agree?

upmcplanetracker:
hi -- i'm stuck. unfortunately I bit off way more than I can chew here. I apologize.  I don't want you guys to take up any more time waiting for me since I don't think I'll be able to make progress on this.  I'm going to close this PR.  If you guys want to move ahead with what i've done, that's fine.  If not, that's fine too.  I appreciate everything you guys do re: pushing this project forward.

davidv1992:
@upmcplanetracker Thank you very much for the effort you have already put in, this is a really nice starting point for us and really useful. Would it be okay with you if we take this to the finish line?

PR #2446, opened 2026-08-19
Integration of timestamped-socket

davidv1992:
Integrating timestamped-socket crate into the repository, and moving away from its timestamp type.

### DIFF
--- a/statime-base/src/time_types.rs
+++ b/statime-base/src/time_types.rs
@@ -112,6 +112,54 @@ impl<'de> Deserialize<'de> for Duration {
     }
 }
 
+/// A timestamp that can be in either the UTC or TAI timescales.
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Hash)]
+pub enum UniversalTimestamp {
+    /// A UTC timestamp
+    Utc(Timestamp<UTC>),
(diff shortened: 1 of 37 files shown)


statime-base/src/time_types.rs
```diff
@@ -112,6 +112,54 @@ impl<'de> Deserialize<'de> for Duration {
     }
 }
 
+/// A timestamp that can be in either the UTC or TAI timescales.
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Hash)]
+pub enum UniversalTimestamp {
```
rnijveld, line 117: As an alternative to this: maybe we can actually update our Timestamp type to `pub struct Timestamp<Timescale>(u128, Timescale)` (should be no difference in actual representation for our existing UTC and TAI markers). And then we can introduce a new marker type `enum Universal { UTC, TAI }`. That feels a little cleaner than this enum.

davidv1992: The problem is that with that change, doing calculations on timestamps becomes a lot harder, as differences are not always meaningfull anymore (they are for TAI and UTC markers, but not in the universal case).

rnijveld (approved):

PR #2460, opened 2026-08-28
Implement listing of active links with impact on system clock phase.

davidv1992:
This provides functionality needed for loop avoidance in e.g. ntp and csptp.

### DIFF
--- a/statime-algo/src/estimator.rs
+++ b/statime-algo/src/estimator.rs
@@ -406,7 +406,8 @@ impl<Storage: KalmanStorageBase> EstimatorState<Storage> {
         direction: DirectedLinkId,
         offset: UncertainValue,
         delay_link: bool,
-    ) -> Result<Self, AlgoError> {
+        system_clock: ClockId,
+    ) -> Result<(Self, f64), AlgoError> {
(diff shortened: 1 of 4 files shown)


statime-algo/src/estimator.rs
```diff
@@ -406,7 +406,8 @@ impl<Storage: KalmanStorageBase> EstimatorState<Storage> {
         direction: DirectedLinkId,
         offset: UncertainValue,
         delay_link: bool,
-    ) -> Result<Self, AlgoError> {
+        system_clock: ClockId,
```
rnijveld, line 409: Why do we need this? Isn't the system clock always the first clock in the list of clocks within the estimator anyway?

davidv1992: Not necessarily once we make the system clock dynamic. The estimator itself doesn't really have a concept of system clock anyway, so being explicit here is better I think.

davidv1992: Ok, after in person discussion, lets fix the awareness in this and the filter layer in a separate PR, see issue #2466

rnijveld (approved):

PR #2500, opened 2026-09-17
Fix statime-base root dispersion in no_std builds

### LINKED ISSUE #2496: statime-base: root_dispersion fails to compile without the std feature
## Summary

`cargo check -p statime-base --lib --no-default-features --locked` fails because
`TimeSnapshot::root_dispersion()` calls `f64::powi` and `f64::sqrt`, which are
unavailable in this crate with `std` disabled. This also prevents building
`statime-algo` without its default features.

This reproduces on a desktop host without an application or custom build integration.

## Version and environment

Rechecked against upstream `main` on 2026-09-17:

- Commit: `01e1362f1fbc32e9bb810bd18682d512d5f1fa95`.
- `statime-base` and `statime-algo`: `2.0.0-alpha.20260715`, edition 2024.
- Declared MSRV: Rust 1.95; tested with the repository's current stable version,
  Rust **1.98.1** (`48a229ceaefd4985c50990b14116b6d856af0985`, 2026-09-01).
- Cargo **1.98.1** (`797e8a9bc`, 2026-08-05).
- Host: `x86_64-apple-darwin`; Homebrew Rust; LLVM 22.1.8.
- Upstream sources and `Cargo.lock` were unmodified. MSRV was not separately tested.

## Minimal reproduction

```sh
git clone https://github.com/pendulum-project/ntpd-rs.git
cd ntpd-rs
git checkout --detach 01e1362f1fbc32e9bb810bd18682d512d5f1fa95
cargo check -p statime-base --lib --no-default-features --locked
```

Expected: the library compiles without `std`.

Actual: exit status 101, with these three diagnostics:

```text
error[E0599]: no method named `powi` found for type `f64` in the current scope
   --> statime-base/src/algorithm.rs:212:21
error[E0599]: no method named `powi` found for type `f64` in the current scope
   --> statime-base/src/algorithm.rs:213:21
error[E0599]: no method named `sqrt` found for type `f64` in the current scope
   --> statime-base/src/algorithm.rs:214:18
```

These exact checks were also run with `--offline` after dependency retrieval:

```sh
cargo check -p statime-base --lib --no-default-features --locked --offline
cargo check -p statime-base --lib --locked --offline
cargo check -p statime-algo --lib --no-default-features --locked --offline
cargo test -p statime-base -p statime-algo --lib --locked --offline
```

Results: the first and third fail with the same E0599 errors; the default-feature
base check passes. Existing default-feature library tests pass: 10 in
`statime-base`, 43 in `statime-algo`.

## Cause and suggested minimal fix

The [affected function](https://github.com/pendulum-project/ntpd-rs/blob/01e1362f1fbc32e9bb810bd18682d512d5f1fa95/statime-base/src/algorithm.rs#L206)
is compiled unconditionally. The `libm` polyfill in `statime-algo` is internal to
that crate, so it cannot supply the missing methods in `statime-base`.

A small fix is to add the existing workspace `libm` dependency to `statime-base`,
replace the two small integer powers with multiplication, and use `libm::sqrt`:

```toml
# statime-base/Cargo.toml, under [dependencies]
libm.workspace = true
```

```rust
let t = (now - self.root_variance_base_time).as_seconds();
let variance = self.root_variance_base
    + t * self.root_variance_linear
    + (t * t) * self.root_variance_quadratic
    + (t * t * t) * self.root_variance_cubic;
Duration::from_f64_seconds(libm::sqrt(variance))
```

This retains the variance polynomial and the required conversion from variance
to standard deviation. It needs no allocation, new feature, or MSRV increase.
`libm` is already a workspace dependency at 0.2.16 and present in the lockfile;
the lockfile's `statime-base` dependency list would need updating. Multiplication
is sufficient for exponents 2 and 3; a general power-function adapter is unnecessary.

This simplest form uses `libm::sqrt` in both feature modes. If preserving the
existing native `std` square-root path is preferred, select that path with
`#[cfg(feature = "std")]` and `libm::sqrt` only without `std`; that adds a little
feature-specific code. Numerical tests should use a stated tolerance where
results are not exactly representable, rather than require bitwise identity
between math implementations. Do not remove the square root or clamp the
polynomial as part of this compilation fix.

This is a proposed fix, not an applied or validated patch. A passing full
`statime-algo` no_std build after this change has not been established.

## Regression coverage

Add isolated library checks for each package in CI, on stable and preferably
MSRV, using the existing toolchain setup:

```sh
cargo check -p statime-base --lib --no-default-features --locked
cargo check -p statime-algo --lib --no-default-features --locked
```

Run these as separate package selections, not as `--workspace` checks. Current
workspace consumers explicitly enable `statime-base/std` and
`statime-algo/std` in `ntpd/Cargo.toml`; `ntp-proto` also enables
`statime-base/serde`, which enables `std`. Cargo feature unification can therefore
mask this failure even in a workspace command using `--no-default-features`.

Also add a `root_dispersion` unit test: set base/linear/quadratic/cubic variance
coefficients to `1, 2, 3, 4`, and evaluate at the base time and two seconds later.
Expected dispersions are exactly 1 second and 7 seconds (`sqrt(49)`), respectively.
Add an all-zero-coefficient case expecting zero. This checks the powers and
preserves the square root independently of the compile check. Default-feature
unit tests plus the isolated no_std checks avoid depending on whether the
existing test harness itself supports no-default-features tests.

## Related reports checked

On 2026-09-17, I searched open and closed issues/PRs for `no_std`,
`no-default-features`, `root_dispersion`, `powi`, `sqrt`, and `statime`, and
reviewed current open PRs. I did not find a report or fix for this specific
`statime-base` compilation failure.

- [Issue 2285](https://github.com/pendulum-project/ntpd-rs/issues/2285) and
  [merged PR 2294](https://github.com/pendulum-project/ntpd-rs/pull/2294) address
  broader feature-combination coverage. PR 2294 added workspace feature-matrix
  entries and changes in `statime-wire`, not isolated checks for this package.
- [Merged PR 1931](https://github.com/pendulum-project/ntpd-rs/pull/1931) fixed a
  missing square root in the earlier `ntp-proto` implementation. It is a distinct
  bug and confirms why the square root must remain.
- [Merged PR 2441](https://github.com/pendulum-project/ntpd-rs/pull/2441) moved
  `TimeSnapshot` into `statime-base`.
- [Open PR 2485](https://github.com/pendulum-project/ntpd-rs/pull/2485) changes
  controller error propagation; its inspected diff does not fix this function.

jowgn:
Fixes #2496

`statime-base` and its consumer `statime-algo` fail isolated library builds with default features disabled: `TimeSnapshot::root_dispersion()` calls `f64::powi` and `f64::sqrt`, which are unavailable without `std`. The float polyfill in `statime-algo` cannot supply methods inside `statime-base`.

Use multiplication for the square and cube and the existing workspace `libm` dependency for `sqrt` in both feature modes. This preserves the variance polynomial, its conversion to standard deviation, and the existing `Duration` conversion. It does not clamp the polynomial, enable `std`, or raise the MSRV. Both workspace lockfiles receive only the new `statime-base` dependency entry.

Regression coverage includes:

- Numerical tests with variance coefficients `1, 2, 3, 4`, checking dispersions of 1 second at the base time and 7 seconds two seconds later, plus zero variance. A fractional-time case checks `sqrt(3.25)` with an absolute tolerance of `1e-12` seconds.
- Separate no-default-features library checks for each crate inside the existing Build and test job, in the stable/default-feature Linux GNU and MSRV Linux GNU matrix entries. These reuse the existing toolchain setup and Rust cache, including the MSRV job's direct-minimal-versions dependency resolution. Separate package selections prevent workspace consumers from masking the failure by enabling `std` through feature unification.

Upstream base: `01e1362f1fbc32e9bb810bd18682d512d5f1fa95`. Both isolated checks reproduced all three reported E0599 errors on that unmodified commit. Upstream main was fetched again before opening this draft and remained at that commit; the issue is open and no overlapping fix was found among existing PRs.

Verification on `x86_64-apple-darwin`, Rust 1.98.1:

```sh
cargo check -p statime-base --lib --no-default-features --locked
cargo check -p statime-algo --lib --no-default-features --locked
cargo test -p statime-base -p statime-algo --lib --locked
cargo test --workspace --locked --no-fail-fast
cargo fmt --all --check
cargo fmt --manifest-path ./fuzz/Cargo.toml --all --check
cargo fmt --manifest-path ./fuzz/fuzz_rand_shim/Cargo.toml --all --check
cargo clippy --target x86_64-apple-darwin --workspace --all-targets --features hardware-timestamping,__internal-fuzz,__internal-test,__internal-api --locked -- -D warnings
git diff --check
```

All passed. The focused library run passed 12 `statime-base` tests and 43 `statime-algo` tests. The first `cargo test --workspace --locked` run inside the sandbox failed 12 NTS certificate-validation tests; the full `--no-fail-fast` rerun outside the sandbox passed. Clippy exited successfully with the existing warning that `clippy::from_iter_instead_of_collect` has been removed.

MSRV verification with Rust 1.95.0 also passed both isolated checks and all 55 library tests. The compiler was selected explicitly because Homebrew Rust takes precedence in the host shell:

```sh
export RUSTC=/Users/joschawagner/.rustup/toolchains/1.95-x86_64-apple-darwin/bin/rustc
rustup run 1.95 cargo check -p statime-base --lib --no-default-features --locked
rustup run 1.95 cargo check -p statime-algo --lib --no-default-features --locked
rustup run 1.95 cargo test -p statime-base -p statime-algo --lib --locked
```

MSRV was tested with the committed lockfile; the Linux direct-minimal-versions CI configuration was not run locally. The workflow YAML was parsed locally and its condition checked to select exactly the intended stable and MSRV matrix entries. This follows the issue's proposed minimal fix.


statime-base/src/algorithm.rs
```diff
@@ -206,13 +206,12 @@ impl TimeSnapshot {
     pub fn root_dispersion(&self, now: Timestamp<TAI>) -> Duration {
         let t = (now - self.root_variance_base_time).as_seconds();
         // Note: dispersion is the standard deviation, so we need a sqrt here.
-        Duration::from_f64_seconds(
-            (self.root_variance_base
+        Duration::from_f64_seconds(libm::sqrt(
```
davidv1992: I would prefer only falling back to libm implementations if we really have to. Would it be possible to convert this into a similar polyfill as is already present in statime-algo?

davidv1992 (changes requested):
Mostly looks good, with good tests. One minor thing I would prefer to have more uniform with the rest of the repository.

statime-base/src/algorithm.rs
jowgn: fair point. im on it.

jowgn: changed it

davidv1992:
Noticed you created a merge commit. We prefer to keep our history linear, so could you replace that with a rebase?

jowgn:
> Noticed you created a merge commit. We prefer to keep our history linear, so could you replace that with a rebase?

Yes sorry for the inconvenience.

jowgn:
@davidv1992 would you prefer the FloatPolyfill to both exist in base and algo separately or only in base since it uses base anyhow?

davidv1992 (approved):
Looks good as is, lets keep the polyfill in both for now. Hopefully core::f64::sqrt and similar will get stabilized somewhat soon.

PR #2508, opened 2026-09-22
Fix ntp-daemon not displaying error messages when receiving invalid arguments

olimar718:
Currently `ntp-daemon` prints nothing and exits when receiving an invalid argument

eg 

`ntp-daemon -l <invalid level>`

Prints nothing

This is due to the main function in `ntpd/src/bin/ntp-daemon.rs`
The content of the error is not printed when it occurs, only the non zero error code (1) is returned

My changes fix this

Now
`ntp-daemon -l <invalid level>`

Prints `invalid log level`

This fixes other messages as well


davidv1992 (dismissed):

davidv1992:
Thank you for the fix:)

davidv1992:
Can you fix the formatting? Just running cargo fmt should be sufficient.

olimar718:
I fixed the formatting as requested. As for the coverage problem it was already not covered before (in my understanding, very new to this codebase)

davidv1992 (approved):

davidv1992:
Thx, don't worry too much about the coverage stuff, thats mostly informative for us

